Issue #170, opened 2020-05-11
Support multi-tenancy for iOS DeviceCheck


sethvargo:
The initial implementation of DeviceCheck only allows a single app per server. We need to support a per-app configuration. Users will likely need to store a _reference_ to a secret in the database and then the app looks up that secret at runtime on a per-app basis.

- Refs #153 
- Pending on #154

PR #236, opened 2020-05-14
Move iOS verification into the database layer

sethvargo:
This introduces the ability specify per-app iOS DeviceCheck settings by adding new database columns. It also cleans up a few default/null values and provides plumbing for resolving secret values in the data layer.

Fixes GH-170


sethvargo:
This has become detached with moving the repo from private -> public. Gonna open another one.

Issue #205, opened 2020-05-13
Create single entrypoint binary


sethvargo:
Create a single binary that has all the handlers installed and can demo a complete e2e working system locally.

PR #244, opened 2020-05-14
Add mono entrypoint that runs all the handlers

sethvargo:
/assign @mikehelmick 

Fixes GH-205


sethvargo:
@googlebot i signed it

sethvargo:
okay this is in a sad state with the open sourcing. gonna retry.

PR #259, opened 2020-05-14
Add mono entrypoint that runs all the handlers

sethvargo:
Fixes GH-205


sethvargo:
Okay, trying this again with a fresh fork 😄 

/assign @mikehelmick

PR #260, opened 2020-05-14
add federation services to terraform config

danawillow:
I haven't actually run this yet, but the `terraform plan` output looks right.

I noticed there were a lot of env variables that seemed like they would apply across all our services, so I extracted those out into a local var. PTAL if any seem like they should not apply across all services.

I only added the noauth IAM for `federationout` since that's how it's configured in the test project.

I also didn't add the extra `gcloud alpha run` to deploy.yaml for these services since I wasn't sure why we would need them if we're using terraform to run the services.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -256,6 +256,51 @@ resource "google_cloudbuild_trigger" "build-and-publish" {
   }
 }
 
+locals {
+  common_cloudrun_env_vars = [
+    {
+      name  = "DB_POOL_MIN_CONNS"
+      value = "2"
+    },
+    {
+      name  = "DB_POOL_MAX_CONNS"
+      value = "10"
+    },
+    {
+      name  = "DB_PASSWORD_SECRET"
+      value = google_secret_manager_secret_version.db-pwd-initial.name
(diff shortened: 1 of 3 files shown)


danawillow:
/assign @crwilcox 
(since helmick is out today)

terraform/main.tf
```diff
@@ -256,6 +256,51 @@ resource "google_cloudbuild_trigger" "build-and-publish" {
...
+      name  = "DB_POOL_MIN_CONNS"
+      value = "2"
+    },
+    {
+      name  = "DB_POOL_MAX_CONNS"
+      value = "10"
+    },
+    {
+      name  = "DB_PASSWORD_SECRET"
+      value = google_secret_manager_secret_version.db-pwd-initial.name
```
sethvargo, line 271: We changed this a bit. Secrets no longer need the `_SECRET` suffix and should start with `secret://` instead:

```terraform
name = "DB_PASSWORD"
value = "secret://${google_secret_manager...}"
```

danawillow: Thanks, fixed.

PR #262, opened 2020-05-14
Move iOS verification into the database layer

sethvargo:
This introduces the ability specify per-app iOS DeviceCheck settings by adding new database columns. It also cleans up a few default/null values and provides plumbing for resolving secret values in the data layer.

Fixes GH-170

### DIFF
--- a/internal/secrets/gcp_secret_manager.go
+++ b/internal/secrets/gcp_secret_manager.go
@@ -28,16 +28,17 @@ import (
 	secretmanagerpb "google.golang.org/genproto/googleapis/cloud/secretmanager/v1"
 )
 
-var (
-	// cacheDuration is how long to cache a secret.
-	cacheDuration = 5 * time.Minute
+const (
+	// defaultCacheDuration is how long to cache a secret.
+	defaultCacheDuration = 5 * time.Minute
 )
 
 type GCPSecretManager struct {
 	client *secretmanager.Client
 
-	cache      map[string]*item
-	cacheMutex sync.RWMutex
+	cache         map[string]*item
+	cacheDuration time.Duration
+	cacheMutex    sync.RWMutex
 }
 
 type item struct {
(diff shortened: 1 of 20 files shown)


internal/secrets/gcp_secret_manager.go
```diff
@@ -36,12 +51,24 @@ func NewGCPSecretManager(ctx context.Context) (SecretManager, error) {
...
 }
 
 func (sm *GCPSecretManager) GetSecretValue(ctx context.Context, name string) (string, error) {
 	logger := logging.FromContext(ctx)
 
+	// Check cache.
+	sm.cacheMutex.RLock()
+	if i, ok := sm.cache[name]; ok && i.expiresAt <= time.Now().UnixNano() {
+		sm.cacheMutex.RUnlock()
+		logger.Debugf("found secret in cache: %v", name)
```
mikehelmick, line 67: can probably remove this debug statement

mikehelmick: the other part being the lookup and return

internal/secrets/gcp_secret_manager.go
```diff
@@ -19,15 +19,30 @@ package secrets
...
 
 	"github.com/google/exposure-notifications-server/internal/logging"
 
 	secretmanager "cloud.google.com/go/secretmanager/apiv1"
 	secretmanagerpb "google.golang.org/genproto/googleapis/cloud/secretmanager/v1"
 )
 
+var (
+	// cacheDuration is how long to cache a secret.
+	cacheDuration = 5 * time.Minute
```
mikehelmick, line 33: should probably be a secret manager config so that this can be tuned - can be a follow up / file an issue

internal/secrets/gcp_secret_manager.go
```diff
@@ -53,7 +80,15 @@ func (sm *GCPSecretManager) GetSecretValue(ctx context.Context, name string) (st
 		return "", fmt.Errorf("failed to access secret version for %v: %w", name, err)
 	}
 	logger.Infof("loaded secret value for %v", name)
-
 	plaintext := string(result.Payload.Data)
+
+	// Cache the value.
+	sm.cacheMutex.Lock()
```
mikehelmick, line 86: There's a subtle race here where technically there could be multiple goroutines that exist the Rlock

it may be better to just lock write lock the whole structure - return quickly on a hit and hold it longer on an update.

mikehelmick: it's not an usafe race- it's just that we're not technically preventing a single refresh on expiration, could be > 1

sethvargo: I really wish Go had upgradable locks 😦 . Moved to a full lock.

internal/secrets/gcp_secret_manager.go
sethvargo: Done

internal/secrets/gcp_secret_manager.go
sethvargo: Done

PR #274, opened 2020-05-15
[WIP] add remaining config details to export config, denormalize on batch c…

### LINKED ISSUE #253: ExportConfig - include bundle_id / app_package_name in config
/kind enhancement
/assign @gurayAlsac

mikehelmick:
…reation, and populate export files

fixes #253


migrations/000014_app_to_exportconfig.up.sql
```diff
@@ -0,0 +1,27 @@
...
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+ALTER TABLE ExportConfig ADD COLUMN app_package_name VARCHAR(1000) NOT NULL DEFAULT '';
+ALTER TABLE ExportConfig ADD COLUMN bundle_id VARCHAR(1000) NOT NULL DEFAULT '';
+ALTER TABLE ExportConfig ADD COLUMN signing_key_version VARCHAR(100) NOT NULL DEFAULT '';
+ALTER TABLE ExportConfig ADD COLUMN signing_key_id VARCHAR(255) NOT NULL DEFAULT '';
```
gurayAlsac, line 20: Just to clarify, this will plumb a single signing key through ExportConfig, but it won't necessarily denormalize in a way that supports 1:n configs to signing infos.

internal/database/export.go
```diff
@@ -156,17 +159,17 @@ func (db *DB) AddExportBatches(ctx context.Context, batches []*model.ExportBatch
 		_, err := tx.Prepare(ctx, stmtName, `
 			INSERT INTO
 				ExportBatch
-				(config_id, bucket_name, filename_root, start_timestamp, end_timestamp, region, status, signing_key)
+				(config_id, bucket_name, filename_root, start_timestamp, end_timestamp, region, status, signing_key, signing_key_id, signing_key_version, app_package_name, bundle_id)
```
sethvargo, line 162: Is there a reason to store the key version and key id separately? I don't see them being used separately?

migrations/000007_drop_bypass_safetynet.down.sql
```diff
@@ -15,6 +15,6 @@
 BEGIN;
 
 ALTER TABLE APIConfig
-  ADD COLUMN bypass_safetynet bool NOT NULL;
```
sethvargo, line 18: Is this correct?

migrations/000014_app_to_exportconfig.up.sql
mikehelmick: Right - that's more complicated.

I'll rework this.

internal/database/export.go
mikehelmick: I'm going to need to extract this information out to a separate table anyway - see @gurayAlsac comment in this PR.

migrations/000014_app_to_exportconfig.down.sql
```diff
@@ -0,0 +1,27 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+ALTER TABLE ExportBatch DROP COLUMN app_package_name;
```
sethvargo, line 17: It's faster and safer to do this all in one `ALTER`:

```
ALTER TABLE ExportBatch
  DROP COLUMN app_package_name,
  DROP COLUMN bundle_id,
  DROP COLUMN ...;
```

migrations/000014_app_to_exportconfig.up.sql
```diff
@@ -0,0 +1,27 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+ALTER TABLE ExportConfig ADD COLUMN app_package_name VARCHAR(1000) NOT NULL DEFAULT '';
```
sethvargo, line 17: Same as above, faster to do the ALTER in one go

mikehelmick:
PTAL 

This introduces the concept of SignatureInfo - an export config can be linked to many signature infos.

For each non-expired signature info, the export batch is signed with each signer.

This allows for different signers for different platforms and rotation of the signing keys.

Keeping the hold - want to test this more on Monday

migrations/000019_app_to_exportconfig.up.sql
```diff
@@ -0,0 +1,39 @@
...
+-- This migration is destructive to any existing data.
+
+CREATE TABLE SignatureInfo (
+  id SERIAL PRIMARY KEY,
+  signing_key VARCHAR(500) NOT NULL,
+  app_package_name VARCHAR(1000) NOT NULL DEFAULT '',
+  bundle_id VARCHAR(1000) NOT NULL DEFAULT '',
+  signing_key_version VARCHAR(100) NOT NULL DEFAULT '',
+  signing_key_id VARCHAR(50) NOT NULL DEFAULT '',
+	thru_timestamp TIMESTAMPTZ
```
sethvargo, line 26: - tab vs space
- I think we want to index this (given we're searching on it in the query)

sethvargo:
Wait wut. It has a hold label. Prrrrooooowwww

PR #275, opened 2020-05-15
Adjust batch boundary conditions so that published keys hit an open batch

### LINKED ISSUE #284: Exports not being created
When we moved to time.Truncate(time.Hour) on the published exposure keys and time.Hour for the export batch period, all exports ceased.

squee1945:
Fixes #284

### DIFF
--- a/internal/api/export/worker.go
+++ b/internal/api/export/worker.go
@@ -53,8 +53,11 @@ func (s *Server) WorkerHandler(w http.ResponseWriter, r *http.Request) {
 			// Fallthrough
 		}
 
+		// Only consider batches that closed a few minutes ago to allow the publish windows to close properly.
+		minutesAgo := time.Now().Add(-5 * time.Minute)
+
 		// Check for a batch and obtain a lease for it.
-		batch, err := s.db.LeaseBatch(ctx, s.config.WorkerTimeout, time.Now())
+		batch, err := s.db.LeaseBatch(ctx, s.config.WorkerTimeout, minutesAgo)
 		if err != nil {
 			logger.Errorf("Failed to lease batch: %v", err)
 			continue
--- a/internal/database/export_test.go
+++ b/internal/database/export_test.go
@@ -390,17 +390,25 @@ func TestKeysInBatch(t *testing.T) {
 
 	// Re-fetch the ExposureBatch by leasing it; this is important to this test which is trying
 	// to ensure our dates are going in-and-out of the database correctly.
-	eb, err := testDB.LeaseBatch(ctx, time.Hour, now)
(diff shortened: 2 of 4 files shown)


internal/api/export/worker.go
```diff
@@ -53,8 +53,11 @@ func (s *Server) WorkerHandler(w http.ResponseWriter, r *http.Request) {
 			// Fallthrough
 		}
 
+		// Only consider batches that closed a few minutes ago to allow the publish windows to close properly.
+		now := time.Now().Add(-5 * time.Minute)
+
 		// Check for a batch and obtain a lease for it.
-		batch, err := s.db.LeaseBatch(ctx, s.config.WorkerTimeout, time.Now())
+		batch, err := s.db.LeaseBatch(ctx, s.config.WorkerTimeout, now)
```
mikehelmick, line 60: `now` is misleading now

internal/database/exposure.go
```diff
@@ -145,12 +145,12 @@ func generateExposureQuery(criteria IterateExposuresCriteria) (string, []interfa
 
 	if !criteria.SinceTimestamp.IsZero() {
 		args = append(args, criteria.SinceTimestamp)
-		q += fmt.Sprintf(" AND created_at > $%d", len(args))
+		q += fmt.Sprintf(" AND created_at >= $%d", len(args))
```
mikehelmick, line 148: Add a comment about why >= and <

this was a subtle bug you found.

crwilcox:
Can you open (or link to) an issue for this pr?

internal/database/export_test.go
```diff
@@ -390,17 +390,25 @@ func TestKeysInBatch(t *testing.T) {
...
 	// to ensure our dates are going in-and-out of the database correctly.
-	eb, err := testDB.LeaseBatch(ctx, time.Hour, now)
+	leased, err := testDB.LeaseBatch(ctx, time.Hour, now)
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	// Lookup the keys; they must be only the key created_at the endTimestamp
-	// (because start is exclusive, end is inclusive).
+	// Check that the batch times from the database are *exactly* what we started with.
```
jba, line 398: Jason, if I understood our conversation last night correctly, the root cause of your problem was not Postgres/Go time mismatch. So I don't think you need these checks, which I suggested while in paranoid mode.

squee1945: Yes agree, but there's no harm in leaving them - e.g., might help someone replatforming this code in the future.

internal/database/exposure.go
squee1945: done.

internal/api/export/worker.go
squee1945: done

PR #278, opened 2020-05-15
Clarify the text on signing the 'incremental files' and distributing public keys for this

mh-:
The original text 
"Sending a public key to devices to digitally sign the incremental files with a private key on the device."
was unclear, it somewhat implied that the mobile devices use the private key that's required for signing the "incremental files".


mh-:
@googlebot I signed it!

PR #288, opened 2020-05-15
Make insert export file idempotent by skipping overwrite.

squee1945:


internal/database/export.go
```diff
@@ -519,6 +519,7 @@ func addExportFile(ctx context.Context, tx pgx.Tx, ef *model.ExportFile) error {
 			(bucket_name, filename, batch_id, region, batch_num, batch_size, status)
 		VALUES
 			($1, $2, $3, $4, $5, $6, $7)
+		ON CONFLICT (filename) DO NOTHING
```
sethvargo, line 522: It's probably fine doing this in the data layer, but we'll lose any visibility and ability to surface this up to the server logs. Thoughts?

squee1945: I can make this change since we're in transaction (add an additional query to lookup existing) and Info/Warn if present. This only happens if a previous batch failed part way through and it is being reprocessed - or if someone is messing around directly with the database.

Note that we also use this technique for exposure keys - we never overwrite and skip silently on conflict.

sethvargo: Got it

PR #292, opened 2020-05-15
fix misspelling

chemidy:
- fix misspelling


crwilcox:
Thanks for the contribution @chemidy

Issue #289, opened 2020-05-15
run VACUUM on postgresql database


mikehelmick:
**Is your feature request related to a problem? Please describe.**

The VACUUM command should be run periodically to ensure that deleted keys are removed from underlying storage and new keys aren't kept consecutively.

```VACUUM FULL VERBOSE```

**Describe the solution you'd like**

New job in `/cmd/vacuum`

Added to terraform config + cloud scheduler to run it.

/kind enhancement
/priority important-soon

jba:
I suggest you look into autovacuum. It makes manual vacuuming unnecessary.

mikehelmick:
looks like it is on by default

https://www.postgresql.org/docs/9.6/runtime-config-autovacuum.html#GUC-AUTOVACUUM

sethvargo:
Is it on by default in Cloud SQL? https://cloud.google.com/sql/docs/postgres/flags

mikehelmick:
I think we want to modify our DDL to enable this at the table level, at minimum on the exposure table.

PR #293, opened 2020-05-15
Enable vacuum on all tables

sethvargo:
Fixes GH-289

/assign @mikehelmick

### DIFF
--- a/migrations/000017_vacuum.down.sql
+++ b/migrations/000017_vacuum.down.sql
@@ -1,0 +1,28 @@
+-- Copyright 2020 Google LLC
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--      http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


migrations/000017_vacuum.down.sql
```diff
@@ -0,0 +1,28 @@
...
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+ALTER TABLE APIConfig SET (autovacuum_enabled = false);
+ALTER TABLE ExportBatch SET (autovacuum_enabled = false);
+ALTER TABLE ExportConfig SET (autovacuum_enabled = false);
+ALTER TABLE ExportFile SET (autovacuum_enabled = false);
+ALTER TABLE Exposure SET (autovacuum_enabled = false);
```
mikehelmick, line 21: I think we need to tune this param lower

https://www.postgresql.org/docs/8.4/runtime-config-client.html#GUC-VACUUM-FREEZE-TABLE-AGE

the default is to wait for up to 150 million transactions - which is many many days

mikehelmick: perhaps others. - @jba - do you have any thoughts?

jba: We seem to use whatever defaults GCP gives us. The only things we change are
```
    {
      name  = "autovacuum_vacuum_cost_limit"
      value = 10000
    },
    {
      name  = "vacuum_cost_limit"
      value = 10000
    }
```

sethvargo: Vacuuming is _very_ expensive and acquires a full table lock. Do we want to do it more frequently?

Issue #287, opened 2020-05-15
transmissionRisk should be per-key on upload not per-patch


mikehelmick:
/kind bug
/assign

mikehelmick:
fixed 

/close

PR #294, opened 2020-05-15
make tranismission risk per-exposure key

mikehelmick:
part of for #287

### DIFF
--- a/internal/model/exposure.go
+++ b/internal/model/exposure.go
@@ -31,7 +31,7 @@ const (
 	keyLength = 16
 
 	// Transmission risk constraints (inclusive..inclusive)
-	minTransmissionRisk = 1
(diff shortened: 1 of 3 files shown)


internal/model/exposure.go
```diff
@@ -200,9 +204,18 @@ func (t *Transformer) TransformPublish(inData *Publish, batchTime time.Time) ([]
 			return nil, fmt.Errorf("interval number %v is in the future, must be < %v", exposureKey.IntervalNumber, maxIntervalNumber)
 		}
 
+		tr := exposureKey.TransmissionRisk
+		// This is temporary - while we reprecate per-request tranismission risk
```
sethvargo, line 208: deprecate*

PR #296, opened 2020-05-15
Emit log if conflict on ExportFile insertion (skipping overwrite).

squee1945:
@sethvargo I guess I was too slow :)

### DIFF
--- a/internal/database/export.go
+++ b/internal/database/export.go
@@ -39,6 +39,9 @@ func (db *DB) AddExportConfig(ctx context.Context, ec *model.ExportConfig) error
 	if ec.Period > oneDay {
 		return errors.New("maximum period is 24h")
 	}
+	if ec.Period == 0 {
+		return errors.New("period must be non-zero")
+	}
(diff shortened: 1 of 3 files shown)


internal/database/export.go
```diff
@@ -512,7 +519,28 @@ func (db *DB) DeleteFilesBefore(ctx context.Context, before time.Time, blobstore
 	return count, nil
 }
 
+// addExportFile adds a row to ExportFile. If the row already exists (based on the primary key),
+// ErrKeyConflict is returned.
 func addExportFile(ctx context.Context, tx pgx.Tx, ef *model.ExportFile) error {
+	row := tx.QueryRow(ctx, `
```
gurayAlsac, line 525: Does the insert below return number of records inserted or anything? Maybe don't need a dedicated query here?

squee1945: Good idea. Let me try.

squee1945: Yup, works. Updated PR.

PR #299, opened 2020-05-15
complete move of transmission risk to per-exposure key

mikehelmick:
/assign @peetsasaki 
/hold

This will be held until Monday the 18th to allow the android reference app to catch up.

Also - awaiting test data from android team to update failing unit tests.


mikehelmick:
PTAL - this now has valid safetynet payloads from the reference android app

mikehelmick:
one more PTAL - tests passing

PR #300, opened 2020-05-15
optional cloudbuild triggers and other terraform improvements

### LINKED ISSUE #272: google_cloudbuild_trigger won't work for everyone
The google_cloudbuild_trigger terraform resources in https://github.com/google/exposure-notifications-server/blob/master/terraform/main.tf create [GitHub app triggers](https://cloud.google.com/cloud-build/docs/automating-builds/create-github-app-triggers) which rebuild and push all the containers to GCR each time master is updated on this repo.

This resource won't work when running from an empty project- you have to first authorize the cloud build app with your github account, and connect it to the project you're running from, which can only be done in the UI. It also requires certain permissions on the repo, so most people will probably not be able to run it.

We have a few options:
1. Don't use cloud build triggers at all in Terraform, and build/deploy the containers separately. In this case we could still do it the first time with Terraform (basically with a basic script that calls out to `ko` a bunch of times) - we just wouldn't get new deployments automatically.
2. Use cloud build triggers, but have a dev/prod flag that turns on whether or not the trigger resource runs. In this case we would add the manual step to our docs as something to be done before running Terraform, and make the GitHub repo into variables so people can connect to their forks if they don't have permission to connect to this one.

Thoughts / other ideas?

cc @ndmckinley @sethvargo

danawillow:
The main change here  is making the creation of the `google_cloudbuild_trigger` resources controlled by a variable `use_build_triggers`. Happy also to rename it to `env` or `environment` if that's better (to use a dev/prod toggle).

Other changes:
* use the correct directories for federation code
* update builders/schema.yaml to use a dockerfile that exists (this doesn't actually succeed, but fails differently from before)
* add a few more depends_ons to make sure APIs are enabled before trying to use them
* use https://cloud.google.com/storage/docs/uniform-bucket-level-access
* use pessimistic constraint operators for versions so we don't accidentally auto-upgrade to new major versions

Fixes #272


danawillow:
cc @sethvargo

terraform/main.tf
```diff
@@ -213,6 +221,11 @@ resource "google_cloudbuild_trigger" "update-schema" {
     "_NAME" : google_sql_database.db.name
     "_SSLMODE" : "disable"
   }
+  depends_on = [google_project_iam_member.cloudbuild-secrets, google_project_iam_member.cloudbuild-sql]
+}
+
+# "build" does first time setup - it is different from "deploy" which we set up to trigger for later.
```
ndmckinley, line 227: Oh, copy-paste error here, this comment applies to the below null resource.

danawillow: Thanks! Fixed.

PR #301, opened 2020-05-15
Allow schema.yaml to work again after changes to docker containers.

ndmckinley:


builders/schema.yaml
```diff
@@ -1,14 +1,15 @@
 steps:
 # Build migrations container as prereq.
 - name: 'gcr.io/cloud-builders/docker'
```
sethvargo, line 3: We actually build and publish a migrate container now `gcr.io/cloud-devrel-public-resources/exposure-notifications/migrate` - do you want to use that?

sethvargo: It includes the Cloud SQL Proxy

ndmckinley: I don't think so - I need the bash script in this one to actually start the cloudsql proxy using the environment variables and to read the secrets text file - you can't do that with environment variables alone as far as I can tell.  The change to this container broke the terraform first-run process which needs to connect to the cloudsql database and run migrate - you can't do both things from cloud build without a script like this one, as far as I can tell.  Unless that container will do both of those things for us.

sethvargo: That container is alpine, so you should be able to:

```yaml
entrypoint: /bin/sh
args: 
- -ec
- |-
    /bin/cloud_sql_proxy -instances=$DB_CLOUDSQL_PATH=tcp:$DB_PORT &
    DB_HOST=localhost
    sleep 5
    /usr/local/bin/migrate up
```

ndmckinley: Ah, I see - okay, I can do that.

qLb: Hi 👋 Any ETA on this one? Temporary workarounds? ..🏜️

ndmckinley: I'll have it today.  :)

ndmckinley: Okay - I do still have to use the container in this project, plus also generally speaking I think it's best to depend on open-source containers in this project, but I was able to address the issues that led to you making this comment - the container is nice and small again.  :) Build times O(seconds), so not worried about it being in any critical paths.

ndmckinley:
Went ahead and rebased on top of dana's changes, so this should be good now.  I'll retest to be sure...

builders/schema.yaml
sethvargo: The container is open source (it's in this repo in `docker/`) and publicly accessible. I'm hesitant to have a _different_ container in dev than prod, since that divergence could be a source of future bugs.

ndmckinley: I agree - but I think the move is to convert prod, though, not to convert dev.  Imagine when people need to make changes to the container (as I had to do here) - seems significantly better for everyone to manage their own version of this container instead of needing to push to prod.  It's relatively low cost to do this this way.

PR #304, opened 2020-05-15
ensure a minimum export file length

### LINKED ISSUE #290: pad daily downloads to ensure minimunm size
**Is your feature request related to a problem? Please describe.**

When generating export batches - generate and shuffle in random keys.
If there are enough keys in the batch, then no additional padding is needed.

/kind enhancement
/priority important-soon

mikehelmick:
This change will pad the export files with a minimum size.

The generated data is random, but uses real data as a basis to ensure that it is realistic. (making the fake keys too old would give them away).

/fixes #290

### DIFF
--- a/internal/api/export/worker.go
+++ b/internal/api/export/worker.go
@@ -116,7 +116,7 @@ func (s *Server) exportBatch(ctx context.Context, eb *model.ExportBatch) error {
 		logger.Infof("No records for export batch %d", eb.BatchID)
 	}
 
-	exposures, err = ensureMinNumExposures(exposures, eb.Region, s.config.MinRecords)
+	exposures, err = ensureMinNumExposures(exposures, eb.Region, s.config.MinRecords, s.config.PaddingRange)
 	if err != nil {
 		return fmt.Errorf("ensureMinNumExposures: %w", err)
 	}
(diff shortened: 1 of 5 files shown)


internal/api/export/worker.go
```diff
@@ -247,3 +254,60 @@ func exportFilename(eb *model.ExportBatch, batchNum int) string {
...
+	return int(n.Int64()) + min, nil
+}
+
+func ensureMinNumExposures(exposures []*model.Exposure, region string, minLength int) ([]*model.Exposure, error) {
+	if len(exposures) == 0 {
+		return nil, fmt.Errorf("cannot pad zero length exposures")
+	}
+	for len(exposures) < minLength {
+		// Pieces needed are
+		// (1) exposure key, (2) inerval number, (3) tranismission risk
```
gurayAlsac, line 273: typo: inerval -> interval

gurayAlsac: tranismission also

internal/api/export/worker.go
```diff
@@ -114,6 +116,11 @@ func (s *Server) exportBatch(ctx context.Context, eb *model.ExportBatch) error {
 		logger.Infof("No records for export batch %d", eb.BatchID)
 	}
 
+	exposures, err = ensureMinNumExposures(exposures, eb.Region, s.config.MinRecords)
```
gurayAlsac, line 119: Is it desirable to prevent inference that the export has been padded? We could add some jitter into the min length arg.

internal/api/export/worker.go
mikehelmick: fixed

internal/api/export/worker.go
mikehelmick: good idea - done.

PR #306, opened 2020-05-15
Allow at most one index file write for empty batch within worker.

### LINKED ISSUE #305: Fast empty batch processing leads to rate limiting on Cloud Storage.
When a single worker is rapidly moving through empty batches, the index file is written for each one. Re-writing an index file on an empty batch is desirable because it facilitates self-healing, but it in unnecessary to re-write the same index file multiple times within the scope of of single worker invocation.

squee1945:
Fix #305

### DIFF
--- a/internal/api/export/worker.go
+++ b/internal/api/export/worker.go
@@ -40,6 +40,7 @@ func (s *Server) WorkerHandler(w http.ResponseWriter, r *http.Request) {
 	defer cancel()
 	logger := logging.FromContext(ctx)
 
+	emitIndexForEmptyBatch := true
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


squee1945:
Unfortunately, this diff looks more complicated than the change really is....

internal/api/export/worker.go
```diff
@@ -145,8 +151,49 @@ func (s *Server) exportBatch(ctx context.Context, eb *model.ExportBatch) error {
 		objectNames = append(objectNames, objectName)
 	}
 
-	// Create the index file. The index file includes _all_ batches for an ExportConfig, so multiple
-	// workers may be racing to update it. We use a lock to make them line up after one another.
+	// Emit the index file if needed.
+	if batchSize > 0 || emitIndexForEmptyBatch {
```
squee1945, line 155: This is the essence of this PR. The rest of the large diff is due to a function extraction.

PR #308, opened 2020-05-15
make truncate window configurable

gurayAlsac:
Eventually there might be a better home for this sort of
cross-cutting global config, but for now, an env var works.

This will allow quicker end-to-end tests of the round trip
of upload a key, wait for it to get into an export batch,
test exposure on another device.

### DIFF
--- a/internal/api/export/batcher.go
+++ b/internal/api/export/batcher.go
@@ -98,7 +98,7 @@ func (s *Server) maybeCreateBatches(ctx context.Context, ec *model.ExportConfig,
 		return 0, fmt.Errorf("fetching most recent batch for config %d: %w", ec.ConfigID, err)
 	}
 
-	ranges := makeBatchRanges(ec.Period, latestEnd, now)
+	ranges := makeBatchRanges(ec.Period, latestEnd, now, s.config.TruncateWindow)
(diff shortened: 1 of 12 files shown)


internal/api/export/batcher.go
```diff
@@ -98,7 +98,7 @@ func (s *Server) maybeCreateBatches(ctx context.Context, ec *model.ExportConfig,
 		return 0, fmt.Errorf("fetching most recent batch for config %d: %w", ec.ConfigID, err)
 	}
 
-	ranges := makeBatchRanges(ec.Period, latestEnd, now)
+	ranges := makeBatchRanges(ec.Period, latestEnd, now, s.config.TruncateWindow)
```
sethvargo, line 101: I have an informal, unwritten, totally-from-my-experience-of-running-go-apps-at-scale rule that once function arity reaches 4 inputs, it should either be restructured into smaller methods or accept an input/request struct:

```go
type MakeBatchRangesInput struct {
  Period string,
  # ...
}

func makeBatchRanges(i *MakeBatchRangesInput) {}
```

gurayAlsac: Filed https://github.com/google/exposure-notifications-server/issues/312

Issue #35, opened 2020-05-01
Separate packages, include internal/model to minimize exported fields.


mikehelmick:
@crwilcox plan
- [x] Move all packages under api/ to internal/package-name
- [x] Combine database and model -> model to avoid dependency cycles
- [ ] don't extend DB, instead pass in for use by model specific methods

crwilcox:
Adding some information.
The goal here is to separate out the models/db to packages so that things rare more componentized and to improve readability.

crwilcox:
Currently the structure is roughly this:
```
cmd /
-- component-1 /
---- main.go
-- component-2 /
---- main.go
-- component-3 /
---- main.go

internal /
-- model /
---- component-1.go
---- component-2.go
---- component-3.go
-- database /
---- component-1.go
---- component-2.go
---- component-3.go
```

After refactoring I expect to have
```
cmd /
-- component-1 /
---- main.go
-- component-2 /
---- main.go
-- component-3 /
---- main.go

internal /
-- database /
---- shared-database-components.go (database.go)
-- component-1 /
--- api /
------ component-1.go
---- model.go
---- database.go
```

This will allow less exports and should keep related components nearer each other.

crwilcox:
#407 and #410. 
Considering this work complete.

PR #309, opened 2020-05-15
Restructure api directories to be under internal matching package name

crwilcox:
First step towards #35


cmd/cleanup-export/main.go
```diff
@@ -20,7 +20,7 @@ import (
 	"log"
 	"net/http"
 
-	"github.com/google/exposure-notifications-server/internal/api/cleanup"
+	cleanup "github.com/google/exposure-notifications-server/internal/cleanup"
```
mikehelmick, line 23: the leading cleanup shouldn't be needed

PR #314, opened 2020-05-16
Add support for Azure KeyVault and HashiCorp Vault interfaces

sethvargo:
This adds two new secret providers: KeyVault and HashiCorp Vault

It also abstracts the caching into another SecretManager which wraps any SecretManager implementation.


internal/secrets/azure_keyvault.go
```diff
@@ -0,0 +1,84 @@
+// Copyright 2020 Google LLC
```
crwilcox, line 1: Do we have a way to test these implementations?

sethvargo: We can test the Vault one fairly easily since it's a standard HTTP server. I don't have a good idea of testing the Azure one without uploading Azure keys. We could do it (e.g. store them in Secret Manager and retrieve them on presubmit), but I'd like to get approval on this general approach before going down that rabbit hole of testing.

crwilcox: Makes sense. This starts to take us down the road of having multiple implementations also. Possible this could be done as an example instead of a hooked up implementation? I agree showing how this works and how it can be extended makes sense.

sethvargo: What do you mean by "example"? 😄

internal/secrets/hashicorp_vault.go
```diff
@@ -0,0 +1,86 @@
...
+	return sm, nil
+}
+
+// GetSecretValue implements the SecretManager interface. Secrets are specified
+// as the path to the secret in Vault. Secrets are expected to have the string
+// value for the secret in a key named "value".
+//
+// For example:
+//
+//     /secret/data/my-secret?version=5 #=> { "value": "dajkfl32ip2" }
```
mikehelmick, line 52: hey - that’s my password!

mikehelmick:
@googlebot resca

mikehelmick:
@googlebot rescan

PR #336, opened 2020-05-18
minor simplifications

jba:
Simplify some code and make it more idiomatic.


internal/federationout/federationout.go
```diff
@@ -298,20 +298,16 @@ func rawToken(ctx context.Context) (string, error) {
 
 func intersect(aa, bb []string) []string {
 	if len(aa) == 0 || len(bb) == 0 {
-		return []string{}
+		return nil
 	}
-	result := []string{}
```
sethvargo, line 303: This is actually a behavior change. The result will be `nil` instead of `[]string{}` on not found.

internal/federationin/federationin.go
```diff
@@ -48,9 +48,11 @@ var (
 	fetchBatchSize = database.InsertExposuresBatchSize
 )
 
-type fetchFn func(context.Context, *pb.FederationFetchRequest, ...grpc.CallOption) (*pb.FederationFetchResponse, error)
-type insertExposuresFn func(context.Context, []*model.Exposure) error
-type startFederationSyncFn func(context.Context, *model.FederationInQuery, time.Time) (int64, database.FinalizeSyncFn, error)
+type (
```
sethvargo, line 51: I guess this makes sense, but I've personally never seen it in any Go for types (common on `var` and `const`). Is this common?

internal/federationout/federationout.go
jba: True, but they are almost indistinguishable. `nil` is a bit more idiomatic.

internal/federationin/federationin.go
jba: No, groups with `type` are relatively rare. In this case I felt it looked better, because each of the function types is a single line.

PR #338, opened 2020-05-18
internal/database: defer resetTestDB

jba:
The existing code failed to reset the DB if t.Fatal was called.


internal/database/authorized_app_test.go
```diff
@@ -167,6 +167,7 @@ func TestGetAuthorizedApp(t *testing.T) {
 				t.Fatal(err)
 			}
 			defer conn.Release()
+			defer resetTestDB(t)
```
sethvargo, line 170: Separate from this, I'm curious your thoughts on `defer` vs `t.Cleanup()`

PR #340, opened 2020-05-18
Terraform docs move and improvement

crwilcox:
Holding as we walk through this.


crwilcox:
Holding for https://github.com/google/exposure-notifications-server/pull/341/files to avoid stepping on one another.

Issue #214, opened 2020-05-13
terraform: ensure all service accounts can read secrets


mikehelmick:

ndmckinley:
As far as I know, they already can - that's why I haven't picked this up - but once we have integration testing we'll know for sure.  :)

ndmckinley:
👍 Mine succeed, but who knows why.  I'll figure that out.
/assign @ndmckinley

radisef347:
![alt text](https://cdni.pornpics.com/1280/7/51/81056079/81056079_141_f080.jpg)

stati0n:
I received the same error:

```
2020-05-15 21:10:47.624 EDTsetup.Setup: error loading environment variables: failed to resolve "projects/623414997131/secrets/dbServerCA/versions/1": failed to access secret projects/623414997131/secrets/dbServerCA/versions/1: rpc error: code = PermissionDenied desc = Permission 'secretmanager.versions.access' denied for resource 'projects/623414997131/secrets/dbServerCA/versions/1' (or it may not exist).
```

yuliumitre:
My cloud-run services can't be started during terraform deployment.  It seems related to the secrete. Any thoughts? 

Here are the error from the GCP logs: 

setup.Setup: error loading environment variables: failed to resolve "projects/624788706964/secrets/dbClientKey/versions/1": failed to access secret projects/624788706964/secrets/dbClientKey/versions/1: rpc error: code = PermissionDenied desc = Permission 'secretmanager.versions.access' denied for resource 'projects/624788706964/secrets/dbClientKey/versions/1' (or it may not exist).

PR #341, opened 2020-05-18
Update Terraform setup from Mike's journey

sethvargo:
- Fixes GH-214

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -73,8 +73,8 @@ resource "google_sql_database_instance" "db-inst" {
   name             = "contact-tracing-${random_string.db-name.result}"
   settings {
     tier              = var.cloudsql_tier
+    disk_size         = var.cloudsql_disk_size_gb
     availability_type = "REGIONAL"
-    disk_size         = 500
     backup_configuration {
       enabled    = true
       start_time = "02:00"
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -544,6 +544,12 @@ resource "google_cloud_run_service_iam_policy" "federationin-noauth" {
   policy_data = data.google_iam_policy.noauth.policy_data
 }
 
+# Cloud Scheduler requires AppEngine projects!
+resource "google_app_engine_application" "app" {
+  project     = data.google_project.project.project_id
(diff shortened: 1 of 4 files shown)


terraform/main.tf
```diff
@@ -352,11 +356,40 @@ resource "google_cloud_run_service" "exposure" {
   depends_on = [null_resource.submit-build-and-publish, google_project_service.services["run.googleapis.com"], google_project_service.services["sqladmin.googleapis.com"]]
 }
 
+resource "google_service_account" "export" {
+  project      = data.google_project.project.project_id
+  account_id   = "en-export-sa"
```
mikehelmick, line 361: this SA is missing secret accessor

terraform/main.tf
```diff
@@ -352,11 +356,40 @@ resource "google_cloud_run_service" "exposure" {
...
+
+  secret_id = google_secret_manager_secret.db-pwd.id
+  role      = "roles/secretmanager.secretAccessor"
+  member    = "serviceAccount:${google_service_account.export.email}"
+}
+
+resource "google_storage_bucket_iam_member" "export-objectadmin" {
+  bucket = google_storage_bucket.export.name
+  role   = "roles/storage.objectAdmin" // overwrite is not included in objectCreator
+  member = "serviceAccount:${google_service_account.exposure.email}"
```
mikehelmick, line 382: this should be : export.email

terraform/main.tf
```diff
@@ -82,7 +72,7 @@ resource "google_sql_database_instance" "db-inst" {
   database_version = "POSTGRES_11"
   name             = "contact-tracing-${random_string.db-name.result}"
   settings {
-    tier              = "db-custom-1-3840" # "db-custom-32-122880"
+    tier              = var.cloudsql_tier
     availability_type = "REGIONAL"
     disk_size         = 500
```
crwilcox, line 77: could you add `var.cloudsql_disk_size`?

sethvargo: Done

terraform/main.tf
sethvargo: Fixed

PR #343, opened 2020-05-18
log effective environment config on startup. Omit DB password

mikehelmick:


internal/database/config.go
```diff
@@ -37,3 +40,18 @@ type Config struct {
...
+
+// String returns the string representation of the database connection config.
+// This omits the Password field to prevent accidental logging.
+func (c *Config) String() string {
+	pwSet := "<set>"
+	if c.Password == "" {
+		pwSet = "<not set>"
+	}
+
+	return fmt.Sprintf("{Name:%v User:%v Host:%v Port:%v SSLMode:%v ConnectionTimeout:%v Password:%v SSLCertPath:%v SSLKeyPath:%v SSLRootCertPath:%v PoolMinConnections:%v PoolMaxConnections:%v PoolMaxConnLife: %v PoolMaxConnIdle:%v PoolHealthCheck:%v}",
```
sethvargo, line 52: This feels error-prone. Could we serialized to json instead and use `json:"-"` on the Password field?

mikehelmick: change approach - copy in setup.go and just reset the field in the copy.

PR #346, opened 2020-05-18
Fix up auth, recurring diff

sethvargo:


terraform/main.tf
```diff
@@ -15,13 +15,12 @@ provider "random" {}
 # This is to ensure that the project / gcloud auth are correctly configured.
 resource "null_resource" "gcloud_check" {
   provisioner "local-exec" {
-    command = "gcloud projects describe ${var.project} || (echo 'please sign in to gcloud using `gcloud auth login`.' && exit 1)"
+    command = "command -v gcloud &>/dev/null || (echo 'Please install and authenticate with gcloud!' && exit 127)"
```
ndmckinley, line 18: This isn't going to confirm that the user is authenticated, that ok with you?

ndmckinley: I'd leave it the way it was, but if you prefer it this way then I'm cool with that - just change the comment to reflect the changes.

sethvargo: Chris fronted a bunch of auth instructions in the README now. I think that's mine. I'd also be okay removing this entirely if you think it's unnecessary.

Since most things depend on the project data source, the configs will fail _pretty_ early on bad permissions.

ndmckinley: They'll fail early on bad ADCs, but late on bad gcloud auth.  Given that, I'd leave it, but if you just change the comment I'll approve.  (I'd approve now and let you decide what to do, but prow would automerge and I think you need to make one of the two changes).

PR #347, opened 2020-05-18
Combine database and model packages

crwilcox:
Towards #35. After this is merged I can begin to separate the db interface without creating dependency cycles.


mikehelmick:
> I'm not sure I understand the need for this change, but I'm not adverse to it either.

this is an intermediate state only

PR #348, opened 2020-05-18
change federation fields/types

### LINKED ISSUE #267: Remove verification_authority_name field
Remove from
model/exposure.go
remove from federation in/out

### LINKED ISSUE #270: federation clean up
- Move transmission risk to an integer
- remove verification authority

mikehelmick:
Remove verification authority name from DB and federation
Change transmission risk to int instead of enum in federation

fixes #270 
fixes #267


mikehelmick:
/hold

for @squee1945 to review

internal/pb/federation.proto
```diff
@@ -39,16 +39,8 @@ message ContactTracingResponse {
 }
 
 message ContactTracingInfo {
-	// TODO(squee1945 | mikehelmick): this should be an int now.
-	TransmissionRisk transmissionRisk = 1;  // required
+	int32 transmissionRisk = 1;  // required
```
squee1945, line 42: Why did you remove the enum? Having it as an enum makes the values self-describing.

squee1945: [starting review]

mikehelmick: because transmission risk has moved to an integer, between 1-8 where the semantic meaning is not handled by the server

mikehelmick:
> Can you add a comment in the proto indicating acceptable values? Can we add validation somewhere?

done.

PR #351, opened 2020-05-18
fix locking error in appconfig caching

mikehelmick:


internal/authorizedapp/database_provider.go
```diff
@@ -71,35 +71,41 @@ func NewDatabaseProvider(ctx context.Context, db *database.DB, config *Config, o
 	return provider, nil
 }
 
-// AppConfig returns the config for the given app package name.
-func (p *DatabaseProvider) AppConfig(ctx context.Context, name string) (*database.AuthorizedApp, error) {
-	logger := logging.FromContext(ctx)
-
+func (p *DatabaseProvider) checkCache(name string) (bool, *database.AuthorizedApp, error) {
```
sethvargo, line 74: Nits:

- Drop a note that the bool returns true on hit, false otherwise
- I think Go style would want `model, bool, error` to match the syntax of a map lookup

mikehelmick: done.

Issue #269, opened 2020-05-14
Allow AuthorizedApp to accept iOS and Android


mikehelmick:
Allow for AuthorizedApp to dual home android and ios in the same row

(same bundle ID / app package name)

/kind enhancement

mikehelmick:
/assign

Fixing this in tandem with #253

mikehelmick:
/assign @sethvargo 

fixed, correct?

sethvargo:
Gonna propose one more change to clarify which columns are for android and which are ios.

PR #352, opened 2020-05-19
Rename/prefix SafetyNet columns

sethvargo:
This updates the `AuthorizedApp` table to prefix the SafetyNet-related columns with Android and updates references.

Per discussion, it also changes the behavior of the `allowed_regions` field and drops the `all_regions` boolean in favor of the follow behavior:

- The list cannot be null (database constraint)
- If the list is empty, all regions are allowed
- If the list contains any values, only exposures from that region are allowed

Fixes GH-269


migrations/000022_authorized_app_prefixes.up.sql
```diff
@@ -0,0 +1,28 @@
...
+
+ALTER TABLE AuthorizedApp RENAME COLUMN apk_digest TO safetynet_apk_digest;
+ALTER TABLE AuthorizedApp RENAME COLUMN cts_profile_match TO safetynet_cts_profile_match;
+ALTER TABLE AuthorizedApp RENAME COLUMN basic_integrity TO safetynet_basic_integrity;
+ALTER TABLE AuthorizedApp RENAME COLUMN allowed_past_seconds TO safetynet_past_seconds;
+ALTER TABLE AuthorizedApp RENAME COLUMN allowed_future_seconds TO safetynet_future_seconds;
+ALTER TABLE AuthorizedApp RENAME COLUMN ios_devicecheck_team_id_secret TO devicecheck_team_id_secret;
+ALTER TABLE AuthorizedApp RENAME COLUMN ios_devicecheck_key_id_secret TO devicecheck_key_id_secret;
+ALTER TABLE AuthorizedApp RENAME COLUMN ios_devicecheck_private_key_secret TO devicecheck_private_key_secret;
+ALTER TABLE AuthorizedApp DROP COLUMN all_regions;
```
mikehelmick, line 25: this is a potentially breaking change - This migration should probably start w/ something like

UPDATE AuthorizedApp SET allowed_regions=ARRAY[] WHERE all_regions=true

same for the downgrade.

testdata/config.sql
```diff
@@ -1,6 +1,29 @@
...
-  ('com.example.ios.app', 'ios', false, false, 60, 60, ARRAY ['GB','US'], false, 'projects/38554818207/secrets/ios-devicecheck-team-id/versions/1', 'projects/38554818207/secrets/ios-devicecheck-key-id/versions/1', 'projects/38554818207/secrets/ios-devicecheck-private-key/versions/1'),
-  ('com.example.android.app', 'android', false, false, 60, 60, ARRAY ['GB','US'], false, NULL, NULL, NULL);
+INSERT INTO AuthorizedApp (
+  app_package_name, platform, allowed_regions,
+  safetynet_cts_profile_match, safetynet_basic_integrity, safetynet_past_seconds, safetynet_future_seconds,
+  devicecheck_team_id_secret, devicecheck_key_id_secret, devicecheck_private_key_secret
+) VALUES (
+  'com.example.ios.app', 'ios', ARRAY[]::VARCHAR[],
+  false, false, 60, 60,
+  'projects/38554818207/secrets/ios-devicecheck-team-id/versions/1', 'projects/38554818207/secrets/ios-devicecheck-key-id/versions/1', 'projects/38554818207/secrets/ios-devicecheck-private-key/versions/1'
```
mikehelmick, line 24: project specific data.

migrations/000022_authorized_app_prefixes.up.sql
sethvargo: Done

testdata/config.sql
sethvargo: This was already here (see above). Since this is raw SQL, I don't think we can accept an envvar or anything. Can we leave as-is (since it's already like this on master) and create a follow-up issue for a more reliable seed script?

mikehelmick: it's not usable to anyone else - fine w/ separate PR, but if people apply this they'll get errors

sethvargo: Created an issue to follow up

Issue #273, opened 2020-05-14
Tool to generate export files


mikehelmick:
Write a tool to generate export files from the command line.

Shouldn't require a database and should be able to take keys from input or generate random.

PR #353, opened 2020-05-19
add tool to generate export files

gurayAlsac:
This doesn't yet take an input of keys. Plan to handle that in subsequent PR.


internal/util/generate.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func RandomTransmissionRisk() int {
+	return RandomInt(maxTransmissionRisk) + 1
+}
+
+func RandomArrValue(arr []string) string {
+	return arr[RandomInt(len(arr))]
+}
+
```
mikehelmick, line 57: there’s a good bit of overlap w/ this and the export padding

now that has different constraints because we’re trying to use realistic interval data and tranismission risk groupings, so maybe there’s nothing there to reuse.

internal/util/generate.go
```diff
@@ -0,0 +1,87 @@
...
+	"log"
+	"math/big"
+	"time"
+
+	"github.com/google/exposure-notifications-server/internal/database"
+)
+
+const (
+	// the length of a diagnosis key, always 16 bytes
+	dkLen               = 16
```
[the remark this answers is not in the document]
gurayAlsac: Agreed. Might be worth a dedicated pull request as its spread to a number of places at the moment:

```
$ find . -name '*.go' | xargs grep -R -E '.*=.*16' 
...
./internal/util/generate.go:    dkLen               = 16
./internal/database/exposure_model_test.go:     key := make([]byte, 16)
./internal/database/exposure_model.go:  KeyLength = 16
./tools/hexconvert/main.go:     if len(bytes) != 16 {
./tools/hexconvert/main.go:             log.Fatalf("decoded hex string, want len(bytes)=16, got: %v", len(bytes))
./tools/exposure-client/main.go:        dkLen               = 16
```

internal/util/generate.go
gurayAlsac: Right, subtly different ways to generate data. I've at the very least gathered some of the extractable util functions that export padding uses into the new util.

gurayAlsac:
fixes #273

PR #355, opened 2020-05-19
Adding default exportconfig to the dbseed.

mgulimonov:


testdata/config.sql
```diff
@@ -1,6 +1,9 @@
...
+INSERT INTO authorizedapp (app_package_name, platform, cts_profile_match, basic_integrity, allowed_past_seconds, allowed_future_seconds, allowed_regions, all_regions, ios_devicecheck_team_id_secret, ios_devicecheck_key_id_secret, ios_devicecheck_private_key_secret) VALUES
   ('com.example.ios.app', 'ios', false, false, 60, 60, ARRAY ['GB','US'], false, 'projects/38554818207/secrets/ios-devicecheck-team-id/versions/1', 'projects/38554818207/secrets/ios-devicecheck-key-id/versions/1', 'projects/38554818207/secrets/ios-devicecheck-private-key/versions/1'),
   ('com.example.android.app', 'android', false, false, 60, 60, ARRAY ['GB','US'], false, NULL, NULL, NULL);
+
+INSERT INTO exportconfig (filename_root, period_seconds, from_timestamp, thru_timestamp, region, bucket_name, signature_info_ids) VALUES
+  ('exposureKeyExport-US', 300, '2020-05-01 07:00:00.000000', '2025-05-01 07:00:00.000000', 'US', 'apollo-public-bucket', '{1,2}');
```
mikehelmick, line 9: this file contains too much individual project specific things (project number + bucket name)

mgulimonov: This is true. However this is a test data, so I'm not sure if we should do anything about it...

mikehelmick: if the test data isn't for general consumption, I'd rather not have it on the repo.

testdata/config.sql
```diff
@@ -27,3 +27,8 @@ INSERT INTO AuthorizedApp (
   false, false, 60, 60,
   NULL, NULL, NULL
 );
+
+INSERT INTO exportconfig (filename_root, period_seconds, from_timestamp,
+                          thru_timestamp, region, bucket_name,
+                          signature_info_ids)
+VALUES ('exposureKeyExport-US', 300, '2020-05-01 07:00:00.000000', '2025-05-01 07:00:00.000000', 'US', 'apollo-public-bucket', '{1, 2}');
```
mikehelmick, line 34: this references signature info records that aren't in this file.

testdata/config.sql
mgulimonov: Ok, I can close this PR for now. However we can apply the same logic to the authorizedapp insert statement as well...
Should we delete this file completely?

mikehelmick: yeah - I’ll fill an issue to create a DB seeding tool

mikehelmick:
Do we think we still need this?

we've made it easier to to configure the system w/ the admin console. Is that better than throwing in random/seed data.

mgulimonov:
Ok, I think we can close this one.

PR #358, opened 2020-05-19
Moving util methods to the enclient lib.

mgulimonov:
These methods will be reused for integration testing.

### DIFF
--- a/testing/enclient/enclient.go
+++ b/testing/enclient/enclient.go
@@ -6,12 +6,12 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"math/big"
 	"net/http"
 	"time"
 
-	"github.com/google/exposure-notifications-server/internal/base64util"
 	"github.com/google/exposure-notifications-server/internal/database"
 )
 
@@ -19,6 +19,8 @@ const (
 	dkLen               = 16
 	maxTransmissionRisk = 8
 	maxIntervals        = 144
+	// httpTimeout is the maximum amount of time to wait for a response.
+	httpTimeout = 10 * time.Second
 )
 
 type Interval int32
@@ -32,7 +34,7 @@ func PostRequest(url string, data interface{}) (*http.Response, error) {
 		return nil, err
 	}
 	r.Header.Set("Content-Type", "application/json")
-	client := &http.Client{}
+	client := &http.Client{Timeout: httpTimeout}
 	resp, err := client.Do(r)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


testing/enclient/enclient.go
```diff
@@ -0,0 +1,149 @@
...
+	maxTransmissionRisk = 8
+	maxIntervals        = 144
+)
+
+type Interval int32
+
+// Posts requests to the specified url.
+// This methods attempts to serialize data argument as a json.
+func PostRequest(url string, data interface{}) (*http.Response, error) {
+	request := bytes.NewBuffer(JsonRequest(data))
```
sethvargo, line 29: Set a timeout on this request please

testing/enclient/enclient.go
```diff
@@ -0,0 +1,149 @@
...
+	r.Header.Set("Content-Type", "application/json")
+	client := &http.Client{}
+	resp, err := client.Do(r)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return resp, fmt.Errorf("post request failed with status: %v", resp.StatusCode)
```
sethvargo, line 43: You might want to copy the body here and put it in the error message. See devicecheck.go for an example.

testing/enclient/enclient.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	return resp, nil
+}
+
+// Serializes the given argument to json.
+func JsonRequest(data interface{}) []byte {
+	jsonData, err := json.Marshal(data)
+	if err != nil {
+		log.Fatalf("unable to marshal json payload")
```
sethvargo, line 53: It's _probably_ fine since this is in `testing/`, but you'd usually want to return this error back to its caller, at which point you might as well move the `json.Marshal` call into the body.

testing/enclient/enclient.go
```diff
@@ -0,0 +1,149 @@
...
+	return ToBase64(RandomBytes(dkLen))
+}
+
+// Encodes bytes array to base64.
+func ToBase64(key []byte) string {
+	return base64.StdEncoding.EncodeToString(key)
+}
+
+// Decodes base64 string to []byte.
+func DecodeKey(b64key string) []byte {
```
sethvargo, line 143: Please use base64util instead (its an internal package here)

tools/exposure-client/main.go
```diff
@@ -110,76 +91,24 @@ func main() {
...
-		// Adjust interval math for next key.
-		intervalCount = randIntervalCount()
-		intervalNumber -= intervalCount
-	}
-	return exposureKeys
-}
-
-func sendRequest(jsonData []byte) {
-	r, err := http.NewRequest("POST", *url, bytes.NewBuffer(jsonData))
+func sendRequest(data interface{}) {
```
sethvargo, line 106: Same note, you probably want to bubble this error up to the top caller.

testing/enclient/enclient.go
mgulimonov: Ack. Leaving it like this to simplify code of the tests.

testing/enclient/enclient.go
```diff
@@ -0,0 +1,156 @@
...
+
+	"github.com/google/exposure-notifications-server/internal/database"
+)
+
+const (
+	dkLen               = 16
+	maxTransmissionRisk = 8
+	maxIntervals        = 144
+	// httpTimeout is the maximum amount of time to wait for a response.
+	httpTimeout = 10 * time.Second
```
mikehelmick, line 23: 10 might not be enough w/ request tunneling + cold starts - I've seen a realistic 8 in some circumstances.

PR #361, opened 2020-05-19
Authorized App through interface

crwilcox:
Towards #35, break up DB into per table interfaces.


internal/database/authorized_app.go
```diff
@@ -25,10 +25,20 @@ import (
...
+	return &AuthorizedAppDB{
+		db: db,
+	}
+}
+
 // GetAuthorizedApp loads a single AuthorizedApp for the given name. If no row
 // exists, this returns nil.
-func (db *DB) GetAuthorizedApp(ctx context.Context, sm secrets.SecretManager, name string) (*AuthorizedApp, error) {
-	conn, err := db.pool.Acquire(ctx)
+func (db *AuthorizedAppDB) GetAuthorizedApp(ctx context.Context, sm secrets.SecretManager, name string) (*AuthorizedApp, error) {
```
crwilcox, line 40: it seems it may be a convention in Go to call these 'Repositories'. So to avoid the double `db.db` I could name this `repo *AuthorizedAppRepository`.

sethvargo: "Repo" is already overloaded. Can we still call them models? `AuthorizedAppModel`?

crwilcox: PTAL.

crwilcox:
/hold
Would like feedback from @mikehelmick as well before merging.

internal/database/authorized_app.go
```diff
@@ -25,10 +25,20 @@ import (
 	pgx "github.com/jackc/pgx/v4"
 )
 
+type AuthorizedAppModel struct {
```
mikehelmick, line 28: I Would move these out to a sub-package first

that would allow for better naming, like

authorizedappdb.Model

crwilcox: Thanks! I had a sync with jba about this. I also got further notes from xtof. 

- Renaming back to AuthorizedAppDB
- Subpackages as noted
- Possibly at the end moving common db to be internal only, but let's not get ahead of things.

internal/android/safetynet.go
```diff
@@ -24,8 +24,8 @@ import (
 	"runtime/trace"
 	"time"
 
+	authorizedappmodel "github.com/google/exposure-notifications-server/internal/authorizedapp/model"
```
mikehelmick, line 27: no need to alias here - can just use model

Issue #197, opened 2020-05-12
Standardize API responses


squee1945:
As discussed briefly in #195 
cc @sethvargo

sethvargo:
Adding more context here. Right now, across the entire project, we have HTTP handlers that:

- Log an error that's wrapped/formatted
- Publish a metric (usually)
- Write a status (usually 200, even for failures)
- Return

The current problems are:

- If we forget to return, the handler continues on
- If we don't log, debugging is a challenge
- Sometimes we do use a non-200 status (e.g. for federation)

A potentially better architecture would be to reduce the in-function requirement to:

- Publish a metric
- Return an error or nil

```go
func handler(w http.ResponseWriter, r *http.Request) error {
  if !foo {
    return fmt.Errorf("not foo")
  }

  if err := thing(); err != nil {
    return fmt.Errorf("thing: %w", err)
  }

  if err := otherThing(); err != nil {
    return ErrorWithStatus(422, fmt.Errorf("thing: %w", err))
  }

  w.WriteHeader(200)
  w.Write([]byte("ok"))
}
```

Then, upstream in `ServerHTTP`:

```go
func ServerHTTP(w http.ResponseWriter, r *http.Request) {
  if err := handler(w, r); err != nil {
    logger.Debugf(err)

    var statusErr StatusError
    if errors.As(err, &statusErr) {
      http.Error(statusErr.Code(), statusErr.Text())
      return
    }

    w.WriteHeader(200)
  }
}
```

Not a totally flushed out design, but hopefully this describes my thinking a bit more

radisef347:
![alt text](https://cdni.pornpics.com/1280/7/51/81056079/81056079_141_f080.jpg)

sethvargo:
We could also roll up https://github.com/google/exposure-notifications-server/issues/350 here, to provide standardized debug responses vs production responses

PR #366, opened 2020-05-19
create debug mode for exposure notifications upload so developers can…

### LINKED ISSUE #350: Implement DEBUG mode for API
Allow for API responses that convey the actual error

/assign
/kind enhancement

mikehelmick:
… see error messages for testing

fixes #350 
part of #197 - doesn't address federation API responses - since those are gRPC, probably want to handle separate.

### DIFF
--- a/internal/publish/publish.go
+++ b/internal/publish/publish.go
@@ -66,11 +66,11 @@ type publishHandler struct {
 }
 
 type response struct {
-	status          int
-	message         string
-	metric          string
-	count           int // metricCount
-	errorInNonDebug bool
+	status      int
+	message     string
+	metric      string
+	count       int // metricCount
+	errorInProd bool
 }
 
 func (h *publishHandler) handleRequest(w http.ResponseWriter, r *http.Request) response {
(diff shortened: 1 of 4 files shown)


internal/publish/publish.go
```diff
@@ -65,21 +65,25 @@ type publishHandler struct {
 	authorizedAppProvider authorizedapp.Provider
 }
 
-// There is a target normalized latency for this function. This is to help prevent
-// clients from being able to distinguish from successful or errored requests.
-func (h *publishHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+type response struct {
+	status          int
+	message         string
+	metric          string
```
gurayAlsac, line 71: It might be worth separating the metric incrementing from the response object. handleRequest() might want to increment multiple metrics during its work, and there doesn't seem to be too much advantage in deferring this to be the responsibility of the caller.

internal/publish/publish.go
```diff
@@ -65,21 +65,25 @@ type publishHandler struct {
...
+	count           int // metricCount
+	errorInNonDebug bool
```
gurayAlsac, line 73: naming nit: could put this in the affirmative as "errorInProd" to avoid double negatives.

mikehelmick: good call.

internal/publish/publish.go
mikehelmick: there are still other metrics that can be incremented - but this centralizes all of the end of response metrics.

sethvargo:
❤️

PR #375, opened 2020-05-19
Update proto comments

mikehelmick:
/hold 

for all reviewers

### DIFF
--- a/internal/pb/export/export.proto
+++ b/internal/pb/export/export.proto
@@ -47,7 +47,9 @@ message TemporaryExposureKeyExport {
 message SignatureInfo {
   // Apple App Store Application Bundle ID
   optional string app_bundle_id = 1;
-  // Android App package name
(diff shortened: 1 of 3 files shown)


internal/pb/export/export.proto
```diff
@@ -47,7 +47,9 @@ message TemporaryExposureKeyExport {
 message SignatureInfo {
   // Apple App Store Application Bundle ID
   optional string app_bundle_id = 1;
-  // Android App package name
+  // Android App package name, concatenated with the sha256 hash of the app’s
```
gurayAlsac, line 50: This is actually no longer the case. This will be just the android package. Version 1.4 of https://static.googleusercontent.com/media/www.google.com/en//covid19/exposurenotifications/pdfs/Exposure-Key-File-Format-and-Verification.pdf properly reflects this.

Issue #377, opened 2020-05-19
Rename database in Terraform


sethvargo:
The current Terraform database name uses `contact-tracing`, which is less preferred to `exposure-notification` or `en`. Since this renames the Cloud SQL database, we need to coordinate with a production update.

Issue #373, opened 2020-05-19
Terraform: GCS bucket not public


mikehelmick:
The GCS bucket that gets created should have 'allUsers' in the viewers permissions.

/kind bug

Issue #372, opened 2020-05-19
Terraform: en-export-sa@ missing permissions


mikehelmick:
The en-export-sa SA that is created is missing the `Cloud KMS CryptoKey Signer/Verifier` permission

/kind bug

Issue #368, opened 2020-05-19
Terraform: Cloud Run service specs are too low


mikehelmick:
/kind bug

For all jobs - we want to specify more than the default resource allocation

CPU: 2.0
RAM: 2gb for export server, 1gb for rest

PR #381, opened 2020-05-20
Incorporate Terraform changes

sethvargo:
- Fixes GH-377
- Fixes GH-373
- Fixes GH-372
- Fixes GH-368

### DIFF
--- a/terraform/database.tf
+++ b/terraform/database.tf
@@ -23,7 +23,7 @@ resource "google_sql_database_instance" "db-inst" {
   project          = data.google_project.project.project_id
   region           = var.region
   database_version = "POSTGRES_11"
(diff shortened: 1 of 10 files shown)


terraform/database.tf
```diff
@@ -23,7 +23,7 @@ resource "google_sql_database_instance" "db-inst" {
   project          = data.google_project.project.project_id
   region           = var.region
   database_version = "POSTGRES_11"
-  name             = "contact-tracing-${random_string.db-name.result}"
+  name             = "en-${random_string.db-name.result}"
```
sethvargo, line 26: A note on this... apparently Cloud SQL names cannot be changed after they are created.

PR #383, opened 2020-05-20
Remove requirement for DC key_id and team_id to be secrets

sethvargo:
I've done a lot of reading and people generally don't consider these secrets.

### DIFF
--- a/migrations/000023_dc_less_secrets.up.sql
+++ b/migrations/000023_dc_less_secrets.up.sql
@@ -1,0 +1,20 @@
+-- Copyright 2020 Google LLC
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--      http://www.apache.org/licenses/LICENSE-2.0
+--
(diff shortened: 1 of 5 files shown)


migrations/000023_dc_less_secrets.up.sql
```diff
@@ -0,0 +1,20 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+ALTER TABLE AuthorizedApp RENAME COLUMN devicecheck_team_id_secret TO devicecheck_team_id;
```
sethvargo, line 17: I don't think we have any in production. I'm not sure how to write a truly backwards-compat migration in pure SQL...

mikehelmick: it’s a straight rename.

the field is being interpreted differently for sure, but I think this is OK.

Issue #369, opened 2020-05-19
Docs for setting up DeviceCheck


sethvargo:
We need instructions for 

- [x] How to get DeviceCheck credentials
- [ ] How to store them in Secret Manager
- [ ] How to share access w/ our service account
- [ ] how to share resource IDs w/ server operator

PR #384, opened 2020-05-20
Add docs on sharing DC with 3ps

sethvargo:
Fixes GH-369


llatif:
Going through this diff and the rest of the file with edits. Putting do not merge until I get those edits pushed to this branch.

Issue #376, opened 2020-05-19
Create a way to rebuild and deploy images


sethvargo:
Terraform creates the initial deployment, but we need a way to build and deploy new images after initial provision.

Issue #360, opened 2020-05-19
Potential issue with using `--no-traffic` in first deployments


steren:
I see that [`builders/deploy.yaml` is deploying to Cloud Run with the `--no-traffic` flag](https://github.com/google/exposure-notifications-server/blob/56c1484ced445579b3fc603affb0496e2642579a/builders/deploy.yaml#L57).

The `--no-traffic` flag cannot be used when creating a new service, traffic will always reach the first revision. If used, the command errors out with the message:
```
Deployment failed
ERROR: (gcloud.run.deploy) --no-traffic not supported when creating a new service.
```

Is this script expected to be used to create the services? If so, I guess it would fail.

Stepping back, what is the rationale for using `--no-traffic`? 
Is the idea that traffic should be manually routed?

mikehelmick:
I *think* that is being used for c/i only at the moment

mikehelmick:
/assign @sethvargo 

The work in progress will make this clear.

Terraform for initial setup and this will be part of incremental, but not first, deploy

Issue #357, opened 2020-05-19
Move off of gcr.io/cloud-builders/gcloud


sethvargo:
It's deprecated:

```
***** DEPRECATION NOTICE *****

This image is deprecated and will no longer be updated.
This recent version of the image will continue to exist.

In place of this image, please use one of the following
images from https://hub.docker.com/r/google/cloud-sdk/:

google/cloud-sdk
google/cloud-sdk:slim
google/cloud-sdk:alpine

Please note that these images support pinned versions
as well.

***** DEPRECATION NOTICE *****
```

steren:
I asked for clarification internally.

gcr.io/cloud-builders/gcloud is used extensively all over the GCP docs. I doubt it's really deprecated (especially in favor of an image hosted on Dockerhub).

My guess is that the deprecation notice is deprecated :)

Please hold until we learn more from the Cloud Build team.

mikehelmick:
@ImJasonH Can you provide any insight?

ImJasonH:
The deprecation message should link to a bug or doc with more motivation, and should recommend gcr.io/google.com/cloudsdktool/cloud-sdk

The issue is https://github.com/GoogleCloudPlatform/cloud-builders/issues/638

cc @bendory

bendory:
Yes -- you'll get that notice if you look in the repo (or build from `:latest`; the hosted service is not yet showing that deprecation notice to users, so I haven't yet done a full cleanup sweep.

sethvargo:
The deprecation message says to use `google/cloud-sdk`, but @ImJasonH you're suggesting `gcr.io/google.com/cloudsdktool/cloud-sdk` instead?

ImJasonH:
As @bendory said, the image you're using isn't rolled out to GCB yet, so users shouldn't be seeing this message at all today. I believe the intention is that the error message will be changed before it is rolled out.

bendory:
Yes -- and indeed, this is already updated here in the repo.
Reference: https://github.com/GoogleCloudPlatform/cloud-builders/pull/674

PR #385, opened 2020-05-20
Add scripts for build/deploy

sethvargo:
- Fixes GH-376
- Fixes GH-360
- Fixes GH-357

### DIFF
--- a/builders/deploy.yaml
+++ b/builders/deploy.yaml
@@ -1,80 +1,112 @@
-options:
-  env:
-  - 'KO_DOCKER_REPO=us.gcr.io/${PROJECT_ID}'
-  - 'DOCKER_REPO_OVERRIDE=us.gcr.io/${PROJECT_ID}'
-steps:
-# Build ko, prerequisite.
-- name: 'gcr.io/cloud-builders/docker'
-  args: ['build', '--tag=gcr.io/$PROJECT_ID/ko', '-f', 'terraform/ko.Dockerfile', 'terraform/']
-- name: 'gcr.io/cloud-builders/docker'
-  args: ['push', 'gcr.io/$PROJECT_ID/ko']
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
-# Tests
-- name: 'mirror.gcr.io/library/golang'
-  env:
-  - GO111MODULE=on
-  args: ['go', 'test', './...']
-# Build and publish containers`
-- name: 'gcr.io/$PROJECT_ID/ko'
+#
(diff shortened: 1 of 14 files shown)


builders/deploy.yaml
```diff
@@ -1,80 +1,112 @@
...
+  - 'bash'
+  - '-c'
+  - |-
+    gcloud run deploy export \
+      --quiet \
+      --project "${PROJECT_ID}" \
+      --platform "managed" \
+      --region "${_REGION}" \
+      --image "us.gcr.io/$PROJECT_ID/github.com/google/exposure-notifications-server/cmd/export:latest" \
+      --no-traffic
```
mikehelmick, line 36: I think we should leverage traffic splitting here and roll out to 5% on the deploy

builders/promote.yaml
```diff
@@ -0,0 +1,106 @@
...
+  args:
+  - 'bash'
+  - '-c'
+  - |-
+    gcloud run services update-traffic export \
+      --quiet \
+      --project "${PROJECT_ID}" \
+      --platform "managed" \
+      --region "${_REGION}" \
+      --to-revisions "${_REVISION}=100"
```
mikehelmick, line 35: where are we getting the revision name from?

scripts/promote
```diff
@@ -0,0 +1,39 @@
...
+  echo "✋ Missing REGION!" >&2
+  exit 1
+fi
+
+if [ -z "${REVISION:-}" ]; then
+  echo "🚧 Using LATEST revision" >&2
+  REVISION="LATEST"
+fi
+
+gcloud builds submit "${ROOT}" \
```
mikehelmick, line 36: while we’re here - we shold also take the target percentgage, allow for gradual rollout

builders/deploy.yaml
sethvargo: The `gcloud run deploy` command only supports "all" or "no" traffic. There's no way to specify rollout.

builders/promote.yaml
sethvargo: `./scripts/promote`. Default is "LATEST", but can override.

scripts/promote
sethvargo: Done

builders/deploy.yaml
mikehelmick: that’s too bad... because it’s in the API :)

we’d have to make that 2 commands then

sethvargo: I'd rather not duplicate promotion effort in deploy.yaml, since we already have promote.yaml. How about I update `./scripts/build` to call promote with 5% after?

sethvargo: Actually, I think this should be two separate commands. That gives the ability to deploy a canary and change promotions independently.

sethvargo: You could always run `./scripts/deploy && ./scripts/promote`, but I don't think we should clobber deployment and promotion.

steren: +1, that's why there is no percentage in deploy, it's OK to type 2 commands :)

scripts/promote
```diff
@@ -0,0 +1,44 @@
...
+fi
+
+if [ -z "${PERCENTAGE:-}" ]; then
+  echo "🚧 Moving 100% of traffic" >&2
+  PERCENTAGE="100"
+fi
+
+gcloud builds submit "${ROOT}" \
+  --config "${ROOT}/builders/promote.yaml" \
+  --substitutions "_REGION=${REGION},_REVISION=${REVISION},_PERCENTAGE=${PERCENTAGE}" \
```
steren, line 43: How can `REVISION` be applied to all services? A revision name is always prefixed by the service name.

PR #391, opened 2020-05-20
Update tek-file documentation

ithinkihaveacat:
Previous docs seem to be inaccurate. This makes it a bit long, but not sure how else to do this. Point to URL explaining the format?

### DIFF
--- a/tools/export-generate/main.go
+++ b/tools/export-generate/main.go
@@ -43,7 +43,7 @@ var (
 	startTimestamp = flag.String("start-timestamp", "2020-05-01T15:00:00Z", "The test export start timestamp (RFC3339).")
 	endTimestamp   = flag.String("end-timestamp", "2020-05-02T15:00:00Z", "The test export end timestamp (RFC3339).")
 	numKeys        = flag.Int("num-keys", 450, "Number of total random temporary exposure keys to generate in the export")
-	tekFile        = flag.String("tek-file", "", "File containing one base64 encoded Temporary Exposure Key per line. If set, num-keys ignored.")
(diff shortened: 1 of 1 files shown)


tools/export-generate/main.go
```diff
@@ -43,7 +43,7 @@ var (
 	startTimestamp = flag.String("start-timestamp", "2020-05-01T15:00:00Z", "The test export start timestamp (RFC3339).")
 	endTimestamp   = flag.String("end-timestamp", "2020-05-02T15:00:00Z", "The test export end timestamp (RFC3339).")
 	numKeys        = flag.Int("num-keys", 450, "Number of total random temporary exposure keys to generate in the export")
-	tekFile        = flag.String("tek-file", "", "File containing one base64 encoded Temporary Exposure Key per line. If set, num-keys ignored.")
+	tekFile        = flag.String("tek-file", "", "JSON file of TEKs in the format {\"temporaryExposureKeys\":[{\"key\":\"…\",\"rollingStartNumber\":1,\"rollingPeriod\":1,\"transmissionRisk\":1}]}")
```
gurayAlsac, line 46: Good catch! Artifact from an earlier phase of development. Agree its a bit wordy. Maybe something like this?

```suggestion
	tekFile        = flag.String("tek-file", "", "JSON file of TEKs in the same format as calling publish endpoint")
```
It's essentially the format used in https://github.com/google/exposure-notifications-server/blob/a8f6578fb5d364610afeee17fe3373d21503a028/internal/publish/publish.go#L43  so we could refer to it and not document it inline.

ithinkihaveacat: Oh yeah that works, I'd expect/hope that most people come across the `keys.json` one way or another too, that's pretty self-explanatory.

PR #393, opened 2020-05-21
Allow per-app config of DeviceCheck and SafetyNet

sethvargo:
Re, un-re-do


sethvargo:
/assign @mikehelmick 

This is actually a relatively small change given the recent refactor.

PR #394, opened 2020-05-21
Separate Export Model and DB

crwilcox:
Towards #35


crwilcox:
@sethvargo happy to wait until am. @jba looked at the initial one and we chatted yesterday.

internal/cleanup/cleanup.go
```diff
@@ -127,7 +129,7 @@ func (h *exportCleanupHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 	timeoutCtx, cancel := context.WithTimeout(ctx, h.config.Timeout)
 	defer cancel()
 
-	count, err := h.database.DeleteFilesBefore(timeoutCtx, cutoff, h.blobstore)
+	count, err := exportdb.NewExportDB(h.database).DeleteFilesBefore(timeoutCtx, cutoff, h.blobstore)
```
mikehelmick, line 132: I don't know if we need these to be typed or just package level methods.
Having worked briefly w/ the first refactor I think I would prefer

```exportdb.DelteFiledBefore(h.databse, ...)```

Comment (crwilcox) [reply internal/cleanup/cleanup.go:132]: @mikehelmick I am pretty indifferent and could make this change. I had thought perhaps once fully moved and separated the handler could be updates do `h.database` was pointing to `exportdb` (so you wouldn't be making the objects) 

Something like:
```
return &exposureCleanupHandler{
		config:   config,
		env:      env,
		exposureDatabase: exposuredb.NewExposureDB(env.Database()),
		exportDatabase: exportdb.NewExportDB(env.Database()),
	}, nil
```

File: internal/cleanup/cleanup.go [hunk 2]
```diff
@@ -22,6 +22,8 @@ import (
 	"time"
 
 	"github.com/google/exposure-notifications-server/internal/database"
+	exportdb "github.com/google/exposure-notifications-server/internal/export/database"
```
tritone, line 25: I see that here you alias exportdb whereas in other files you alias coredb. Is there any logic to this? Seems like it would be better to keep consistent with what is aliased or not.

internal/export/database/export.go
```diff
@@ -32,8 +34,18 @@ const (
 	oneDay       = 24 * time.Hour
 )
 
+type ExportDB struct {
+	db *database.DB
+}
+
+func NewExportDB(db *database.DB) *ExportDB {
```
tritone, line 41: Will this eventually take additional params, or use a default for db if the db is nil? If so what do you think about using an option pattern:

```
func ExportDB(opts ...ExportDBOption) *ExportDB {
...
}

func WithDB(db *DB) ExportDBOption {
}
```

internal/cleanup/cleanup.go
jba, line 132: Since the package is already called `exportdb`, you could call the function `New` to avoid stutter.

internal/export/exportfile.go
```diff
@@ -39,12 +41,12 @@ const (
...
 
 type ExportSigners struct {
-	SignatureInfo *database.SignatureInfo
+	SignatureInfo *model.SignatureInfo
 	Signer        crypto.Signer
 }
 
 // MarshalExportFile converts the inputs into an encoded byte array.
-func MarshalExportFile(eb *database.ExportBatch, exposures []*database.Exposure, batchNum, batchSize int, signers []ExportSigners) ([]byte, error) {
+func MarshalExportFile(eb *model.ExportBatch, exposures []*database.Exposure, batchNum, batchSize int, signers []ExportSigners) ([]byte, error) {
```
gurayAlsac, line 49: Is this PR the precursor for a forthcoming deeper change? I had imagined from the description that we'd be breaking the link between the internal operations of exports and how data is represented in db (i.e., model objects vs DAO objects). I definitely found it weird during https://github.com/google/exposure-notifications-server/pull/353 that in order to generate an export file with no data from db I had to go through db concepts.

But unless I misread, model is just database renamed. And we've simply relocated stuff from internal/database to internal/export/database. I'd have imagined this would head towards duplicating stuff in internal/database to an internal/model and making DAO layer more explicit.

crwilcox: this change is moving models to a separate place from db. Db operations for export are separate from db operations for `_other_thing_`.

internal/export/database/export.go
crwilcox: Unsure at this point if we will need more options. Thanks for the pattern though!

internal/cleanup/cleanup.go
jba: Is `h.database` a reference to the core DB object? So anyone with that could execute arbitrary statements against the DB? I thought part of the idea was to prevent the api packages (like `internal/cleanup`) from holding on to one of those.

crwilcox: correct `h.database` is a `coredb` object reference. `env.Database()` provides one. By moving it into the `exposureCleanupHandler`, it would be hard to accidentally be using it though. Also, as part of this refactor `h.database` has none of the functions of tables, but has functions like `Pool`, `InTx`

jba: Got it. So then, I like your design two comments up where the handler struct holds only the wrapped DB handles. Then it is clear from inspection that the handler methods can't do anything with core handle; they have to use the api-specific ones.

crwilcox: Thanks for the feedback. I have made the changes and pushed.

PR #397, opened 2020-05-21
Filesystem support

mgulimonov:
Making Blobstore configurable via environment variables.
The feature allows configuring different blob store systems for different installations.

### DIFF
--- a/internal/cleanup/config.go
+++ b/internal/cleanup/config.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/google/exposure-notifications-server/internal/database"
 	"github.com/google/exposure-notifications-server/internal/setup"
+	"github.com/google/exposure-notifications-server/internal/storage"
 )
 
 // Compile-time check to assert this config matches requirements.
@@ -28,10 +29,11 @@ var _ setup.DBConfigProvider = (*Config)(nil)
 // Config represents the configuration and associated environment variables for
 // the cleanup components.
 type Config struct {
-	Port     string        `envconfig:"PORT" default:"8080"`
-	Timeout  time.Duration `envconfig:"CLEANUP_TIMEOUT" default:"10m"`
-	TTL      time.Duration `envconfig:"CLEANUP_TTL" default:"336h"`
-	Database *database.Config
+	Port          string        `envconfig:"PORT" default:"8080"`
+	Timeout       time.Duration `envconfig:"CLEANUP_TIMEOUT" default:"10m"`
+	TTL           time.Duration `envconfig:"CLEANUP_TTL" default:"336h"`
+	Database      *database.Config
+	BlobstoreType string `envconfig:"BLOBSTORE_TYPE" default:"GCS"`
 }
 
 // DB return the databsae configuration.
@@ -40,6 +42,8 @@ func (c *Config) DB() *database.Config {
 }
 
 // BlobStorage returns the BlobStorage configuration.
-func (c *Config) BlobStorage() bool {
-	return true
+func (c *Config) BlobStorage() storage.BlobstoreConfig {
+	return storage.BlobstoreConfig{
+		Factory: storage.NewBlobstoreFactory(storage.BlobstoreType(c.BlobstoreType)),
+	}
 }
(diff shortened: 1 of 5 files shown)


internal/cleanup/config.go
```diff
@@ -28,10 +29,11 @@ var _ setup.DBConfigProvider = (*Config)(nil)
...
 type Config struct {
-	Port     string        `envconfig:"PORT" default:"8080"`
-	Timeout  time.Duration `envconfig:"CLEANUP_TIMEOUT" default:"10m"`
-	TTL      time.Duration `envconfig:"CLEANUP_TTL" default:"336h"`
-	Database *database.Config
+	Port          string        `envconfig:"PORT" default:"8080"`
+	Timeout       time.Duration `envconfig:"CLEANUP_TIMEOUT" default:"10m"`
+	TTL           time.Duration `envconfig:"CLEANUP_TTL" default:"336h"`
+	Database      *database.Config
+	BlobstoreType string `envconfig:"BLOBSTORE_TYPE" default:"GCS"`
```
mikehelmick, line 36: Throughout - Change GCS to GoogleCloudStorage

assume the TLA isn't well known outside of Google.

internal/cleanup/config.go
```diff
@@ -40,6 +42,8 @@ func (c *Config) DB() *database.Config {
 }
 
 // BlobStorage returns the BlobStorage configuration.
-func (c *Config) BlobStorage() bool {
-	return true
+func (c *Config) BlobStorage() storage.BlobstoreConfig {
+	return storage.BlobstoreConfig{
```
mikehelmick, line 46: this same function over and over again seems like an anti-pattern.

Why not have BlobStorage return a string and centralize the creation in Setup.

The whole "factory" concept can possible be removed if this is only done in one place.

internal/storage/storage.go
```diff
@@ -25,3 +44,43 @@ type Blobstore interface {
...
+	return nil
+}
+
+// No op.
+func (s *NoopBlobstore) DeleteObject(ctx context.Context, folder, filename string) error {
+	return nil
+}
+
+// Creates a new BlobstoreFactory based on the provided BlobstoreType.
+func NewBlobstoreFactory(blobstoreType BlobstoreType) BlobstoreFactory {
```
mikehelmick, line 69: yeah - if you change the config provider func to return a string and not a factory - this can just be a method that is something like

func CreateBlobstore - accepting the blobstore type

internal/setup/setup.go
```diff
@@ -86,13 +86,15 @@ func Setup(ctx context.Context, config DBConfigProvider) (*serverenv.ServerEnv,
 		}
 		opts = append(opts, serverenv.WithKeyManager(km))
 	}
-	// TODO(mikehelmick): Make this extensible to other providers.
+
 	if _, ok := config.(BlobStorageConfigProvider); ok {
-		storage, err := storage.NewGoogleCloudStorage(ctx)
+		provider := config.(BlobStorageConfigProvider)
```
jba, line 91: In Go we're usually suspicious of providers and factories. 
If you're trying to create a particular kind of blob store based on a string (coming from the environment, or wherever), then all you need is
```
switch blobStoreKind {
case "gcs":
   return gcsblobstore.New(...)
case "file":
   return fileblobstore.New(...)
}
```

internal/cleanup/config.go
mgulimonov: I changed this as you suggested to keep at simple. The initial idea was to keep it more extensible, because different apps that use the same Setup might want to use different strategies for Blobstore creation.

internal/setup/setup.go
mgulimonov: Fixed.
Even if I inline the factory code it still remains factory code. =)

jba: I think of everything in an API as having a weight. Struct fields have less weight than global functions, which have less weight than concrete structs, which have less weight than interfaces. I strive to minimize weight. A single top-level "factory" function is lighter than a Factory interface implemented by multiple structs each with their own methods. (Not that you had something that complicated—I honestly didn't look that carefully—but that is what tends to happen.)

Anyway, thanks for simplifying.

PR #398, opened 2020-05-21
internal/publish: don't pass a pointer to a pointer to json Unmarshal

davecheney:
Pass the address of data to jsonutil.Unmarshal rather than the address
of a pointer to an uninitalised data.

Signed-off-by: Dave Cheney <davecheney@github.com>


mikehelmick:
/lgtm
/approve

Thanks for catching that.

PR #399, opened 2020-05-21
fixed the way to hold http.Client

Code-Hex:
The Client's Transport typically has internal state (cached TCP connections), so Clients should be reused instead of created as needed. Clients are safe for concurrent use by multiple goroutines.

https://golang.org/pkg/net/http/#Client


Code-Hex:
@googlebot I signed it!

PR #401, opened 2020-05-21
Fix/terraform regions

jexh:
The infrastructure deployment must target a region that supports both CloudRun and AppEngine otherwise terraform fails with errors. I've added a locals entry mapping the valid regions for CloudRun to the associated region for AppEngine. This has the side effect of preventing deployments to invalid regions from progressing past the "terraform plan" stage.


jexh:
@googlebot I signed it!

krazykid:
@googlebot I signed it!

mikehelmick:
/ok-to-test
/hold

I think we should intead separate out the app engine (cloud scheduler region) and compute region. 

Cloud Run  + Cloud SQL (all the data at rest and data in flight) should be more specific, and the region that the scheduler is run from, can be close by, but can be diffrent.

sethvargo:
It _is_ currently separated out. Customers can choose `appengine_location` and `region` separately. I'm -1 on this change.

jexh:
> It _is_ currently separated out. Customers can choose `appengine_location` and `region` separately. I'm -1 on this change.
> 
> I'd be okay renaming "region" to something more specific, if you think that would clarify things.

But the deployment doesn't work unless the CloudRun components run in the same region as appengine - terraform bails out with an error like the following:

Error: Error creating Job: googleapi: Error 400: Location must equal us-central1 because the App Engine app that is associated with this project is located in us-central1
                                                                                                                                                      
  on service_cleanup_export.tf line 107, in resource "google_cloud_scheduler_job" "cleanup-export-worker":                                       
 107: resource "google_cloud_scheduler_job" "cleanup-export-worker" {

sethvargo:
Okay, then the bug is that cloud scheduler jobs need to use appengine_location, not region 😄

jexh:
> Okay, then the bug is that cloud scheduler jobs need to use appengine_location, not region

I'm not trying to catch you out, but, unless there is auto-renaming that I don't know about, there is also a problem with region ID names. These are sometimes different between AppEngine and CloudRun. In the map I have these differences: europe-west1 = "europe-west" and us-central1 = "us-central". (I was bringing up a deployment in europe which led me to this).

jexh:
How about I change the variables to have "data_region" and "exec_region". exec_region is used for appengine and the cloudrun services (using the map I added already to do the transformation). data_region is used for everything else?

sethvargo:
I think the only change we need to make is on all the `google_cloud_scheduler_job` resources and add:

```tf
region = var.appengine_location
```

jexh:
So, I tried this out in a new project, but it still doesn't work without the map. Setting the appengine_location to "europe-west" results in an error from the scheduler_job like this:

Error: Error creating Job: googleapi: Error 403: The principal (user or service account) lacks IAM permission "cloudscheduler.jobs.create" for the resource "projects/approoval-dev-trace-ecrf5/locations/europe-west" (or the resource may not exist).                                                                                                                                                                 
                                                                                                                                                                                                            
  on service_export.tf line 159, in resource "google_cloud_scheduler_job" "export-create-batches":                                                                                                          
 159: resource "google_cloud_scheduler_job" "export-create-batches" {   

In another new project, setting the appengine_location to "europe-west1" results in a problem creating the appengine itself:

Error: Error creating App Engine application: googleapi: Error 400: Unrecognized value "europe-west1"., badRequest

  on main.tf line 154, in resource "google_app_engine_application" "app":
 154: resource "google_app_engine_application" "app" {

So, how about I change the pull request to use the appengine_location variable but also add a scheduler_region map to generate the correct region ID for the jobs and have the jobs set their region based on the map?

jexh:
Of course, now I have to change tack a bit because I had lost sight of the real limiting factor which is the supported regions for CloudRun. If this is driven off var.region then deployment all goes well until the last 4 resources need to be deployed and then it fails. 

There are a small number of regions which support CloudRun:
asia-east1 (Taiwan)
asia-northeast1 (Tokyo)
europe-north1 (Finland)
europe-west1 (Belgium)
europe-west4 (Netherlands)
us-central1 (Iowa)
us-east1 (South Carolina)
us-east4 (Northern Virginia)
us-west1 (Oregon)

I am happy to adjust the code to put in more region variables, but then I think docs/deploying.md will need an update. Should it not be kept simple - one region variable and limited to those regions that work?

sethvargo:
Ah I see. Yes, not all services are available in all regions. We should solve this with docs (I don't think Terraform should do this validation)

crwilcox:
I agree we likely don't want to hardcode the list as it would only capture what exists today. locations and supported services evolves and capturing it as a point in time seems bad. We could include a list (as of now) in docs and link out to https://cloud.google.com/about/locations

jexh:
After thinking about this a bit more, I have to respectfully disagree. The problem is the balance between simplicity and flexibility. The code as it stands has 2 region related variables, 1 for the appengine and 1 for everything else. It's got quite a simple concept of region but it has a bug. I tried to do a deployment in Europe and I had a couple issues as a result:

* I came across the bug, which, in the existing code base, effectively ties the 2 variables together: the scheduler_job resources currently use the default region but they need to be deployed in the region that matches the appengine location_id(not necessarily the same strings).
* I also hit a problem by initially choosing a region that didn't support cloudrun. The terraform deployment fails at a very late stage in that case.

The PR solves both issues by simplifying the code. It takes away the choice of appengine location and makes sure the selected region supports cloudrun. (Requiring both appengine and cloudrun support really restricts the set of valid regions)

If you want to keep the flexibility without hard-coding any region identifiers then I believe the code will need more changes along with extended documentation. There are the existing variables: region and appengine_region. To fix the bug, there will need to be a scheduler_job_region, which must match the appengine_region but not necessarily use the same string. Lastly, it would make sense to also add a cloudrun_region as that is supported in a relatively small set of regions compared to the rest of the deployment. The documentation would need to explain these variables, how they are related, and where to find the documentation to find valid values.

After the arguments in previous comments and the summary above, I still think this PR is the basis of the correct answer, but I now realize that it should also contain documentation changes. Specifically, it needs to explain why there is a limited set of supported regions (keeping the code simple while offering limited flexibility), that the set of valid regions will likely expand in the future, that the code can be changed to distribute different resource types to different regions to expand the set of valid options, and where the information about region support can be found.

sethvargo:
Hi @jexh 

We're definitely interested in supporting configurable regions in the Terraform configuration, but I'm not sure this is the path we'd like to go. Personally I don't like limiting the list of available regions to a point-in-time in the Terraform configurations, nor do I think Terraform should be responsible for maintaining the map of GAE regions and Cloud Run regions. Those may expand over time, at which point we have to keep an arbitrary map in sync.

jexh:
OK, change of approach and I'll submit a new PR shortly

PR #404, opened 2020-05-21
Suggests being transparent about user anonymity in the requirements

qutorial:
Here I would like to politely and friendly suggest being transparent
on the security/privacy implications of people using the app.
From my comments in this commit it should (I hope) be evident,
why we potentially deanonymize (at least to Google and Apple) the
user while trying to perform device attestation.
qutorial@gmail.com

### DIFF
--- a/docs/server_functional_requirements.md
+++ b/docs/server_functional_requirements.md
@@ -5,6 +5,11 @@
 This documents the functional requirements for building a decentralized exposure
 notification system. For deployment strategies, see [Server Deployment Options](server_deployment_options.md).
 
+## Security requirements
+
+The current version of this software does not guarantee anonymity to Apple and/or Google
+of the person submitting exposure notification.
+
 ### System Components
 
 The Exposure Notification Server's architecture has been split into components.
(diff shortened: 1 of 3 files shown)


qutorial:
/assign @mikehelmick
@sethvargo , dear Mr. Vargo, please, do not consider this pull request impolite.
I am out of context on what you are doing, and maybe it is evident to you 100%.
Here I just wanted to be honest with the potential users and notice this.

qutorial:
Have submitted my CLA

mikehelmick:
/hold

Thank you for the PR.

Use of device attestations is optional in this system, but recommended.

App develoeprs are encourgaged to send “invalid” infection request periodically. In doing so, they would access DeviceCheck or safety net when the user of the app is not reporting an infection. The operators of the attestation system would not be able to distingish beween a user who has the app installed and a user that has reported they have been exposed.

The face that the app is installed is information that Apple or Google already has as they operate the appstore/playstore.

Let me know if this addresses your concern.

sethvargo:
Hi @qutorial 

Thank you for submitting a pull request. As of https://github.com/google/exposure-notifications-server/pull/393, individual apps can choose to enable/disable both Android SafetyNet and Apple DeviceCheck on a per-app basis. The server no longer requires these fields, and we've updated our documentation as such.

mikehelmick:
/close

Closing as there has been no response from author and we believe this has been addressed by making this functionality optional.

Issue #386, opened 2020-05-20
When should Terraform stop?


sethvargo:
For discussion and thoughts @danawillow @ndmckinley @mikehelmick 

Right now, Terraform configures an empty GCP project with all the things required to run an EN server. It _also_ does the initial provisioning and deployment of the Cloud Run services.

However, we need to build, deploy, promote, scale-up/scale-down those Cloud Run services outside of Terraform. In GH-385, I found that Terraform gets reaaaalll unhappy when someone updates a Cloud Run revision outside of Terraform. It actually goes into an unconvergable state.

- Should Terraform build and deploy the Cloud Run services, or should it stop at setting up all the things and maybe spit out a gcloud command?
- If Terraform should manage all those services, how can we easily force it to re-build and deploy those containers?

mikehelmick:
I would be happy with always deploying by modifying terraform configs, but it has to be able to build/deploy after first deploy.

sethvargo:
I think the current challenge is that Terraform has no visibility into building images. We shell out to gcloud and submit a build to Cloud Build. In an _ideal world_:

- Terraform submits build to Cloud Build
- Cloud Build returns the image SHA
- The Cloud Run service depends on image SHA (thus creating a dependency)

But support for submitting/receiving build information from Cloud Build doesn't exist in Terraform (mostly because `gcloud builds submit` is actually 40 API calls to 5 different services). In lieu of that, I'm kinda leaning towards:

- Use Terraform for all the things + initial Cloud Run deployment
- Add `ignore_changes` to everything on Cloud Run services

danawillow:
I think if you need to update the cloud run services outside of Terraform, it's going to be error-prone to try to use Terraform for them at all. We already do the initial deploy inside of cloud build; it wouldn't be too absurd to also create the services at that time with gcloud.

Would support for a Build resource/data source (https://cloud.google.com/cloud-build/docs/api/reference/rest/v1/projects.builds) help at all?

sethvargo:
The problem is that we need information in the build output, like information from gcr

krazykid:
I'm pretty new to TF and GCP, but have had pretty much a crash course on it in the last month and half. I was just wondering why not:
* Have TF create a Cloud Trigger which does:
  * Build
  * Deploy containers to Cloud Run (not shift traffic) via `gcloud run deploy ...` in the `cloudbuild.yaml` file
* Pull trigger (either through TF or hand)
[...]

PR #406, opened 2020-05-21
Get Terraform out of the CI/CD business

sethvargo:
I did a spike on this today, and I think it fixes GH-386. 

Terraform does the initial _everything_, but then ignores all future changes on Cloud Run services. This lets us run `terraform apply` and get a fully-functional setup, but then iterate and deploy on services individually.

The containers built by Terraform are tagged and deployed as `:initial`. Future builds use a timestamp-based tag.

I also addressed some of Steren's feedback in previous issues about being able to deploy individual services.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,5 +15,6 @@
 *.tfstate*
 *.terraform/
 terraform.tfvars
+terraform/state.tf
 
 /local/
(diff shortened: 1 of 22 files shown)


mikehelmick:
This 
/lgtm
But I'll let others review before merging.

PR #407, opened 2020-05-21
Separate db and model for Exposure/Publish

crwilcox:


internal/export/batcher.go
```diff
@@ -25,6 +25,8 @@ import (
 	coredb "github.com/google/exposure-notifications-server/internal/database"
 	"github.com/google/exposure-notifications-server/internal/export/database"
 	"github.com/google/exposure-notifications-server/internal/export/model"
+	exposuremodel "github.com/google/exposure-notifications-server/internal/publish/model"
```
crwilcox, line 28: I think we should go to either exposure everywhere or publish, for now I left it alone and try to stick to the way we were using it though.

sethvargo: I like "publish" since all the other names are verbs, but either way I agree. This confused me a few times.

mikehelmick: changing publish to exposure is fine -can be a follow up package rename only.

PR #415, opened 2020-05-22
Adding method for extracting keys from the export file.

mgulimonov:
Adding method for extracting keys from the export file.


internal/export/exportfile_test.go
```diff
@@ -0,0 +1,66 @@
...
+		SignatureInfos: infos,
+		Keys:           keys,
+	}
+
+	got, err := UnmarshalExportFile(bytes)
+	if err != nil {
+		t.Errorf("Unmarshal failed: %w", err)
+	}
+
+	assert.Equal(t, keyExport.StartTimestamp, got.StartTimestamp)
```
sethvargo, line 59: We generally don't recommend using testing libraries in Go

mgulimonov: What's the reason?

sethvargo: It's generally frowned upon in the community. Also, instead of comparing individual fields, you might use `reflect.DeepEqual` or https://github.com/google/go-cmp (which we use in other tests).

internal/export/exportfile_test.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"io/ioutil"
+	"testing"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/google/exposure-notifications-server/internal/pb/export"
+	"github.com/stretchr/testify/assert"
+)
+
+const sampleFilename = "../../examples/export/testExport-2-records-1-of-1.zip"
```
gurayAlsac, line 12: optional: You might consider restructuring this to be a test of marshalling followed by unmarshalling and decouple from the example file. Or if we find it specifically beneficial to use a file, you could put one in ./testdata

internal/export/exportfile.go
```diff
@@ -84,8 +86,51 @@ func MarshalExportFile(eb *model.ExportBatch, exposures []*publishmodel.Exposure
 	return buf.Bytes(), nil
 }
 
+// Unmarshals export file and extract exposure keys information.
```
odeke-em, line 89:
```go
// Unmarshal extracts the protobuf encoded exposure key present in the zip archived payload.
```

internal/export/exportfile.go
```diff
@@ -84,8 +86,51 @@ func MarshalExportFile(eb *model.ExportBatch, exposures []*publishmodel.Exposure
...
+func UnmarshalExportFile(payload []byte) (*export.TemporaryExposureKeyExport, error) {
```
odeke-em, line 90: Perhaps let’s update the name of “payload” to “zippedProtoPayload”

internal/export/exportfile.go
```diff
@@ -84,8 +86,51 @@ func MarshalExportFile(eb *model.ExportBatch, exposures []*publishmodel.Exposure
...
+		return nil, err
+	}
+
+	content, err := ioutil.ReadAll(f)
+	if err != nil {
+		return nil, err
+	}
+
+	prefix := content[:fixedHeaderWidth]
+	sprefix := string(prefix)
```
odeke-em, line 118: If we make fixedHeader as []byte, we can do if !bytes.Equal and will avoid the many bytes->string conversions.

internal/export/exportfile_test.go
```diff
@@ -0,0 +1,66 @@
+package export
```
odeke-em, line 1: Missing license header.

internal/export/exportfile_test.go
odeke-em: Same thoughts here about putting it in ./testdata, thanks @gurayAlsac.

internal/export/exportfile.go
mgulimonov: We can't declare a constant with []byte. 
Should we use a variable for this?

internal/export/exportfile.go
```diff
@@ -33,6 +34,7 @@ import (
 
 const (
 	fixedHeaderWidth     = 16
+	fixedHeader          = "EK Export v1    "
```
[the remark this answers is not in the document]
gurayAlsac: The 16 is baked into the format, so if we infer from an improper string and go longer/shorter, that'd create invalid files. If we really wanted we could change this to unpaddedFixedHeader, pad in code, and throw if it exceeds 16.

internal/export/exportfile_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+	blob, err := MarshalExportFile(batch, exposures, 1, 1, []ExportSigners{
+		{SignatureInfo: signatureInfo,
+			Signer: signer}})
+	if err != nil {
+		t.Fatalf("Can't marshal export file, %v", err)
+	}
+
+	filePath := tmp + "/test-export.zip"
+	err = ioutil.WriteFile(filePath, blob, 0644)
```
gurayAlsac, line 94: I suppose the test doesn't need actually to exercise writing to storage/reading from storage to verify marshalling/unmarshalling. That might make it more flexible where it runs, but I'm not fully up to speed on our continuous integration testing story so this might be moot.

internal/export/exportfile.go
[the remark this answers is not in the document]
odeke-em: Gotcha, thanks!

internal/export/exportfile.go
odeke-em: I'd go for it, reduces conversion costs but moreover is usable only in one place in code, aside from tests.

odeke-em: Actually usable in 2 places and prevents unnecessary conversions in 2 places :)

internal/export/exportfile_test.go
```diff
@@ -0,0 +1,158 @@
...
+		BatchNum:       proto.Int32(1),
+		BatchSize:      proto.Int32(1),
+		SignatureInfos: infos,
+		Keys:           keys,
+	}
+
+	ignoredTemporaryExposureKeyExportFields := cmpopts.IgnoreUnexported(export.TemporaryExposureKeyExport{})
+	ignoredSignatureInfoFields := cmpopts.IgnoreUnexported(export.SignatureInfo{})
+	ignoredTemporaryExposureKeyFields := cmpopts.IgnoreUnexported(export.TemporaryExposureKey{})
+	if diff := cmp.Diff(keyExport, got, ignoredTemporaryExposureKeyExportFields, ignoredSignatureInfoFields, ignoredTemporaryExposureKeyFields); diff != "" {
```
odeke-em, line 145: Minor nit: given how long these fieldnames are perhaps we can put this on its own line so
```go
	diff := cmp.Diff(keyExport, got, ignoredTemporaryExposureKeyExportFields, ignoredSignatureInfoFields, ignoredTemporaryExposureKeyFields)
        if diff != "" {
```

odeke-em:
LGTM, after addressing my suggestions, thank you @mgulimonov!

PR #418, opened 2020-05-22
fixed map allocation in federationout

Code-Hex:
We know how many you want to allocate, so I think it’s better to do it in one go as much as possible.


jba:
Just for the record, I would consider these premature optimizations. Not wrong, just not recommended style.

PR #426, opened 2020-05-23
Changed timestamp handling in the export-generate tool.

mh-:
@gurayAlsac The start and end timestamps generated by tools/export-generate (and published in examples/export/testExport-2-records-1-of-1.zip) are a bit confusing: The export file apparently contains keys that have been uploaded on May 1, before they became valid. Whereas in production use, the timestamps from most devices (non-whitelisted) will be at least 1 day old. This can be fixed by manually setting start and end timestamp values via command line flags, but I think it's more convenient if the default (no flag specified) creates a valid timestamp / interval combination.
Also, the start intervals of randomly generated keys are replaced by new random values <=144, which means 1970.

This PR contains:
The export's start and end timestamps now reflect the past 24h period.
If TEKs are supplied in a JSON file, their start intervals must be at least 24h old.
If random TEKs are generated by this tool, their parameters are now taken from the random keys that util.GenerateExposureKeys() generates, but their start interval is reduced by 144 (24h).


mikehelmick:
@mh- 

If the goal is more realistic data, then I think the correct thing to do is allow for [GenerateExposureKeys](https://github.com/google/exposure-notifications-server/blob/master/internal/util/generate.go#L73) to be allowed to generate keys that start up to 14 (15th midnight UTC boundary) in the past.

Any given export will not just contain new keys from that day, but up to the past 14 days.

In reality, most keys will have a start interval that aligns w/ the midnight UTC boundary and a period of 144

mh-:
> If the goal is more realistic data, then I think the correct thing to do is allow for [GenerateExposureKeys](https://github.com/google/exposure-notifications-server/blob/master/internal/util/generate.go#L73) to be allowed to generate keys that start up to 14 [days] (15th midnight UTC boundary) in the past.

@mikehelmick My first goal was to avoid 'impossible' data, and I tried to do that with minimal changes, so that code review doesn't take too much effort for something that isn't really a major issue.

As it is now, [GenerateExposureKeys](https://github.com/google/exposure-notifications-server/blob/master/internal/util/generate.go#L73) generates keys 'backwards from now', i.e. the first generated key is valid from 'up to -144 intervals' until 'now', and each subsequently generated key is valid for 'up to -144 intervals' before the previously generated key. The durations are random, and there is no bound on the number of days this can go back. 
This has advantages in that it produces corner cases which don't happen often in reality, but it also isn't very realistic. 

I'd be happy to implement something that gives more realistic data, i.e. keys up to the past 14 days, and with most keys having a start interval that aligns w/ the midnight UTC boundary and a period of 144. Let me know.

tools/export-generate/main.go
```diff
@@ -124,12 +128,12 @@ func main() {
 				log.Fatalf("unable to decode key: %v", k.Key)
 			}
 			exposureKeys[i].ExposureKey = decoded
-			n, err := util.RandomIntervalCount()
-			if err != nil {
-				log.Fatalf("problem with interval count: %v", err)
+			exposureKeys[i].IntervalNumber = k.IntervalNumber - 144 // typically the key will be at least 1 day old
```
gurayAlsac, line 131: nit: instead of 144 we could make use of the MaxIntervalCount constant in publish/model/exposure_model.go

mh-: @gurayAlsac Ok, I [changed](https://github.com/google/exposure-notifications-server/pull/426/commits/91cd84361bd9b953bb28408bddbfcc33104d9223) that.

PR #428, opened 2020-05-24
Fix database tests

sethvargo:
Database tests haven't been running since https://github.com/google/exposure-notifications-server/commit/f113c0d2#diff-f7ac49848dfd0ce5e8f3251a560f5a0f, because of the package being renamed to no longer include a _test.go suffix. Not shown in that diff is the inclusion of a TestMain, which only runs if the file is actually a test (ends in the _test.go suffix). This and other dependent database tests have been silently skipping.

This refactors the tests to use a new Docker container per instance (where test implementers decide how they want to manage the instance), and updates all the database tests to run in parallel. This results in a net increase of 8s to the overall test suite runtime on my local machine, but seems worth it given this is the 4th (5th?) time a seemingly unrelated change as silently broken the database tests.

Folks can opt of out running the tests (either using -short or setting an environment variable), but the default behavior is now to run the tests. The only requirement is Docker (which was already a requirement anyway).


go.mod
```diff
@@ -44,6 +44,7 @@ require (
 	github.com/mitchellh/mapstructure v1.3.1 // indirect
 	github.com/oklog/run v1.1.0 // indirect
 	github.com/oracle/oci-go-sdk v19.3.0+incompatible // indirect
+	github.com/ory/dockertest v3.3.5+incompatible
```
sethvargo, line 47: There were not changes to go.sum, meaning we were already indirectly depending on this somewhere anyway.

internal/database/database_util.go
```diff
@@ -0,0 +1,162 @@
+// Copyright 2020 Google LLC
```
sethvargo, line 1: Grr. Something is wrong with my `git mv` command.

internal/database/database_util.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	// Create the pool (docker instance).
+	pool, err := dockertest.NewPool("")
+	if err != nil {
+		tb.Fatalf("failed to create Docker pool: %s", err)
+	}
+
+	// Start the container.
+	dbname, username, password := "en-server", "my-username", "abcd1234"
```
sethvargo, line 63: I'm open to generating random values here. Pretty easy to drop in `sethvargo/go-password`, but it seemed like overkill.

scripts/presubmit.sh
```diff
@@ -101,21 +101,11 @@ echo "🚧 Compile"
...
-
-if [ "${DB_USER:-}" == "" ]; then
-   echo "🚨 DB_USER is not configured. Test DB tests will not run."
-fi
-
-
 echo "🧪 Test"
 go test ./... \
   -coverprofile=coverage.out \
+  -count=1 \
```
sethvargo, line 107: Just in case we add a caching layer to CI as discussed in the vendoring issue.

sethvargo:
Wow, barely any additional timing increase on CI either 😄

internal/database/database_util.go
mikehelmick: I think it's Overkill since this is just for containerized tests

johanbrandhorst:
This is great, I was going to try contributing something like this myself, `dockertest` is awesome.

scripts/presubmit.sh
johanbrandhorst: I'm confused, why do we want to overrule the go test cache? This will ensure we run tests that may not need to be run.

PR #449, opened 2020-05-27
Add ExportConfig to admin console

### LINKED ISSUE #370: Create more reliable seed script
The current seed (`testdata/config.sql`) has hardcoded project IDs for the test server. We should create a more robust script for seeding the database, likely inside `scripts/dev`.

### LINKED ISSUE #382: Question about export-config
I am wondering who is supposed to add export-config.  It seems like the export worker and create-batch are relying on the export-config to generate key files. I saw there is a tool to create the export-config and it will be nice to add some documentation to explain this export flow.  Thanks in advance.

mikehelmick:
Last major UI change

/fixes #370
/fixes #382

### DIFF
--- a/internal/admin/exports/form.go
+++ b/internal/admin/exports/form.go
@@ -1,0 +1,55 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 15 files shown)


internal/admin/siginfo/view.go
```diff
@@ -39,12 +39,14 @@ func (v *viewController) Execute(c *gin.Context) {
 	m := admin.TemplateMap{}
 
 	sigInfo := &model.SignatureInfo{}
-	if sigID := c.Param("id"); sigID == "0" {
+	if IDParam := c.Param("id"); IDParam == "0" {
```
sethvargo, line 42: I think this would be `idParam`?

PR #455, opened 2020-05-27
Adding tool to read content of export files.

mgulimonov:


tools/export-analyzer/main.go
```diff
@@ -0,0 +1,50 @@
...
+		log.Fatalf("can't read export file: %v", err)
+	}
+
+	keyExport, err := export.UnmarshalExportFile(blob)
+	if err != nil {
+		log.Fatalf("error unmarshalling export file: %v", err)
+	}
+
+	prettyJSON, err := json.MarshalIndent(keyExport, "", "  ")
+	log.Printf("%v", string(prettyJSON))
```
gurayAlsac, line 49: I've no objection to having a dedicated tool to display contents of an export file, but note that it can also be done with protoc fairly easily:
`unzip -p export.zip export.bin | tail +17c | protoc --decode TemporaryExposureKeyExport export.proto`

This is demonstrated in https://github.com/google/exposure-notifications-server/tree/master/examples/export#inspecting-an-export

PR #456, opened 2020-05-27
Add optional deployment of generate service

### LINKED ISSUE #454: Add data generation to terraform config?
### Question

should we add the new data generation server to the default terraform config?

could be done without adding the cloud scheduler component.

pro: would be useful in getting people up and running with test data

con: should be disabled before going to production.
-- Low risk of harm if left on, as we also generate padded fake data on downloads.

sethvargo:
Fixes #454

### DIFF
--- a/terraform/service_generate.tf
+++ b/terraform/service_generate.tf
@@ -1,0 +1,140 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#
+# Create and deploy the service
+#
+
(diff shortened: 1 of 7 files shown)


terraform/service_generate.tf
```diff
@@ -0,0 +1,148 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#
+# Create and deploy the service
+#
+
+resource "google_service_account" "generate" {
```
sethvargo, line 19: I kinda wonder if we should always build and deploy this service, but instead have the default Cloud Scheduler settings be "never" (default of February 31st or something). That would technically add $0.10, but it would reduce the variability in these configs (and the conditional on count).

terraform/vars.tf
```diff
@@ -50,6 +50,13 @@ variable "cloudsql_disk_size_gb" {
   description = "Size of the Cloud SQL disk, in GB."
 }
 
+variable "create_generate_service" {
```
sethvargo, line 53: If we went the "always create and deploy this service, but never actually run it unless the user configures it" route, this would be something like `generate_schedule_cron` instead.

terraform/service_generate.tf
mikehelmick: yeah - I think always deploy it but not schedule it is good.

sethvargo: Done. Please re-review @mikehelmick

PR #457, opened 2020-05-27
fix issues with index file

### LINKED ISSUE #453: Remove deleted filename entries from index.txt on cleanup-export operation
Cleanup-export operation updates exportfile table status as well as deletes the files from GCS but it doesn't update the index.txt by removing deleted file entries.

gurayAlsac:
This fixes two problems in the index files:
1) no longer limit available export files to those in same
   export config. During a config rollover this would have
   caused all old files to disappear from index
2) fixes #453 by adding time constraint to lookup query.
   since emitIndexForEmptyBatch = true at least once we
  don't need to specially invoke index cleanup after wipeout
  and can rely on next batch worker

### DIFF
--- a/internal/export/config.go
+++ b/internal/export/config.go
@@ -40,6 +40,7 @@ type Config struct {
 	TruncateWindow time.Duration `envconfig:"TRUNCATE_WINDOW" default:"1h"`
 	MinWindowAge   time.Duration `envconfig:"MIN_WINDOW_AGE" default:"2h"`
 	BlobstoreType  string        `envconfig:"BLOBSTORE_TYPE" default:"CLOUD_STORAGE"`
+	TTL            time.Duration `envconfig:"CLEANUP_TTL" default:"336h"`
 }
 
 // DB returns the database config.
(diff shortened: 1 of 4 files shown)


sethvargo:
Will this cleanup existing generated index.txts?

gurayAlsac:
> Will this cleanup existing generated index.txts?

Yes, it should. The index file is regenerated by the workers for a batch (even empty ones).

varshithr:
This doesn't remove the deleted records which are already present in index.txt, right?
Since it is an append operation, the deleted entries still exist in index.txt.

Is this expected behavior?

gurayAlsac:
> This doesn't remove the deleted records which are already present in index.txt, right?
> Since it is an append operation, the deleted entries still exist in index.txt.
> 
> Is this expected behavior?

The index.txt file is generated from scratch on each batch run, by the worker that wins the race:
https://github.com/google/exposure-notifications-server/blob/e40434c04ef0895fa4faea17fe555f7f683aae8c/internal/export/worker.go#L279

If you are referring to the append at https://github.com/google/exposure-notifications-server/blob/e40434c04ef0895fa4faea17fe555f7f683aae8c/internal/export/worker.go#L261  that's only because the newest files generated by the worker aren't in the db yet. But the db lookup now has a TTL which will remove old files that the cleanup jobs also delete.

varshithr:
Thank you very much @gurayAlsac for the detailed explanation.
It makes understanding the implementation that much easier.

PR #462, opened 2020-05-27
proto: use full package path in go_option

johanbrandhorst:
The [official docs](https://developers.google.com/protocol-buffers/docs/reference/go-generated#invocation) for the `go_option` option specify
that the full go package import path should be used.
This, together with the `paths=source_relative` parameter, results
in the same output paths.


internal/pb/export/export.pb.go
```diff
@@ -21,12 +21,11 @@
...
 
 import (
-	reflect "reflect"
-	sync "sync"
-
 	proto "github.com/golang/protobuf/proto"
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
+	reflect "reflect"
+	sync "sync"
```
johanbrandhorst, line 28: I see that we traditionally format the generated files - this is a bit messy, IMO as it makes it harder to check if regenerating causes a diff. Would it be OK for me to change the CI so that we skip formatting of generated files? I have a helpful command that could accomplish this:
```shel
$ grep -L -R "^\/\/ Code generated .* DO NOT EDIT\.$$" --exclude-dir=.git --exclude-dir=vendor --include="*.go" . | xargs -n 1 goimports -w
```

johanbrandhorst: This skips any files that start with the "official" generated code header: https://github.com/golang/go/issues/13560#issuecomment-288457920

johanbrandhorst: I'll do this in a separate PR

sethvargo:
@gurayAlsac @squee1945 as FYI. Shouldn't impact anything.

internal/pb/export/export.pb.go
johanbrandhorst: Just to loop back on this, I couldn't find a nice way of doing it while keeping the assumption that we don't care about the precise `protoc` and `protoc-gen-go` versions used to generate the files.

PR #465, opened 2020-05-27
Exports from multiple regions

### LINKED ISSUE #445: Allow export config to read multiple regions
**Is your feature request related to a problem? Please describe.**

Export config currently only allows for a single `region` in the configuration. This means that the input and output regions have to be the same. This possibly requires up-leveling of regions at the app.

**Describe the solution you'd like**

Change export config to have `output_region` (The current `region` field is fine).

Allow for an optional array of `input_regions`

If empty, the current behvaiour should be used. If present, the `input_regions` should be used to read keys from the `exposure` table and the `output_region` should be the `region` field in the export file.

/assign @gurayAlsac

mikehelmick:
* Allow exportconfig to have a list of regions as part of the input
* Defaults to the single output region if not specfied
* Add multi input region config to admin console

/fixes #445


internal/export/database/export.go
```diff
@@ -563,13 +563,14 @@ func (db *ExportDB) FinalizeBatch(ctx context.Context, eb *model.ExportBatch, fi
...
-				Filename:   file,
-				BatchID:    eb.BatchID,
-				Region:     eb.Region,
-				BatchNum:   i + 1,
-				BatchSize:  batchSize,
-				Status:     model.ExportBatchComplete,
+				BucketName:   eb.BucketName,
+				Filename:     file,
+				BatchID:      eb.BatchID,
+				Region:       eb.Region,
```
gurayAlsac, line 569: It might be worth renaming Region to OutputRegion to spare some confusion in the future.

This is particularly true now that it has no necessary relationship to regions as they are stored in Exposure table - this is basically just a string now that gets put in export files.

mikehelmick: OK if I do that rename in code, but not in database? 

trying to avoid unnecessary DB changes at this stage.

gurayAlsac: Yeah, makes sense. Is it possibly easier to add a new column for it and no-op the current one (to eventually remove at some future time)? I've been spoiled working with tables of protos where field deprecation isn't too painful. :) 

Whichever works.

mikehelmick: updated.

PR #468, opened 2020-05-27
Add cleanup for Container Registry

sethvargo:
As we build more containers, GCR gets littered with a lot of images. This runs a job (disabled by default) that cleans up old images, keeping only the most recent 3.


mikehelmick:
/lgtm

In practice, people generally keep their old images around foreverish.

PR #469, opened 2020-05-27
ci: stop formatting generated files

johanbrandhorst:
Avoid formatting generated files by checking for
[the official generated file header](https://github.com/golang/go/issues/13560#issuecomment-288457920) before formatting.

Because everyone uses the same docker image to
generate protobuf files, this shouldn't cause much disruption.
Every time the docker image is updated the corresponding
regeneration will have to be made in the repo.

Happy to have a discussion on this as it's a bit more heavy handed than my other contributions.

### DIFF
--- a/internal/pb/export/export.pb.go
+++ b/internal/pb/export/export.pb.go
@@ -21,12 +21,11 @@
 package export
 
 import (
-	reflect "reflect"
-	sync "sync"
-
 	proto "github.com/golang/protobuf/proto"
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
+	reflect "reflect"
+	sync "sync"
 )
 
 const (
--- a/scripts/presubmit.sh
+++ b/scripts/presubmit.sh
@@ -17,8 +17,10 @@
 set -eEuo pipefail
 
 ROOT="$(cd "$(dirname "$0")/.." &>/dev/null; pwd -P)"
-SOURCE_DIRS="cmd internal tools"
-
+# Find all non-generated Go files. This works by excluding
+# files that start with the official "generated file" header.
+# See https://github.com/golang/go/issues/13560#issuecomment-288457920
+SOURCE_FILES=($(grep -L -HR "^\/\/ Code generated .* DO NOT EDIT\.$" --include="*.go" ${ROOT}))
 
 echo "🌳 Set up environment variables"
 eval $(${ROOT}/scripts/dev init)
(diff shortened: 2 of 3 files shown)


scripts/presubmit.sh
```diff
@@ -17,16 +17,23 @@
 set -eEuo pipefail
 
 ROOT="$(cd "$(dirname "$0")/.." &>/dev/null; pwd -P)"
-SOURCE_DIRS="cmd internal tools"
-
+# Find all non-generated Go files. This works by excluding
+# files that start with the official "generated file" header.
+# See https://github.com/golang/go/issues/13560#issuecomment-288457920
+mapfile -t SOURCE_FILES < <( grep -L -R "^\/\/ Code generated .* DO NOT EDIT\.$$" --include="*.go" ${ROOT} )
```
sethvargo, line 23: Is `mapfile` a common command? I don't seem to have it on my Mac.

mikehelmick: same.

johanbrandhorst: It's a [bash 4.0 (2009)](https://www.tldp.org/LDP/abs/html/bashver4.html) feature, I think it might be that the default version of bash shipping with Mac is ancient, but I understand it's practically unreasonably to expect all users to upgrade their bash version from the default one, so I'll see if I can find a workaround.

johanbrandhorst:
> Can you help me understand more about _why_ this change is necessary?

Sorry, I should've included this in the PR!

* As a regular user of go protobuf files, it was surprising to me that I was required to both generate _and_ format the files. In my experience, formatting generated files (like linting generated files from a generator you don't own), is unusual, since it isn't something you really control, and it shows diff changes that are not related to either the protofile or the generator versions.
* Having the diff check immediately after regenerating, before formatting _anything_ in the repo, should let us tell the user that the protobuf generation _specifically_, is out of date, rather than jumbling it in with all the other potential import or formatting issues.
* This also removes the extra regex we were applying to all the Go git diff operations, which might uncover some subtle formatting issues.

None of these are particularly important, but it's not insignificant, and I think it should be possible to apply with effectively zero cost to existing developers, so let me figure out a better way to create the source files array.

scripts/presubmit.sh
johanbrandhorst: Please try again, the new command should work better.

internal/pb/export/export.pb.go
```diff
@@ -21,12 +21,11 @@
...
 
 import (
-	reflect "reflect"
-	sync "sync"
-
 	proto "github.com/golang/protobuf/proto"
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
+	reflect "reflect"
+	sync "sync"
```
johanbrandhorst, line 28: This returns the protofiles to their post-generation state with no formatting applied

sethvargo:
/assign @crwilcox 

This looks fine to me, and I downloaded the code and ran the presubmit successfully on my Mac. @crwilcox was the original person who set this up, so I'd like him to give it a look.

/approve

johanbrandhorst:
I've rebased this on master

Issue #349, opened 2020-05-18
Implement diagnosis signature verification


mikehelmick:
**Describe the solution you'd like**

Currently the publish exposures API accepts `VerificationPayload` but the use is of now undefined.

This field will contain a digital signature that can be generated by a health care authority and verified by our server.

The signing protocol must be documented as part of this work

mikehelmick:
I wanted to give a quick update for anyone watching this issue.

We have a candidate design here and I'll be detailing it along with a PR to show the verification piece in the next couple of days.

PR #470, opened 2020-05-28
first draft of verification design doc

mikehelmick:
Part of #349

/hold

### DIFF
--- a/docs/design/verification_protocol.md
+++ b/docs/design/verification_protocol.md
@@ -10,7 +10,7 @@ document, but will be at a later date.
 
 The current exposure notifications publication protocol only has the ability
 to authenticate that a request came from a mobile device via the vendor
-specific device attestation protocols. This doesn't give us the ability
+specific device attestation protocols. This doesn't provide a way
 to certify that the diagnosis came from a trusted (by the server) public
 health authority in the jurisdiction.
 
--- a/docs/design/verification_protocol.md
+++ b/docs/design/verification_protocol.md
@@ -32,9 +32,9 @@ health authority in the jurisdiction.
    with calculated transmission risk value, rolling period start and rolling
 	 period count values (the key data + metadata).
 	 These values are sent to the PHA server that issued the PIN.
-4. If the PIN is valid, the PHA issues a JWT that is signed using their
-   ECDSA p-256 algorithm. The JWT includes additional claims about the data
-	 (see below).
+4. If the PIN is valid, the PHA issues a JWT that is signed using ECDSA over
+   the P-256 elliptic curve with SHA-256 as a hash function. The JWT includes
+   additional claims about the data (see below).
 5. The app on the user's device sends this signed JWT to the exposure
    notifications server (this project).
 6. If the JWT is valid and signed by a trusted PHA (verified because the public
@@ -53,12 +53,12 @@ re-calculating the HMAC that was signed by the PHA verification server. The
 The JWT itself will be unpacked and the following validations will be performed.
 First, using the standard claims.
 
-* `iss` : The issuer will be used to determine which pubic key(s) are valid for
+* `iss` : The issuer will be used to determine which public key(s) are valid for
 verification. This is to allow for key rotation.
 * `aud` : The audience must be as configured for this installation of the
 exposure notifications server.
-* `iat` : Tokens issued in the future will not be accepted.
-* `exp` : Will verify that the token is not expired.
+* `iat` : The unix timestamp at which the token was issued.
+* `exp` : The unix timestamp at which the token will expire.
 * `nbf` : If present, the "not before" timestamp will be honored.
 
 We also prescribe a set of private claims to transmit data from the PHA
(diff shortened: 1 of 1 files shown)


docs/design/verification_protocol.md
```diff
@@ -0,0 +1,117 @@
...
+
+## High Level Flow
+
+1. App user is diagnosed with Covid-19. PHA issues a PIN code to the user.
+2. The user enters the PIN code on their App
+3. The app creates a HMAC value that combines the TEKs from the device along
+   with calculated transmission risk value, rolling period start and rolling
+	 period count values (the key data + metadata).
+	 These values are sent to the PHA server that issued the PIN.
+4. If the PIN is valid, the PHA issues a JWT that is signed using their
```
sethvargo, line 35: Can you talk more about PIN generation and validation? Or is that out of scope here?

docs/design/verification_protocol.md
```diff
@@ -0,0 +1,117 @@
...
+   ECDSA p-256 algorithm. The JWT includes additional claims about the data
+	 (see below).
+5. The app on the user's device sends this signed JWT to the exposure
+   notifications server (this project).
+6. If the JWT is valid and signed by a trusted PHA (verified because the public
```
sethvargo, line 40: Maybe emphasize this is an offline verification?

sethvargo, line 40: We'll also somehow need to unpack with PHA key version did the signing (in case of rotation).

docs/design/verification_protocol.md
```diff
@@ -0,0 +1,117 @@
...
+	 key has been previously shared with the server), then the keys are imported
+	 into the server for distribution to other devices in the geography.
+
```
sethvargo, line 43: Should we store any of this verification alongside the data?

docs/design/verification_protocol.md
mikehelmick: out of scope, will be in the PHA server design

docs/design/verification_protocol.md
mikehelmick: added a `keyVersion` claim

docs/design/verification_protocol.md
mikehelmick: That's a good question.

We could store the `iss` field, as in which authority issued this diagnosis. 

We could also build in storage of the `phadata` - but as the reference sever won't make use of it, I lean against storing it. 

Good debate to be had.

sethvargo: I don't like the idea of the TEK server being used to "lookup" a verification, but it could be useful in the opposite direction (given a TEK, from where was it verified, when, etc)

mikehelmick: I'm inclined to agree that we shouldn't sore this in the reference

docs/design/verification_protocol.md
```diff
@@ -0,0 +1,122 @@
...
+diagnosis certifications from public health authorities.
+
+The actual process of issuing these certificates is not covered in this
+document, but will be at a later date.
+
+## Motivation
+
+The current exposure notifications publication protocol only has the ability
+to authenticate that a request came from a mobile device via the vendor
+specific device attestation protocols. This doesn't give us the ability
```
crwilcox, line 13:
```suggestion
specific device attestation protocols. This doesn't provide a way
```

docs/design/verification_protocol.md
```diff
@@ -0,0 +1,122 @@
...
+	 key has been previously shared with the server), then the keys are imported
+	 into the server for distribution to other devices in the geography. Because
+   we are using pre-shared public keys, this verification can be done with
+   offline verification.
+
+## JWT Verification + Accepted Claims
+
+The exposure notifications server will accept this JWT in the current
+`verificationPayload` field in the exposure reporting API. When this is present,
+device attestations should be disabled. In addition, the exposure publishing
```
crwilcox, line 50: Is 'should' correct? Or is it 'may'? If may be helpful to expand why? I agree both are not necessary.

docs/design/verification_protocol.md
```diff
@@ -0,0 +1,122 @@
...
+implementation of the exposure notifications server will disregard and never
+save the `phadata`
+* `trisk` : Contains an array of transmission risk overrides to enact when
+importing the associated keys. If data is present in this field, it will
+override the data in the upload from the device.
+* `keyVersion` : The version of the public key to use for verification.
+
+### Transmission Risk Overrides
+
+An array of JSON objects of the following structure:
```
crwilcox, line 82: would it be useful to require these be ordered? or should we specify sorting on the `sinceRollingPeriod` before computation

docs/design/verification_protocol.md
mikehelmick: I think should is correct. I also think we should move the device verification into the reference implementation of the verification server.

docs/design/verification_protocol.md
mikehelmick: we can suggest it - the server will sort upon receipt either way

PR #472, opened 2020-05-28
Drop Cloud SQL proxy and switch to VPC connections

sethvargo:
This also tunes the max instances to what each service can reasonable scale to before exhausting the maximum number of connections. I think we still need to do some additional benchmarking, but I was able to execute the following benchmark steadily:

```text
Summary:
  Total:	517.7146 secs
  Slowest:	13.7738 secs
  Fastest:	5.0447 secs
  Average:	5.1063 secs
  Requests/sec:	19.3157


Response time histogram:
  5.045 [1]	|
  5.918 [9978]	|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  6.791 [1]	|
  7.663 [1]	|
  8.536 [1]	|
  9.409 [1]	|
  10.282 [0]	|
  11.155 [10]	|
  12.028 [0]	|
  12.901 [0]	|
  13.774 [7]	|


Latency distribution:
  10% in 5.0512 secs
  25% in 5.0588 secs
  50% in 5.0923 secs
  75% in 5.1053 secs
  90% in 5.1224 secs
  95% in 5.1465 secs
  99% in 5.3603 secs

Details (average, fastest, slowest):
  DNS+dialup:	0.0008 secs, 5.0447 secs, 13.7738 secs
  DNS-lookup:	0.0000 secs, 0.0000 secs, 0.0020 secs
  req write:	0.0000 secs, 0.0000 secs, 0.0015 secs
  resp wait:	5.1054 secs, 5.0446 secs, 13.7079 secs
  resp read:	0.0000 secs, 0.0000 secs, 0.0016 secs

Status code distribution:
  [200]	10000 responses
```

I also uncovered a few small bugs when using the configurations in a brand new project.

### DIFF
--- a/terraform/service_export.tf
+++ b/terraform/service_export.tf
@@ -95,11 +95,13 @@ resource "google_cloud_run_service" "export" {
           value = google_storage_bucket.export.name
         }
       }
+
+      container_concurrency = 10
     }
 
     metadata {
       annotations = {
-        "autoscaling.knative.dev/maxScale" : "2",
+        "autoscaling.knative.dev/maxScale" : "10",
         "run.googleapis.com/vpc-access-connector" : google_vpc_access_connector.connector.id
       }
     }
--- a/terraform/service_export.tf
+++ b/terraform/service_export.tf
@@ -32,16 +32,17 @@ resource "google_service_account_iam_member" "cloudbuild-deploy-export" {
   ]
 }
 
-resource "google_project_iam_member" "export-cloudsql" {
-  project = data.google_project.project.project_id
-  role    = "roles/cloudsql.client"
-  member  = "serviceAccount:${google_service_account.export.email}"
-}
-
-resource "google_secret_manager_secret_iam_member" "export-db-pwd" {
+resource "google_secret_manager_secret_iam_member" "export-db" {
   provider = google-beta
 
-  secret_id = google_secret_manager_secret.db-pwd.id
+  for_each = toset([
+    "sslcert",
+    "sslkey",
+    "sslrootcert",
+    "password",
+  ])
+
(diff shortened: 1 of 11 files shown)


terraform/service_export.tf
```diff
@@ -85,28 +94,20 @@ resource "google_cloud_run_service" "export" {
...
-          }
-        }
       }
     }
 
     metadata {
       annotations = {
-        "autoscaling.knative.dev/maxScale" : "1000",
-        "run.googleapis.com/cloudsql-instances" : google_sql_database_instance.db-inst.connection_name
+        "autoscaling.knative.dev/maxScale" : "2",
```
mikehelmick, line 102: 2 might be too low here.

I think we should perhaps turn down the max concurrency (10) and increase the max scale also (10)

it's really for extreme cases, "catch up" or massive tenancy.

terraform/service_exposure.tf
```diff
@@ -76,15 +77,15 @@ resource "google_cloud_run_service" "exposure" {
 
     metadata {
       annotations = {
-        "autoscaling.knative.dev/maxScale" : "1000",
-        "run.googleapis.com/cloudsql-instances" : google_sql_database_instance.db-inst.connection_name
+        "autoscaling.knative.dev/maxScale" : "50",
```
mikehelmick, line 80: that's only 4000 concurrent requests total. Which is a potential risk due to the 5s latency hold.

Trying to decide if I want it changed or not though.

sethvargo: I updated the database max_connections to 100k and changed this to 500 concurrent (40k total conns)

terraform/service_export.tf
sethvargo: Done

PR #476, opened 2020-05-28
Initial implementation of integration_util.

stati0n:
Exports a utility method to bring up the database and the monolith SUT.

### DIFF
--- a/testing/integration/integration_test.go
+++ b/testing/integration/integration_test.go
@@ -1,0 +1,27 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package integration
+
+import (
+	"context"
+	"testing"
+)
+
+func TestIntegrationUtil(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+	StartSystemUnderTest(t, ctx)
+}
--- a/testing/integration/integration_test.go
+++ b/testing/integration/integration_test.go
@@ -16,15 +16,29 @@ package integration
 
 import (
 	"context"
+	"io/ioutil"
+	"net/http"
 	"testing"
 )
 
 // This tests that the integration util can bring up the Monolith.
 // It should eventually be swapped out for a smoke test that ensures
 // the server is listening at all the relevant addresses.
-func TestIntegrationUtil(t *testing.T) {
+func TestIntegrationUtilSmoke(t *testing.T) {
 	t.Parallel()
 
 	ctx := context.Background()
 	StartSystemUnderTest(t, ctx)
+	resp, err := http.Get("http://localhost:8080/health")
+	if err != nil {
+		t.Errorf("Health check generated error: %v", err)
+	}
+	ok, err := ioutil.ReadAll(resp.Body)
+	resp.Body.Close()
+	if err != nil {
+		t.Errorf("Couldn't read health check response: %v", err)
+	}
+	if string(ok) != "OK" {
+		t.Errorf("Health check did not generate response \"OK\". Actual: %v", ok)
+	}
 }
--- a/testing/integration/integration_test.go
+++ b/testing/integration/integration_test.go
@@ -19,6 +19,9 @@ import (
 	"testing"
 )
 
+// This tests that the integration util can bring up the Monolith.
+// It should eventually be swapped out for a smoke test that ensures
+// the server is listening at all the relevant addresses.
 func TestIntegrationUtil(t *testing.T) {
 	t.Parallel()
 
--- a/testing/integration/integration_util.go
+++ b/testing/integration/integration_util.go
@@ -25,6 +25,15 @@ import (
 	"github.com/google/exposure-notifications-server/internal/monolith"
 )
 
+func SetEnvAndRunServer(ctx context.Context) {
+	// Set all of these to not need to connect to external resources.
+	os.Setenv("BLOBSTORE", "FILESYSTEM")
+	os.Setenv("KEY_MANAGER", "NOOP")
+	os.Setenv("SECRET_MANAGER", "NOOP")
+
+	monolith.RunServer(ctx)
+}
+
 func StartSystemUnderTest(tb testing.TB, ctx context.Context) *monolith.MonoConfig {
 	tb.Helper()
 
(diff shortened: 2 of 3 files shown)


stati0n:
/assign @mgulimonov 

Max, here's my first thought at a test runner.  I have to do a different change first to expose the MonoConfig directly (since my struct is a logical and less complete copy), then rebase this on top, but this is generally what I'm thinking.  Still working on teardown too.

testing/integration/integration_util.go
```diff
@@ -0,0 +1,50 @@
...
+
+	if skip, _ := strconv.ParseBool(os.Getenv("SKIP_INTEGRATION_TESTS")); skip {
+		tb.Skipf("🚧 Skipping integration tests (SKIP_INTEGRATION_TESTS is set)!")
+	}
+
+	database.NewTestDatabase(tb)
+
+	monolith.RunServer(ctx)
+
+	sut := &SystemUnderTest{"http://localhost:80/do-work", "http://localhost:80/publish"}
```
jba, line 47:
```
return &SystemUnderTest{...}
```

testing/integration/integration_util.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func StartSystemUnderTest(tb testing.TB, ctx context.Context) *SystemUnderTest {
+	tb.Helper()
+
+	if testing.Short() {
+		tb.Skipf("🚧 Skipping integration tests (short!")
+	}
+
+	if skip, _ := strconv.ParseBool(os.Getenv("SKIP_INTEGRATION_TESTS")); skip {
```
jba, line 39: Check error, call  t.Fatal.

Or if you want any value to mean skip, then just check for the empty string.

sethvargo: I believe this is WAI. The default return of `ParseBool` is false. If the envvar isn't set, then `skip` is false (which is the intended behavior. If the envvar is set to a non-truthy value, it parses and returns false. Iff the envvar is set to a truthy value are tests skipped, which is the intended behavior.

testing/integration/integration_util.go
```diff
@@ -0,0 +1,47 @@
...
+		tb.Skipf("🚧 Skipping integration tests (SKIP_INTEGRATION_TESTS is set)!")
+	}
+
+	database.NewTestDatabase(tb)
```
mgulimonov, line 38: Shouldn't we pass the database connection parameters to the RunServer somehow?

testing/integration/integration_util.go
```diff
@@ -0,0 +1,47 @@
...
+
+	monoConfig, err := monolith.RunServer(ctx)
```
mgulimonov, line 40: How will test know about the url/port to use for connection to monolith?

stati0n: The monoConfig should contain that info.  Definition is here: https://github.com/google/exposure-notifications-server/blob/14f2d3aaa1103466f4ff252ae076197d90d37f5f/internal/monolith/monolith_util.go#L45-L55

testing/integration/integration_util.go
stati0n: Yes, but I haven't figured out how the Monolith gets it now.  I think it relies on the env variables set by `/scripts/dev init`.  This is still WIP, I just wanted to put something in front of you so you could see what I had in mind.

testing/integration/integration_util.go
stati0n: Removed this entirely.

testing/integration/integration_util.go
sethvargo: The connection information is shared via envvars (that's also how the monolith gets them)

testing/integration/integration_util.go
stati0n: This will stand up the Monolith server and it will be torn down at the end.  No piping of addresses back and forth yet.  I need to separate the RunServer method into two different methods, one that generates the config without starting anything, and a second that runs.

testing/integration/integration_util.go
stati0n: This is apparently my problem in one of my two dev setups.  This appears to work locally for me now.

stati0n:
/assign @sethvargo 

This works for me and seems like it's ready to go.

testing/integration/integration_test.go
```diff
@@ -0,0 +1,27 @@
...
+// limitations under the License.
+
+package integration
+
+import (
+	"context"
+	"testing"
+)
+
+func TestIntegrationUtil(t *testing.T) {
```
sethvargo, line 22: What does this do?

testing/integration/integration_util.go
```diff
@@ -0,0 +1,76 @@
...
+	"testing"
+	"time"
+
+	"github.com/google/exposure-notifications-server/internal/database"
+	"github.com/google/exposure-notifications-server/internal/monolith"
+)
+
+func SetEnvAndRunServer(tb testing.TB, ctx context.Context, dbConfig *database.Config) {
+
+	// Set all of these to not need to connect to external resources.
```
sethvargo, line 30: FYI (no changes needed), but https://github.com/google/exposure-notifications-server/pull/549 will make this easier. You can construct the config object instead of setting the envvars.

testing/integration/integration_util.go
```diff
@@ -0,0 +1,76 @@
...
+		tb.Skipf("🚧 Skipping integration tests (short!")
+	}
+
+	if skip, _ := strconv.ParseBool(os.Getenv("SKIP_INTEGRATION_TESTS")); skip {
+		tb.Skipf("🚧 Skipping integration tests (SKIP_INTEGRATION_TESTS is set)!")
+	}
+
+	db, dbconfig := database.NewTestDatabaseWithConfig(tb)
+
+	go SetEnvAndRunServer(tb, ctx, dbconfig)
```
sethvargo, line 70: We need a way to stop this, like via a cleanup function or defer.

testing/integration/integration_util.go
```diff
@@ -0,0 +1,76 @@
...
+
+	time.Sleep(10 * time.Second)
```
sethvargo, line 72: Is this just waiting for the server to start?

stati0n: Yes.  It was padded.  It could be reduced I think.

testing/integration/integration_util.go
stati0n: Needs work.

testing/integration/integration_util.go
stati0n: Exciting!

testing/integration/integration_test.go
stati0n: It demonstrates that the integration_util works.  Right now it just starts everything.  This will eventually be a smoke test or something similar.

sethvargo: Ah got it. Can you drop a comment in there nothing it's a wi

stati0n: Done.

testing/integration/integration_util.go
stati0n: Done.

testing/integration/integration_util.go
stati0n: Removed.

sethvargo:
@stati0n can we close this and build on top of GH-561 now?

PR #478, opened 2020-05-28
add shared API definitions for PHA verification protocol

mikehelmick:
shared type definitions for #349 

Reason these are in /pkg

The idea is that these type definitions will be shared with the TEK verification server

### DIFF
--- a/pkg/api/v1alpha1/verification_types.go
+++ b/pkg/api/v1alpha1/verification_types.go
@@ -1,0 +1,55 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


pkg/api/v1alpha1/verification_types.go
```diff
@@ -0,0 +1,55 @@
...
+// codebase in their alpha form.
+// These APIs are not considered to be stable at HEAD.
+package v1alpha1
+
+import "sort"
+
+const (
+	exposureKeyHMACClaim          = "tekmac"
+	healthAuthorityDataClaim      = "phadata"
+	transmissionRiskOverrideClaim = "tris"
```
crwilcox, line 25: trisk?

mikehelmick: fixed in future PR, coming soon!

PR #479, opened 2020-05-28
observability: bring back ocagent after updating to v0.0.7

### LINKED ISSUE #466: protoc-gen-go is deprecated?
Build output:

```text
WARNING: Package "github.com/golang/protobuf/protoc-gen-go/generator" is deprecated.
 A future release of golang/protobuf will delete this package,
 which has long been excluded from the compatibility promise.
```

We need to investigate why and what to do.

odeke-em:
Brings back the ocagent-exporter, after updating to
ocagent version v0.0.7 which uses newer dependencies of
the grpc-gateway, thus no longer needs the deprecated
"github.com/golang/protobuf/protoc-gen-go/generator"
package.

Fixes #466
Reverts #467

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.14
 require (
 	cloud.google.com/go v0.57.0
 	cloud.google.com/go/storage v1.6.0
+	contrib.go.opencensus.io/exporter/ocagent v0.7.0
 	contrib.go.opencensus.io/exporter/stackdriver v0.13.1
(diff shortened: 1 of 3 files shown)


mikehelmick:
/lgtm
/approve

I verified in the build log that the error no longer shows up.

Thanks for turning this around so fast in the upstream module!

odeke-em:
Thank you for the review and approval, Mike!

PR #480, opened 2020-05-29
Adding clear-export sql script

mgulimonov:


testdata/clear-export.sql
```diff
@@ -0,0 +1,7 @@
+BEGIN;
+
+DELETE FROM exportfile;
```
sethvargo, line 3: You probably want `truncate` instead

mgulimonov: Didn't know about TRUNCATE. Thanks.

mgulimonov:
>> Whats the reason for this instead of just killing the contianer?
Having script like this is useful for manual testing and e2e testing to avoid restarting entire application.
Later e2e tests will restart app and db automatically but for now it's not the case.

sethvargo:
If the test data is large, it's going to be _slower_ to truncate all the tables (because it rebuilds the indexes) vs just kicking the container.

mgulimonov:
Agreed, but this won't be the case for dev environment. My current use case is to create ~100-1000 keys, 10 export batches do the tests and then clear the tables.
There is no much sense in recreating DB container for each test run.

sethvargo:
> There is no much sense in recreating DB container for each test run.

I disagree 😄 . Using a new container for each test is relatively low overhead (milliseconds) and gives us a ton of free benefits like the ability to run tests in parallel, no need to worry about pre-existing state, ability to easily test against different Postgres version.

mikehelmick:
> Agreed, but this won't be the case for dev environment. My current use case is to create ~100-1000 keys, 10 export batches do the tests and then clear the tables.
> There is no much sense in recreating DB container for each test run.

In this case, do you really want to delete the entire DB or just clear certain tables?

I wouldn't want my authorized apps and export configs to go away (for example)

mgulimonov:
This is exactly my point:
I don't want to clear entire DB, only certain tables (related to exposure and export),
This is also why I don't want to recreate the container. 

Once we have a good seeding mechanism that will be used by automatic tests, we still might want to have this script to remove just a portion of data.

PR #490, opened 2020-05-29
Add tests for GCS interface

sethvargo:


sethvargo:
These tests _should_ fail because we don't set `GOOGLE_CLOUD_PROJECT` on prow...

sethvargo:
Excellent, I have predicted test failures. @mikehelmick @crwilcox we should chat quickly about how to run these on CI.

sethvargo:
After talking in standup, we're gonna refactor this a bit to use a single bucket and prefix filenames with random bits.

PR #492, opened 2020-05-29
Support runtime configuration of all current components

sethvargo:
This resolves most TODOs in setup by resolving runtime configuration values for:

- Key Management
- Secret Management
- Storage (already done, but refactored a bit)
- Observability (already done, but tuned a bit)

And documents it 😄.


internal/admin/config.go
```diff
@@ -23,24 +23,45 @@ import (
...
 
-var _ setup.DBConfigProvider = (*Config)(nil)
+var _ setup.DatabaseConfigProvider = (*Config)(nil)
+var _ setup.KeyManagerConfigProvider = (*Config)(nil)
+var _ setup.SecretManagerConfigProvider = (*Config)(nil)
+var _ setup.BlobstoreConfigProvider = (*Config)(nil)
 
 type Config struct {
+	Database      *database.Config
+	KeyManager    *signing.Config
```
mikehelmick, line 39: is the admin console using these dependencies now?

mikehelmick: well, I guess implicitly secret manager was being used before.

sethvargo: Yea I put Secret Manager everywhere we have a database. I can remove KeyManager, but it doesn't hurt...

mikehelmick: agreed, doesn't hurt. may be useful in the future even :)

PR #493, opened 2020-05-30
implement verification and processing of health authority diagnosis v…

mikehelmick:
…erification certificates.

Next step in #349 

This adds 2 new database tables

1. HealthAuthority - represents an individual health authority
2. HealthAuthorityKey - represents key versions. Each time a new signing key is issued, the public key and corresponding version will need to be inserted into this table.

AuthorizedApps can now be scoped to a set of HealthAuthorities. This allows for there to be multi diagnosis verification authorities within a state (Say a university medical system and state dept. of health).

- This is NOT hooked up to the live API yet.
- Happy path tests only

/cc @peetsasaki


internal/authorizedapp/database/authorized_app.go
```diff
@@ -47,12 +47,12 @@ func (aa *AuthorizedAppDB) InsertAuthorizedApp(ctx context.Context, m *model.Aut
 		result, err := tx.Exec(ctx, `
 			INSERT INTO
 				AuthorizedApp
-			  (app_package_name, platform, allowed_regions,
+			  (app_package_name, platform, allowed_regions, health_authority_ids,
```
sethvargo, line 50: Unrelated, but do we have a long-term strategy for safer database migrations? Right now the app always has to be in complete sync.

internal/authorizedapp/database/authorized_app_test.go
```diff
@@ -134,29 +135,31 @@ func TestGetAuthorizedApp(t *testing.T) {
...
 				SafetyNetBasicIntegrity:  true,
 				SafetyNetCTSProfileMatch: true,
 			},
 		},
 		{
 			name: "all_regions",
 			sql: `
-				INSERT INTO AuthorizedApp (app_package_name, platform, allowed_regions)
-				VALUES ($1, $2, $3)
+				INSERT INTO AuthorizedApp (app_package_name, platform, allowed_regions, health_authority_ids)
```
sethvargo, line 154: This mismatch between the table column and struct field is a bit weird.

internal/publish/model/exposure_model.go
```diff
@@ -67,9 +67,10 @@ type Publish struct {
 	Keys                      []ExposureKey `json:"temporaryExposureKeys"`
 	Regions                   []string      `json:"regions"`
 	AppPackageName            string        `json:"appPackageName"`
-	Platform                  string        `json:"platform"`
-	DeviceVerificationPayload string        `json:"deviceVerificationPayload"`
+	Platform                  string        `json:"platform"`                  // DEPRECATED
```
sethvargo, line 70: Maybe move these deprecated ones to the bottom of the struct?

internal/verification/database/health_authority_db.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return nil, err
+	}
+	ha.Keys = haks
+
+	return &ha, nil
+}
+
+func (db *HealthAuthorityDB) AddHealthAuthorityKey(ctx context.Context, ha *model.HealthAuthority, hak *model.HealthAuthorityKey) error {
+	if err := hak.Validate(); err != nil {
```
sethvargo, line 96: It's probably fine, but both ha and hak could be nil, in which case this would panic

internal/verification/database/health_authority_db.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	var haks []*model.HealthAuthorityKey
+	for rows.Next() {
+		if rows.Err() != nil {
+			return nil, rows.Err()
+		}
+		var hak model.HealthAuthorityKey
+		var thru *time.Time
+		if err := rows.Scan(&hak.AuthorityID, &hak.Version, &hak.From, &thru, &hak.PublicKeyPEM); err != nil {
```
sethvargo, line 148: Maybe extract this into a `scanOneHAKey` heklper like we did elsewhere?

internal/verification/model/health_authority.go
```diff
@@ -0,0 +1,94 @@
...
+	"errors"
+	"fmt"
+	"reflect"
+	"time"
+)
+
+// HealthAuthority represents a public health authority that is authorized to
+// issue diagnosis verification certificates accepted by this server.
+type HealthAuthority struct {
+	ID       int64  `db:"id"`
```
sethvargo, line 30: We don't use the db fields afaik. Can you remove them to avoid confusion?

internal/verification/model/health_authority.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	if ha.Name == "" {
+		return errors.New("name cannot be empty")
+	}
+	return nil
+}
+
+// HealthAuthorityKey represents a public key version for a given health authority.
+type HealthAuthorityKey struct {
+	AuthorityID  int64     `db:"health_authority_id"`
```
sethvargo, line 53: Same note on tags

internal/verification/model/health_authority.go
```diff
@@ -0,0 +1,94 @@
...
+func (k *HealthAuthorityKey) Validate() error {
+	if _, err := k.PublicKey(); err != nil {
+		return fmt.Errorf("invalid public key PEM block: %w", err)
+	}
+	return nil
+}
+
+// IsValid returns true if the key is valid based on the current time.
+func (k *HealthAuthorityKey) IsValid() bool {
+	return k.IsValidAt(time.Now().UTC())
```
sethvargo, line 70: I thought we moved away from UTC?

internal/verification/model/health_authority.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	pub, err := x509.ParsePKIXPublicKey(block.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("x509.ParsePKIXPublicKey: %w", err)
+	}
+	switch pub := pub.(type) {
+	case *ecdsa.PublicKey:
+		return pub, nil
+	}
+	return nil, fmt.Errorf("unsupported public key type: %v", reflect.TypeOf(pub))
```
sethvargo, line 93: You can switch this around to avoid reflect:

```go
switch typ := pub.(type) {
case *ecdsa.PublicKey:
  return pub, nil
default:
  return nil, fmt.Errorf("unsupported public key type: %T", typ)
}
```

internal/verification/model/health_authority.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// PublicKey decodes the PublicKeyPEM text and returns the `*ecdsa.PublicKey`
+// This system only supports verifying ECDSA JWTs, `alg: ES256`
+func (k *HealthAuthorityKey) PublicKey() (*ecdsa.PublicKey, error) {
+	block, _ := pem.Decode([]byte(k.PublicKeyPEM))
+	if block == nil || block.Type != "PUBLIC KEY" {
+		return nil, errors.New("unable to decode PEM block containing PUBLIC KEY")
+	}
+	pub, err := x509.ParsePKIXPublicKey(block.Bytes)
```
sethvargo, line 85: This is fine so long as it's actually formatted that way.

internal/verification/phaverify.go
```diff
@@ -0,0 +1,115 @@
...
+
+	if !token.Valid {
+		return nil, fmt.Errorf("invalid verificationPayload")
+	}
+	if err := claims.Valid(); err != nil {
+		return nil, err
+	}
+
+	// JWT is valid and signature is valid.
+	if _, ok := authApp.AllowedHealthAuthorities[healthAuthorityID]; !ok {
```
sethvargo, line 94: Maybe note that this is after the auth check to prevent a timing attack?

internal/verification/phaverify.go
```diff
@@ -0,0 +1,115 @@
...
+		return nil, fmt.Errorf("app %v has not authorized health authority issuer: %v", authApp.AppPackageName, claims.Issuer)
+	}
+
+	// Verify the HMAC.
+	jwtHMAC, err := base64util.DecodeString(claims.SignedMAC)
+	if err != nil {
+		return nil, fmt.Errorf("error decoding HMAC from claims: %w", err)
+	}
+	secret, err := base64util.DecodeString(publish.HMACKey)
```
sethvargo, line 103: I didn't see any docs or requirements that the HMAC is b64. We should add some tests or validation.

internal/verification/phaverify.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error decoding HMAC secret from publish request: %w", err)
+	}
+	wantHMAC, err := utils.CalculateExposureKeyHMAC(publish.Keys, secret)
```
sethvargo, line 107: Unchecked err

internal/verification/phaverify_test.go
```diff
@@ -0,0 +1,154 @@
...
+	coredb "github.com/google/exposure-notifications-server/internal/database"
+	pubmodel "github.com/google/exposure-notifications-server/internal/publish/model"
+	"github.com/google/exposure-notifications-server/internal/verification/database"
+	"github.com/google/exposure-notifications-server/internal/verification/model"
+
+	verifyapi "github.com/google/exposure-notifications-server/pkg/api/v1alpha1"
+	utils "github.com/google/exposure-notifications-server/pkg/verification"
+)
+
+// TODO(mikehelmick) : this only tests the happy path. Write non happy tests.
```
sethvargo, line 43: https://www.youtube.com/watch?v=y6Sxv-sUYtM

pkg/api/v1alpha1/verification_types.go
```diff
@@ -53,3 +57,18 @@ func (a TransmissionRiskVector) Less(i, j int) bool {
 func (a TransmissionRiskVector) Swap(i, j int) {
 	a[i], a[j] = a[j], a[i]
 }
+
+type VerificationClaims struct {
+	PHAClaims         map[string]string      `json:"phadata"`
+	TransmissionRisks TransmissionRiskVector `json:"trisk"`
+	SignedMAC         string                 `json:"tekmac"`
+	KeyVersion        string                 `json:"keyVersion"`
+	jwt.StandardClaims
```
sethvargo, line 66: Pointer

pkg/verification/utils.go
```diff
@@ -0,0 +1,50 @@
...
+		return nil, fmt.Errorf("cannot calculate hmac on empty exposure keys")
+	}
+	// Sort by the key
+	sort.Slice(keys, func(i int, j int) bool {
+		return strings.Compare(keys[i].Key, keys[j].Key) <= 0
+	})
+
+	// Build the cleartext.
+	perKeyText := []string{}
+	for _, ek := range keys {
```
sethvargo, line 40: We might want to optimize this depending on how frequently it's called (lots of allocs right now)

internal/authorizedapp/database/authorized_app.go
mikehelmick: To do that, we would have to start cutting releases on the repo and provide instructions about code + DB migration order.

internal/authorizedapp/database/authorized_app_test.go
mikehelmick: aligned the names... by changing both.

internal/verification/database/health_authority_db.go
mikehelmick: will do if I add another query method. Right now this is the only one.

internal/verification/model/health_authority.go
mikehelmick: removed - technically didn't need to be there anyway since time does the right thing.

internal/verification/model/health_authority.go
mikehelmick: That's part of the reason I'm being prescriptive on key type.

internal/verification/phaverify.go
mikehelmick: it's in the design

https://github.com/google/exposure-notifications-server/blob/master/docs/design/verification_protocol.md#jwt-verification--accepted-claims

I will go through and add more tests after this

pkg/api/v1alpha1/verification_types.go
mikehelmick: This is the way the documentation specifies
https://godoc.org/github.com/dgrijalva/jwt-go#example-NewWithClaims--CustomClaimsType

Nil pointer errors in use otherwise.

pkg/verification/utils.go
mikehelmick: changed slice allocation.

sherifkozman:
@sethvargo is it part of the scope for #349 and #493 that an HA would be able to approve/verify a Payload by the user to be published/uploaded or not. To avoid false payloads to be ingested into the system without an HA approving it ?

PR #499, opened 2020-05-31
Fix nil pointer on environment setup.

mikehelmick:
If the config structs, if there are pointers to sub-configs, those pointers are initialized to nil and aren't traversed by envconfig.

### DIFF
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -71,11 +71,14 @@ func Setup(ctx context.Context, config DatabaseConfigProvider) (*serverenv.Serve
 	var sm secrets.SecretManager
 	if provider, ok := config.(SecretManagerConfigProvider); ok {
 		smConfig := provider.SecretManagerConfig()
+		if err := envconfig.Process(ctx, smConfig, nil); err != nil {
+			return nil, nil, fmt.Errorf("unable to process secret manager environment: %w", err)
+		}
 
 		var err error
 		sm, err = secrets.SecretManagerFor(ctx, smConfig.SecretManagerType)
 		if err != nil {
-			return nil, nil, fmt.Errorf("unable to connect to secret manager: %v", err)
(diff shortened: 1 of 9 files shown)


mikehelmick:
prior to this 

```
❯ go run ./cmd/exposure 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1a711a0]

goroutine 1 [running]:
github.com/google/exposure-notifications-server/internal/setup.Setup(0x1fea980, 0xc000042158, 0x1fc4a80, 0xc0000b12c0, 0x0, 0x0, 0x0, 0x0)
	/Users/mhelmick/go/src/github.com/google/exposure-notifications-server/internal/setup/setup.go:76 +0xe70
main.main()
	/Users/mhelmick/go/src/github.com/google/exposure-notifications-server/cmd/exposure/main.go:37 +0xed
exit status 2
```

internal/setup/setup.go
```diff
@@ -71,11 +71,14 @@ func Setup(ctx context.Context, config DatabaseConfigProvider) (*serverenv.Serve
 	var sm secrets.SecretManager
 	if provider, ok := config.(SecretManagerConfigProvider); ok {
 		smConfig := provider.SecretManagerConfig()
+		if err := envconfig.Process(ctx, smConfig, nil); err != nil {
```
sethvargo, line 74: Do we need this here? It's down on line 91. Won't that re-process?

mikehelmick: we need to pre-process the secret manager config, so the default (or overridden) secret manager can be used to process the rest of the configs.

If you try to use a secret:// to configure secret manager, you've broken the space time continuum.

sethvargo: https://github.com/google/exposure-notifications-server/pull/500

PR #503, opened 2020-05-31
Provide a way to return config from the test database

sethvargo:
Going to use this in some upcoming stuff, but here's an easier PR to review 😄

### DIFF
--- a/internal/database/database_util.go
+++ b/internal/database/database_util.go
@@ -43,7 +43,7 @@ import (
 //
 // All database tests can be skipped by running `go test -short` or by setting
 // the `SKIP_DATABASE_TESTS` environment variable.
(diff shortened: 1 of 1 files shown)


internal/verification/database/health_authority_db_test.go
```diff
@@ -77,7 +77,7 @@ func TestAddRetrieveHealthAuthorityKeys(t *testing.T) {
 	}
 
 	wantKeys := []*model.HealthAuthorityKey{
-		&model.HealthAuthorityKey{
+		{
```
mikehelmick, line 80: covered in https://github.com/google/exposure-notifications-server/pull/502

sethvargo: Done

PR #504, opened 2020-05-31
Add tests for setup

sethvargo:
Add tests for setup


cmd/export/main.go
```diff
@@ -17,37 +17,102 @@ package main
...
+}
+
+func realMain() error {
+	ctx := context.Background()
+
+	config, env, closer, err := doSetup()
+	if err != nil {
+		return fmt.Errorf("setup: %w", err)
+	}
+	defer closer()
```
sethvargo, line 49: Having this defer in `main` will _never_ run because `logger.Fatal` calls `os.Exit` and that doesn't run defers.

cmd/export/main.go
```diff
@@ -17,37 +17,102 @@ package main
...
+
+	server, err := NewServer(ctx, config, env)
+	if err != nil {
+		return fmt.Errorf("newserver: %w", err)
+	}
+
+	go server.Run()
```
sethvargo, line 56: I'm not really sure what the best design is here. Cloud Run seems to send SIGKILL (which we can't trap), but that also means we have no opportunity to cleanup connections.

cmd/export/main.go
```diff
@@ -17,37 +17,102 @@ package main
...
+	return nil
+	// return server.Stop()
+}
+
+func doSetup() (*export.Config, *serverenv.ServerEnv, setup.Defer, error) {
```
sethvargo, line 61: I don't like this, but I can't think of a better way to individually test "config/environment parsing".

cmd/export/main_test.go
```diff
@@ -0,0 +1,150 @@
...
+		os.Setenv("DB_SSLMODE", databaseConfig.SSLMode)
+		os.Setenv("DB_PASSWORD", databaseConfig.Password)
+
+		config, env, closer, err := doSetup()
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer closer()
+
+		// Database
```
sethvargo, line 49: At some point, I feel like we're testing envconfig here 🤒 , but I'm not sure how else we would have caught the nil pointer bug from #499

cmd/export/main.go
mikehelmick: known issue.
Also cloud run stops container instances with an un-trappable signal.

This is there in case someone wanted to let this have a more graceful shutdown.

cmd/export/main.go
mikehelmick: correct - I've already relayed this feedback that graceful shutdown signals are needed.

cmd/export/main.go
```diff
@@ -17,37 +17,102 @@ package main
...
 func main() {
 	ctx := context.Background()
 	logger := logging.FromContext(ctx)
 
+	if err := realMain(); err != nil {
+		logger.Fatal(err)
+	}
+}
+
+func realMain() error {
```
mikehelmick, line 42: if we're extracting this out for testability, it should probably go into /internal

cmd/export/main_test.go
mikehelmick: we could have caught that with unit tests on internal setup

actually that could be sufficient

internal/setup/setup_test.go
```diff
@@ -0,0 +1,201 @@
+// Copyright 2020 Google LLC
```
mikehelmick, line 1: yeah - maybe just submit this, and I think it closes out the "things that could go wrong" category.

The mains are as simple as they can be currently.

sethvargo: Done

PR #509, opened 2020-06-01
ci: stop formatting generated files

johanbrandhorst:
Avoid formatting generated files by checking for
the official [1] generated file header before formatting.

Because everyone uses the same docker image to
generate protobuf files, this shouldn't cause much disruption.
Every time the docker image is updated the corresponding
regeneration will have to be made in the repo.

[1]: https://github.com/golang/go/issues/13560#issuecomment-288457920

This version skips the generation of the array altogether, since it was not working in older versions of bash found on Macs.

### DIFF
--- a/scripts/presubmit.sh
+++ b/scripts/presubmit.sh
@@ -17,8 +17,6 @@
 set -eEuo pipefail
 
 ROOT="$(cd "$(dirname "$0")/.." &>/dev/null; pwd -P)"
-SOURCE_DIRS="cmd internal tools"
-
 
 echo "🌳 Set up environment variables"
 eval $(${ROOT}/scripts/dev init)
@@ -26,7 +24,12 @@ eval $(${ROOT}/scripts/dev init)
 
 echo "🚒 Verify Protobufs are up to date"
 ${ROOT}/scripts/dev protoc
-# Don't verify generated pb files here as they are tidied later.
+git diff *.go | tee /dev/stderr | (! read)
+if [ $? -ne 0 ]; then
+   echo "✋ Found uncommited changes after regenerating Protobufs."
+   echo "✋ Commit these changes before merging."
+   exit 1
+fi
 
 
 echo "🧽 Verify goimports formattting"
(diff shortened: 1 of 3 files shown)


johanbrandhorst:
Hi @mikehelmick and @sethvargo, thanks for your patience, could you please verify that this does what it should on your machines? I've tested it on my own machine and bash 3.2.57 (in a container) and I hope it should be fine this time :grimacing:.

scripts/presubmit.sh
```diff
@@ -38,11 +41,12 @@ if [ $? -ne 0 ]; then
    echo "✋ to enable import cleanup. Import cleanup skipped."
 else
    echo "🧽 Format with goimports"
-   goimports -w $(echo $SOURCE_DIRS)
+   # Find all non-generated Go files. This works by excluding
+   # files that start with the official "generated file" header.
+   # See https://github.com/golang/go/issues/13560#issuecomment-288457920
+   grep -L -HR "^\/\/ Code generated .* DO NOT EDIT\.$" --include="*.go" "${ROOT}" | xargs goimports -w
```
mikehelmick, line 47: returns no results on my machine

```
✦ ❯ grep -L -HR "^\/\/ Code generated .* DO NOT EDIT\.$" --include="*.go" "${ROOT}"                     

```

johanbrandhorst: Alright, I give up. That doesn't make any sense. Is `${ROOT}` set when running this? Did you try running `presubmit.sh`? What version of `grep` is installed on Mac? Sorry for the inconvenience :disappointed:, again, I really did think it would work this time.

johanbrandhorst:
The `prow-oss` error looked unrelated to the changes, so I've pushed a rebase. Really happy to see a public CI log, that should help debugging this. I tried running the grep command on the presubmit docker image without issues, so I think this should work in CI, but I'd still like to confirm whether it works on MacOS. Any volunteers :grin:?

scripts/presubmit.sh
mikehelmick: ok, verified this is working for me now.

johanbrandhorst:
Thanks a lot Mike!

PR #511, opened 2020-06-01
Enable to use scripts/presubmit.sh as a pre-push git hook

alcortesm:
The CONTRIBUTING.md file suggests the presubmit script can be used as a
pre-push git hook but it fails because the script relies on being stored
in a directory directly below the project root and the git hooks are
stored two levels below it.

The current patch allows to run the checks as a pre-push git hook while
still being able to run it by direct invocation.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,16 +57,6 @@ $ go test ./...
 You should run the presubmit checks before committing changes. The presubmit script
 is located at `scripts/presubmit.sh`.
 
-You can add a prepush hook by linking to the presubmit script to automatically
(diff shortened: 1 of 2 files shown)


sethvargo:
Hi @alcortesm 

We should remove this recommendation from the README instead.

alcortesm:
Ok, I'll add a new commit restoring the previous changes and removing the recommendation.

alcortesm:
Hi @sethvargo 

Please take a look now, a squash-merge might be in order.

sethvargo:
Thanks!

alcortesm:
You are welcome! 😀

PR #518, opened 2020-06-01
plugged aws-s3 adapter

asutosh97:
+ fixed s3 docstrings
  
 + plugged-in AWSS3 blob adapter
  
 + added AWS S3 to docs


sethvargo:
/lgtm
/approve

Thanks for catching this.

PR #521, opened 2020-06-01
Removing time window used by worker to query for exposure batches available for export.

mgulimonov:


mgulimonov:
I'm open to suggestions of a better name for this configuration parameter.

internal/export/config.go
```diff
@@ -47,6 +47,8 @@ type Config struct {
 	TruncateWindow time.Duration `envconfig:"TRUNCATE_WINDOW" default:"1h"`
 	MinWindowAge   time.Duration `envconfig:"MIN_WINDOW_AGE" default:"2h"`
 	TTL            time.Duration `envconfig:"CLEANUP_TTL" default:"336h"`
+	// The worker considers export batches closed before the now() minus BatchCloseTimeShift minutes ago.
+	BatchCloseTimeShift time.Duration `envconfig:"BATCH_CLOSE_TIME_SHIFT" default:"5m"`
```
mikehelmick, line 51: I think this whole concept is obsoleted by the MinWindowAge (above).

mgulimonov: I don't quite understand. Should we eliminate this timeshift, or should we replace BatchCloseTimeShift param by MinWindowAge?

internal/export/config.go
```diff
@@ -47,6 +47,8 @@ type Config struct {
 	TruncateWindow time.Duration `envconfig:"TRUNCATE_WINDOW" default:"1h"`
 	MinWindowAge   time.Duration `envconfig:"MIN_WINDOW_AGE" default:"2h"`
 	TTL            time.Duration `envconfig:"CLEANUP_TTL" default:"336h"`
+	// The worker considers export batches closed before the now() minus BatchCloseTimeShift minutes ago.
```
crwilcox, line 50: I have a hard time comprehending this comment. I am not entirely sure my recommendation here is any better though.

```suggestion
	// The worker considers export batches closed during the last BatchCloseTimeShift minutes.
```

internal/export/config.go
mikehelmick: when minWindowAge was added it makes it so the batch isn’t even created until MinWindowAge is past, so the waiting an extra 5 minutes never comes into play.

@gurayAlsac to confirm - but I think we can remove the 5 minute delay which makes this PR unnecessary. and for testing set MinWindowAge to a smaller value

mgulimonov: @mikehelmick Ok, then I'm removing the time window logic from the worker.go
PTAL.

gurayAlsac: What are we trying to achieve with an export batch time shift?

mikehelmick: @gurayAlsac  - refresh - we removed it. and removed the old 5 min padding you had added, I don't think it's necessary any longer

squee1945: I believe the original issue was when the truncate window was larger than the export period (e.g., for test scenarios). In that case, batches would be created every, e.g., 1 minute as the minute ended, but the truncate window might still be open, e.g,. 15 minutes. If the batch was processed, it would miss keys that were still arriving in the open truncate window.

mikehelmick: ok - That confirms that this is no longer necessary.

since we don't create the window until 2 hours after it's close.

PR #528, opened 2020-06-02
Does this make CI run faster?

sethvargo:


sethvargo:
@mikehelmick plz re-review. Shaved off about 30s.

Issue #530, opened 2020-06-03
Design and implement revocation certificates


mikehelmick:
### TL;DR

After receiving a publish request, respond with a _revocation certificate_

The revocation certificate can be used to later delete a set of published TEKs from the server's database. This does not remove the TEKs from already generated exports.

### Design

**Proposal**

On receiving a batch of TEKs - return an encrypted token that contains the those TEKs.

If that token is returned to the server, delete those TEKs from primary storage.

The token must be encrypted because anyone intercepting it should not be able to determine what the TEKs covered by it are.

/kind enhancement
/assign

mikehelmick:
This is currently on hold. Will revisit if this becomes necessary.

/close

Issue #495, opened 2020-05-30
OpenAPI/Swagger Specs for API endpoints


sherifkozman:
### Question
Is there any OpenAPI/Swagger specs available for the services enpoints.
To enhance security deployment for public facing services we are testing exposing them through Cloud endpoints with additional authenticaion layer. Cloud endpoints require OpenAPI specs to define the different endpoints
thanks

mikehelmick:
/kind enhancement

Not at this time, but something we can add.

This codebase intentionally doesn’t include authententication as we don’t provide any kind of account / identity managment or integration.

Also take a look at #349 - We are in the process of rolling out diagnisis verification through signed JWTs which would provide authentication, but is handled within the server.

sherifkozman:
Great, I will keep an eye on that as well.

I understand the reason behind not implementing authentication this is why we were exploring cloud endpoints as an external implementation that isn’t coupled with the server, hence the openapi spec.

sethvargo:
Hi @sherifkozman are you looped into the plans re: #493. The longer-term plan is to allow PHAs to employ their own authentication requirements via a token verification system.

sherifkozman:
@sethvargo i wasnt but @mikehelmick looped me onto #349 , i took a look at #493 however requesting OpenAPI specs was in relation to API management and endpoint security, monitoring through cloud endpoints which doesnt conflict with the token verification system proposal.

Thanks for keeping me looped in as we are testing the deployment in different modes so we can enable more HAs to be able to deploy it in various environment configurations.

mikehelmick:
Not quite swagger - but we have this in a common package now that can be imported into other codebases.

sherifkozman:
Thanks @mikehelmick thats useful for devs who will be willing to build on top of the work done.
My main interest in OpenAPI spec or swagger was in relation defining and  running Cloud endpoints on top of the Cloud run services to enable more control over the exposure Cloud Run Service.
This would enable an HA to add an additional layer for API control and monitoring either with Cloud run or within an GKE deployment.

lmartensson:
Thanks @mikehelmick - is there any progress on the API specification Sherif is requesting?
I just spent well over an hour trying to dig through the code and help one of our app developers figure out what endpoints he can call, but I'm really not doing too well.

mikehelmick:
Hi - no progress to report. 

We do have our documentation that talks about how to call the publish API: https://google.github.io/exposure-notifications-server/
[...]

PR #532, opened 2020-06-03
refactor API definitions to PKG instead of internal so they can be in…

mikehelmick:
…cluded via our go mod.

Setup for #530
And not quite what is asked for in #495  - but makes the structs for the API importable via the pkg/api package.


mikehelmick:
/retest

Hoping we don't have a flake.

mikehelmick:
resolved merge conflict.

PR #540, opened 2020-06-03
[master] Format markdown

mattmoor:
Produced via:
  `prettier --write --prose-wrap=always $(find -name '*.md' | grep -v vendor | grep -v .github | grep -v docs/cmd/)`
/assign helmick
/cc helmick

### DIFF
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -30,26 +30,26 @@ Each service's `main` package is located in the `cmd` directory.
 Each service is deployed in the same way, but may accept different configuration
 options. Configuration options are specified via environment variables.
 
-| Service          | Folder               | Description |
-|------------------|----------------------|-------------|
+| Service          | Folder               | Description                                                             |
+| ---------------- | -------------------- | ----------------------------------------------------------------------- |
 | cleanup-export   | cmd/cleanup-export   | Deletes old exported files published by the exposure key export service |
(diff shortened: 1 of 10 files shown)


docs/deploying.md
```diff
@@ -81,13 +83,13 @@ order of operations is:
 Build new services by using the script at `./scripts/build`, specifying the
 following values:
 
--   `PROJECT_ID` (required) - your Google Cloud project ID.
```
sethvargo, line 84: This indentation is intentional, because then nested fields (e.g. codeblocks inside a list) align correctly.

sethvargo: (everywhere)

terraform/README.md
```diff
@@ -12,16 +12,17 @@ continuous deployment.
 
 - gcloud. [Installation guide](https://cloud.google.com/sdk/install)
 
-    Note: Make sure you **unset** `GOOGLE_APPLICATION_CREDENTIALS` in your
```
sethvargo, line 15: This is incorrect for GHFM.

docs/deploying.md
mikehelmick: these are robotic changes - there is nobody on the other end to adjust

I can ask matt to turn off the markdown formatter.

mattmoor:
it's off

Issue #537, opened 2020-06-03
AppPackage/BundleID is case sensisitive 


mikehelmick:
### TL;DR

Currently the appPackageNAme in the API (dual homed for BundleID) is case sensitive when loading from the database / appconfig cache.

This should be case insensitive.

/kind good-first-issue

PR #541, opened 2020-06-03
Make AppPackageName and BundleID case-insensitive

sethvargo:
Fixes GH-537

### DIFF
--- a/internal/export/database/export.go
+++ b/internal/export/database/export.go
@@ -209,7 +209,7 @@ func (db *ExportDB) AddSignatureInfo(ctx context.Context, si *model.SignatureInf
  				SignatureInfo
 				(signing_key, app_package_name, bundle_id, signing_key_version, signing_key_id, thru_timestamp)
 			VALUES
-				($1, $2, $3, $4, $5, $6)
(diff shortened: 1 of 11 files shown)


internal/export/database/export.go
```diff
@@ -253,10 +253,10 @@ func (db *ExportDB) ListAllSigntureInfos(ctx context.Context) ([]*model.Signatur
 
 	rows, err := conn.Query(ctx, `
     SELECT
-      id, signing_key, app_package_name, bundle_id, signing_key_version, signing_key_id, thru_timestamp
+      id, signing_key, LOWER(app_package_name), LOWER(bundle_id), signing_key_version, signing_key_id, thru_timestamp
```
sethvargo, line 256: This could/might affect export generation. Any ideas?

mikehelmick: @gurayAlsac - any insight on if this impacts export importing?

gurayAlsac: All lower is what we were expecting anyway in the EN API public key configuration lookups, and we don't look at app bundle id at all.

Issue #525, opened 2020-06-02
Separate envvar from secret parsing


sethvargo:
### TL;DR
envconfig is overloaded. Now that a bunch of things are configurable, we have complicated inter-graph dependencies that are unnecessary.

### Design

**Proposal**
1. Process configs using `kelseyhightower/envconfig`
1. Iterate over configs using reflection to find any `string` or string-like fields that begin with `secret://`
  - Alternative design: require `secret` struct tags, but that seems like it will lead to fun edge cases
1. Parse all found secrets and update them in the struct

**Alternatives considered**
Keep it as is, but we've uncovered quite a few bugs with the current implementation.

Issue #510, opened 2020-06-01
Setup tests are failing in environments without default GCP credentials


johanbrandhorst:
### TL;DR
Running `go test ./internal/setup` fails in certain user environments. We should either make these tests work in any environment or only run them under certain conditions (flags, build tags?).

**Expected behavior**
Tests should pass regardless of external environments.

**Observed behavior**
Tests failed because of local configuration missing.

### Reproduction
Run the setup tests in a clean environment:
```shell
$ go test ./internal/serverenv/
--- FAIL: TestSetup (0.00s)
    --- FAIL: TestSetup/blobstore (5.24s)
        setup_test.go:166: unable to connect to storage system: storage.NewClient: dialing: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
    --- FAIL: TestSetup/key_manager (5.49s)
        setup_test.go:144: unable to connect to key manager: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
    --- FAIL: TestSetup/secret_manager (6.44s)
        setup_test.go:122: unable to connect to secret manager: secretmanager.NewClient: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
FAIL
FAIL    github.com/google/exposure-notifications-server/internal/setup  6.730s
FAIL
```

**Environment**
- OS: Linux amd64

**Additional information**
I don't have any default GCP credentials configured.

I'd be happy to submit a fix for this, if a resolution is agreed.

johanbrandhorst:
This is still happening on latest master:
```shell
$ git rev-parse HEAD
9ec98717e3e7255529f4b8980d69543c12acf400
$ go test ./...
?       github.com/google/exposure-notifications-server/cmd/cleanup-export      [no test files]
?       github.com/google/exposure-notifications-server/cmd/cleanup-exposure    [no test files]
?       github.com/google/exposure-notifications-server/cmd/export      [no test files]
?       github.com/google/exposure-notifications-server/cmd/exposure    [no test files]
?       github.com/google/exposure-notifications-server/cmd/federationin        [no test files]
?       github.com/google/exposure-notifications-server/cmd/federationout       [no test files]
?       github.com/google/exposure-notifications-server/cmd/generate    [no test files]
?       github.com/google/exposure-notifications-server/cmd/monolith    [no test files]
?       github.com/google/exposure-notifications-server/internal/admin  [no test files]
[...]

PR #549, opened 2020-06-04
Use custom envconfig resolver, decouple secrets

sethvargo:
This PR is rather large.

- **Write our own envconfig parser.** There are a few challenges with the current envconfig. It's not extensible (we already wrap it with our own custom logic), and it only works with envvars. For testing, we need a way to pass in custom values to test various scenarios. We can use `os.Setenv`, but that makes the tests very difficult to parallelize. This new implementation allows us to provide a custom Lookup function (e.g. from a map), and also supports arbitrary decoder functions (more on that later).

    EDIT: moved to sethvargo/envconfig.

- **Don't overwrite struct values on parsing.** Another major drawback of envconfig is that is doesn't take into account existing struct field values. Creating a struct with values and then passing it to envconfig will overwrite any values in the struct. This is why GH-510 is still happening, and is the reason we have to call `os.Setenv` when passing structs to `setup.Setup`, **even if they've been populated with the right data**. This makes integration testing especially painful - we already have a `database.Config` struct, but we have to map it to envvars in order to correctly setup. This fixes that - only zero-value fields are overwritten.

    EDIT: moved to sethvargo/envconfig.

- **Separate secrets from config.** - There was a really tight coupling between secret resolution and envconfig parsing (it was a dependency input to the Process function afterall). This drops that dependency in favor of a more agnostic `MutatorFunc` that can alter values before they are written into the struct field. This allows us to extend mutations of other types in the future (e.g. `uppercase`, `decrypt`, etc). The secret resolution logic has moved into the secrets package, where it probably belongs in the first place.

- **Closer to serverenv** - Simplify things a bit by having the serverenv hold the Closer via a `Close` method instead of passing around a function.

- Fixes GH-525
- Fixes GH-510


internal/envconfig/envconfig.go
```diff
@@ -12,168 +12,499 @@
...
+	ErrRequiredAndDefault = fmt.Errorf("field cannot be required and have a default value")
+	ErrUnknownOption      = fmt.Errorf("unknown option")
 )
 
-// BaseConfig is the default base configuration.
-type BaseConfig struct {
-	// SecretsDir is the base directory where secrets are stored.
-	SecretsDir string `envconfig:"SECRETS_DIR" default:"/var/run/secrets"`
+// Lookuper is an interface that provides a lookup for a string-based key.
+type Lookuper interface {
```
gurayAlsac, line 110: naming nit (the worst kind of nit): maybe LookerUpper? Or ConfigResolver? Or just ConfigMap, since most programmers associate that with simple key lookup interfaces.

internal/envconfig/envconfig.go
```diff
@@ -12,168 +12,499 @@
...
+//     func resolveSecretFunc(key, value string) (string, error) {
+//         if strings.HasSuffix(key, "_SECRET") {
+//             return secretmanager.Resolve(value)
+//         }
+//     }
+//
+//     var config Config
+//     ProcessWith(&config, new(OsLookuper), resolveSecretFunc)
+//
+// It is conceptually similar to kelseyhightower/envconfig, with the following
```
gurayAlsac, line 65: Is it worth it or possible to import that and add the behavior changes on top, rather than reinvent from scratch? I don't know the landscape well enough; are there other standard config libraries that are worth mentioning (e.g. not using "foo" because xyz)?

internal/export/config.go
```diff
@@ -64,3 +64,96 @@ func (c *Config) KeyManagerConfig() *signing.Config {
 func (c *Config) SecretManagerConfig() *secrets.Config {
 	return &c.SecretManager
 }
+
+// TestConfigDefaults returns a configuration populated with the default values.
```
gurayAlsac, line 68: general comment: should the responsibility of providing test config live in the config_test.go files? Or if they need to live together with the definition of the config, is there a way to reduce some of the boilerplate a bit? Or maybe use conditionals to represent the test scenarios. E.g., 
```
CreateTimeout: Env = test ? 1 * time.Minute : 5 * time.Minute
```

internal/envconfig/envconfig.go
gurayAlsac: Ah, I see below we were actually straightup importing "github.com/kelseyhightower/envconfig" and now we've rolled our own. Maybe its worth a mention of other design options that were considered.

sethvargo: I commented in the PR body about why we can't use it. It doesn't provide the extensibility hooks we need.

internal/export/config.go
sethvargo: These exist for two reasons:

1. A failsafe in case defaults are changed
1. A way to share configs with subpackages (e.g. database)

They can't live in `_test.go` because then we can't share them with other packages. For example, anything that depends on the database has database defaults and database overrides. Instead of duplicating that logic everywhere, I chose to export these on the main package. This _is_ reducing the boilerplate 😄

sethvargo: We've also been bitten by enough configuration issues that I'm okay with a bunch of code honestly.

internal/envconfig/envconfig.go
sethvargo: Happy to change it, but this is an internal package (only we can depend on it).

internal/envconfig/envconfig.go
```diff
@@ -12,168 +12,499 @@
...
+//
+// In the environment, maps are specified as comma-separated key:value pairs:
+//
+//     export MYVAR="a:b,c:d" // map[string]string{"a":"b", "c":"d"}
+//
+// If you need to modify environment variable values before processing, you can
+// specify a custom mutator:
+//
+//     type Config struct {
+//         Password `env:"PASSWORD_SECRET"`
```
mikehelmick, line 53: i don't like that this isn't drop in replacement for the existing secret:// syntax

internal/envconfig/envconfig.go
mikehelmick: I, too, would prefer to see a simpler approach.

we first did that by using kelseyhightower/envconfig to load the environment and then post-processing all the secret values.

I thought that's the direction we were headed back towards (at least that's the proposal in the issue)

internal/federationin/config_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/google/exposure-notifications-server/internal/envconfig"
+	"github.com/google/exposure-notifications-server/internal/federationin"
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestEnvconfigProcess(t *testing.T) {
```
mikehelmick, line 27: this is overkill - no reason to delete it if it's there, but testing for this belong w/ envconfig and each individual use shouldn't need extensive testing.

internal/database/config.go
```diff
@@ -19,23 +19,99 @@ import (
...
 }
 
-func (c *Config) DB() *Config {
+func (c *Config) DatabaseConfig() *Config {
 	return c
 }
+
+// TestConfigDefaults returns a configuration populated with the default values.
+// It should only be used for testing.
+func TestConfigDefaults() *Config {
```
mikehelmick, line 45: looking through - I think all of these tests is overkill and confusing in the main files.

internal/envconfig/envconfig.go
mikehelmick: ah - this is confusing, since we used this example but not how our secrets are specified :)

internal/database/config.go
sethvargo: Removed

internal/federationin/config_test.go
sethvargo: Removed

internal/envconfig/envconfig.go
sethvargo: The key here is that _we_ define how secrets are parsed in the secrets package. That was just an example. I moved it to its own package and I think that makes it less confusing now.

internal/admin/config.go
```diff
@@ -86,3 +86,76 @@ func (c *Config) RenderTemplate(w http.ResponseWriter, tmpl string, p TemplateMa
 
 	return nil
 }
+
+// TestConfigDefaults returns a configuration populated with the default values.
+// It should only be used for testing.
+func TestConfigDefaults() *Config {
```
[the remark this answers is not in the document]
sethvargo: Removed

internal/envconfig/envconfig.go
sethvargo: Per chat, there are challenges with kenvconfig that aren't easily patched or contributed upstream. I've pulled all the core logic into sethvargo/envconfig and cleaned this up.

internal/envconfig/envconfig.go
sethvargo: External package now.

internal/setup/setup.go
```diff
@@ -81,79 +94,96 @@ func Setup(ctx context.Context, config DatabaseConfigProvider) (*serverenv.Serve
...
+		serverEnvOpts = append(serverEnvOpts, blobStorage)
 	}
-	opts = append(opts, serverenv.WithDatabase(db))
 
-	// AuthorizedApp must come after database setup due to the dependency.
-	if typ, ok := config.(AuthorizedAppConfigProvider); ok {
-		logger.Infof("Effective AuthorizedApp config: %+v", typ.AuthorizedAppConfig())
-		provider, err := authorizedapp.NewDatabaseProvider(ctx, db, typ.AuthorizedAppConfig(), authorizedapp.WithSecretManager(sm))
+	// Setup the database connection.
+	if provider, ok := config.(DatabaseConfigProvider); ok {
```
mikehelmick, line 157: one thing we have lost here, and I'm not sure if it is important or not, is that previously being a DatabaseConfigProvider provider was required (compile time interface check). Not, setup will be successful without one.

sethvargo: Do you think that's a blocker? Happy to add it back, but it felt weird to accept a typed interface and then assert other types in the function body.

mikehelmick: let's go with this.

PR #550, opened 2020-06-04
Accept, but embargo currently valid keys

### LINKED ISSUE #543: Evaluate holding current day's key
Currently, a key that is still valid (rolling period + rolling period count covers the current time) causes a whole publish API request to fail.

As this has come up a couple of times, perhaps it is better to hold the still valid key until the end of the UTC day, but allow others to be sent out immediately. 

/assign

mikehelmick:
Instead of rejecting keys that have started before the instant now, but aren't yet expired, accept those keys and emabargo them until they can no longer be used.
This is done by setting the createdAt time on that key.

There is an invariant here that there can only be one key in this state since we don't allow overlapping intervals and we don't allow keys to start in the future.

/fixes #543

### DIFF
--- a/internal/publish/config.go
+++ b/internal/publish/config.go
@@ -43,8 +43,8 @@ type Config struct {
 	TruncateWindow     time.Duration `envconfig:"TRUNCATE_WINDOW" default:"1h"`
 
 	// Flags for local development and testing.
-	DebugAPIResponses   bool `envconfig:"DEBUG_API_RESPONSES"`
-	DebugAllowRestOfDay bool `envconfig:"DEBUG_ALLOW_REST_OF_DAY"`
+	DebugAPIResponses       bool `envconfig:"DEBUG_API_RESPONSES"`
+	DebugReleaseSameDayKeys bool `envconfig:"DEBUG_RELEASE_SAME_DAY_KEYS"`
 }
 
 func (c *Config) AuthorizedAppConfig() *authorizedapp.Config {
--- a/internal/publish/model/exposure_model.go
+++ b/internal/publish/model/exposure_model.go
@@ -95,8 +95,7 @@ func TruncateWindow(t time.Time, d time.Duration) time.Time {
 }
 
 // TimeForIntervalNumber returns the time at which a specific interval starts.
-// This is done by turning the internal number into the corresponding unix timestamp,
-// multiplying by 600 seconds (10 minutes).
+// The interval number * 600 (10m = 600s) is the corresponding unix timestamp.
 func TimeForIntervalNumber(interval int32) time.Time {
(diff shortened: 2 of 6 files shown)


mikehelmick:
/hold 

missed tests in another package

internal/publish/config.go
```diff
@@ -43,8 +43,8 @@ type Config struct {
 	TruncateWindow     time.Duration `envconfig:"TRUNCATE_WINDOW" default:"1h"`
 
 	// Flags for local development and testing.
-	DebugAPIResponses   bool `envconfig:"DEBUG_API_RESPONSES"`
-	DebugAllowRestOfDay bool `envconfig:"DEBUG_ALLOW_REST_OF_DAY"`
```
sethvargo, line 47: We'll need to update the Terraform configurations for the server instances with this and coordinate a deploy.

internal/publish/model/exposure_model.go
```diff
@@ -92,37 +94,53 @@ func TruncateWindow(t time.Time, d time.Duration) time.Time {
 	return t.Truncate(d)
 }
 
+// TimeForIntervalNumber returns the time at which a specific interval starts.
+// This is done by turning the internal number into the corresponding unix timestamp,
+// multiplying by 600 seconds (10 minutes).
```
sethvargo, line 99: Is this comment correct? I don't see the multiplication by 600s. It's kinda unclear to me what this is and why 10min.

internal/publish/model/exposure_model.go
```diff
@@ -92,37 +94,53 @@ func TruncateWindow(t time.Time, d time.Duration) time.Time {
...
+	return time.Unix(int64(verifyapi.IntervalLength.Seconds())*int64(interval), 0)
+}
+
 // Transformer represents a configured Publish -> Exposure[] transformer.
 type Transformer struct {
 	maxExposureKeys     int
 	maxIntervalStartAge time.Duration // How many intervals old does this server accept?
 	truncateWindow      time.Duration
-	debugAllowRestOfDay bool // raises end time of keys to the end of day, but doesn't embargo. For e2e testing only.
+	debugRelesaeSameDay bool // If true, still valid keys are not embarged.
```
sethvargo, line 109: relesae, embarged

sethvargo: (Rename field everywhere)

internal/publish/config.go
mikehelmick: this shouldn't be in the terraform - as we want the default value of false.

I set the old value manually in our UAT environment.

internal/publish/model/exposure_model.go
mikehelmick: the comment was poorly written.

PR #551, opened 2020-06-04
Gracefully stop servers on interrupt

sethvargo:


cmd/cleanup-export/main.go
```diff
@@ -17,35 +17,64 @@ package main
...
+		return fmt.Errorf("cleanup.NewExportHandler: %w", err)
 	}
+
 	mux := http.NewServeMux()
 	mux.Handle("/", handler)
-	logger.Infof("starting export cleanup server on :%s", config.Port)
-	instrumentedHandler := &ochttp.Handler{Handler: mux}
-	log.Fatal(http.ListenAndServe(":"+config.Port, instrumentedHandler))
+
+	server := server.New(config.Port, mux)
```
sethvargo, line 60: I'm open to unifying `New` and `Start` into a single API, since there's no _real_ value in `New` since all the fields are private. 

I also thought about an API where the context provided to `server.Start` is used to stop it:

```go
ctx, done := context.WithCancel(ctx)
if err := server.Start(ctx); err != nil {
  return err
}

<-ctx
done()
```

but it felt too magical, and it's much harder to return any errors that occur while stopping back to the main thread.

sethvargo: I also thought about a `server.Wait(ctx)`, which made a bit more sense. I'm open to other designs here that reduce duplication.

mikehelmick: seems like everything here on down is "common main" should we pass the mux to a "run server" method

adg: usually it's `server.Serve(ctx)` that blocks until it shuts down gracefully. that way you don't need a stop method.

I think separate constructor is fine
lets you separate serve errors from init errors

sethvargo: @adg thanks for the feedback. The problem we're trying to solve is intercepting a signal (like SIGTERM/SIGINT) and having that close the server. `interrupt.Context()` creates a context that closes on various signals, but I'm seeking advice on the best way to "plumb" that through to the servers.

adg: Right, I can see that. I'm saying have a single Serve method that takes a ctx and have it block until that context is done, then it does the graceful shutdown (using a background context, with timeout, for any shutdown-related things that require a context), then returns. Then the last line of your main function can be `return server.Serve(ctx)`

sethvargo: @adg @mikehelmick updated. PTAL.

PR #553, opened 2020-06-05
Fix reference to verification_protocol.md

nmiodice:
This change fixes a reference to the `verification_protocol.md` file. The current reference is a broken link.


nmiodice:
@googlebot I signed it!

Issue #555, opened 2020-06-05
Implement caching for health authority public keys


mikehelmick:
/assign
/kind enhancement

sethvargo:
We have enough cachable things (authorized apps, secrets, now pkeys) that we should probably unify into a single `cacher` internal package.

mikehelmick:
yep - I'm taking that into account.

mikehelmick:
fixed

/close

PR #557, opened 2020-06-05
Factor out cache implementation to standalone package.

mikehelmick:
* Create write thru cache implementation
* Move database provider to it
* First part of #555

Follow on PR will bring this to other parts of the codebase that use caching.

### DIFF
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -1,0 +1,138 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package cache implements an inmemory cache for any interface{} object.
+package cache
+
+import (
+	"errors"
+	"sync"
+	"time"
+)
+
+var ErrInvalidDuration = errors.New("expireAfter duration cannot be negative")
+
+type WriteThru func() (interface{}, error)
+
+type Cache struct {
+	data map[string]item
+	mu   sync.RWMutex
+}
+
+type item struct {
+	object    interface{}
+	expiresAt int64
+}
+
+func (i *item) expired() bool {
+	return i.expiresAt < time.Now().UnixNano()
+}
+
+// New creates a new in memory cache.
+func New() *Cache {
+	return &Cache{
+		data: make(map[string]item),
+	}
+}
+
+func (c *Cache) purgeExpired(name string, exp int64) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if item, ok := c.data[name]; ok && item.expiresAt == exp {
+		// found, and the expiry time is still the same as when the purge was requested.
+		delete(c.data, name)
+	}
+}
+
+// Size returns the number of items in the cache.
+func (c *Cache) Size() int {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return len(c.data)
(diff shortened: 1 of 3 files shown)


internal/cache/cache.go
```diff
@@ -0,0 +1,138 @@
...
+	return &Cache{
+		data: make(map[string]item),
+	}
+}
+
+func (c *Cache) purgeExpired(name string, exp int64) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if item, ok := c.data[name]; ok && item.expiresAt == exp {
```
sethvargo, line 53: This shouldn't be ==, right? item.expired()?

internal/cache/cache.go
```diff
@@ -0,0 +1,138 @@
...
+	val, hit := c.Lookup(name)
+	if hit {
+		return val, nil
+	}
+
+	// Escalate the lock to a RW lock.
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	// double check that the value hasn't been set by another goroutine.
+	if val, hit := c.data[name]; hit && !val.expired() {
```
sethvargo, line 81: It might be helpful to separate these apart for readability.

internal/cache/cache.go
```diff
@@ -0,0 +1,138 @@
...
+
+}
+
+// Lookup checks the cache for a non-expired object by the supplied key name.
+// The bool return informs the caller if there was a cache hit or not.
+// A return of nil, true means that nil is in the cache.
+// Where nil, false indicates a cache miss or that the value is expired and should
+// be refreshed.
+func (c *Cache) Lookup(name string) (interface{}, bool) {
+	c.mu.RLock()
```
sethvargo, line 107: If we take out a full lock here, we can delete from the map without spawning the background goroutine.

internal/cache/cache.go
sethvargo: Oh I see after reading the code below. The behavior and purpose of this is unclear. I think it should either accept a name and item or be moved inline.

internal/cache/cache.go
```diff
@@ -0,0 +1,138 @@
...
+	object    interface{}
+	expiresAt int64
+}
+
+func (i *item) expired() bool {
+	return i.expiresAt < time.Now().UnixNano()
+}
+
+// New creates a new in memory cache.
+func New() *Cache {
```
sethvargo, line 43: Expiration feels like it should be a property of the cache, not a property of each individual cache item. WDYT?

internal/cache/cache.go
```diff
@@ -0,0 +1,138 @@
...
+
+import (
+	"errors"
+	"sync"
+	"time"
+)
+
+var ErrInvalidDuration = errors.New("expireAfter duration cannot be negative")
+
+type WriteThru func() (interface{}, error)
```
sethvargo, line 26: Can we just name this Func (`cache.Func`)?

internal/cache/cache_test.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	if err.Error() != "nope" {
+		t.Errorf("incorrect error, want: `nope` got: %v", err.Error())
+	}
+	if got != nil {
+		t.Errorf("unexpecetd cached item, want: nil, got %v", got)
+	}
+}
+
+func TestInvalidDuration(t *testing.T) {
```
sethvargo, line 121: Can you add a test for concurrency using two goroutines please?

internal/cache/cache.go
mikehelmick: for our purposes, duration per instance is sufficient.

internal/cache/cache.go
mikehelmick: added additional comments + changed name of exp

internal/cache/cache.go
mikehelmick: tried a few ways - i like this the best.

internal/cache/cache.go
mikehelmick: for a read heavy cache - I think's it's best to use the RLock that can be shared.

internal/cache/cache_test.go
mikehelmick: how about 10?

PR #558, opened 2020-06-05
fix(signing): add configuration option for azure key vault

ultimateboy:
The code to support signing with Azure Key Vault was added in #527, but it was not added as a startup configuration option, this simply fixes that.

### DIFF
--- a/internal/signing/config.go
+++ b/internal/signing/config.go
@@ -19,6 +19,7 @@ type KeyManagerType string
 
 const (
 	KeyManagerTypeAWSKMS         KeyManagerType = "AWS_KMS"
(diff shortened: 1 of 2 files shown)


sethvargo:
Hey @ultimateboy thanks for the PR! Let us know if you have any questions about the CLA signing bits.

ultimateboy:
@sethvargo Will do. Might take a little bit of time to figure out those details. But I'm on it :)

sethvargo:
Got it. If other folks plan to contribute, they can start the process early at https://cla.developers.google.com/ before opening a PR. Good news is that it's a one-time process.

ultimateboy:
@googlebot I signed it!

sethvargo:
Thanks @ultimateboy!

PR #559, opened 2020-06-05
Minor changes to support e2e testing

sethvargo:


internal/authorizedapp/memory_provider.go
```diff
@@ -27,26 +29,44 @@ var _ Provider = (*MemoryProvider)(nil)
 // MemoryProvider is an Provider that stores values in-memory. It is primarily
 // used for testing.
 type MemoryProvider struct {
-	Data map[string]*model.AuthorizedApp
+	lock sync.RWMutex
+	data map[string]*model.AuthorizedApp
```
mikehelmick, line 33: this is just for tests right, why not have tests use the test DB infrastructure?

internal/authorizedapp/memory_provider.go
```diff
@@ -27,26 +29,44 @@ var _ Provider = (*MemoryProvider)(nil)
...
 }
 
 // NewMemoryProvider creates a new Provider that reads from a database.
```
mikehelmick, line 36: comment out of date

internal/authorizedapp/memory_provider.go
sethvargo: This is faster and cheaper.

PR #560, opened 2020-06-05
Move to single cache impl.

mikehelmick:
* Move secret manager cache to internal/cache write through impl
* Start caching HealthAuthoirty configuration in verificaiton code

/fixes 555

### DIFF
--- a/internal/verification/config.go
+++ b/internal/verification/config.go
@@ -19,5 +19,5 @@ import "time"
 // Config represents the available configuration for the public health authority
 // verification piece.
 type Config struct {
-	CacheDuration time.Duration `env:"HEALTH_AUTHORITY_CACHE_DURATION, default=5m"`
+	CacheDuration time.Duration `env:"VERIFICATION_CACHE_DURATION, default=5m"`
 }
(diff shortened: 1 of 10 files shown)


internal/verification/config.go
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+package verification
+
+import "time"
+
+// Config represents the available configuration for the public health authority
+// verification piece.
+type Config struct {
+	CacheDuration time.Duration `env:"HEALTH_AUTHORITY_CACHE_DURATION, default=5m"`
```
sethvargo, line 22: Should this be `VERIFICATION_CACHE_DURATION`?

mikehelmick: done

sethvargo:
@mikehelmick seems like a legit failure.

mikehelmick:
> @mikehelmick seems like a legit failure.

yeah, looking.

mikehelmick:
pushed fix - was indeed a legit bug

race condition that isn't always triggered by the test

mikehelmick:
added test that always cause the race condition to be hit ⏲️

PR #561, opened 2020-06-05
Add initial e2e framework

sethvargo:
This is a WIP (but mergeable) that does a full e2e test of setup, publish, export.

TODO: verify export, verify signature, create pha signature and verify.

### DIFF
--- a/internal/integration/integration_test.go
+++ b/internal/integration/integration_test.go
@@ -1,0 +1,195 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


internal/integration/integration_test.go
```diff
@@ -0,0 +1,195 @@
...
+		TruncateWindow:          1 * time.Hour,
+		DebugAPIResponses:       true,
+		DebugReleaseSameDayKeys: true,
+	}
+
+	publishHandler, err := publish.NewHandler(ctx, publishConfig, env)
+	if err != nil {
+		tb.Fatal(err)
+	}
+	mux.HandleFunc("/publish", handlers.WithMinimumLatency(publishConfig.MinRequestDuration, publishHandler))
```
mikehelmick, line 144: we don't need the minimum latency wrapper here, but doesn't hurt.

mikehelmick:
/lgtm

good start.

PR #563, opened 2020-06-05
Add Navigation, theming, for docs

crwilcox:
Separating out some work to add navigation to our doc site.

this allows locally hosting with `bundle exec jekyll serve -i`


crwilcox:
For reference 
image

PR #564, opened 2020-06-06
Iam auth

frodopwns:
improve the auth so its just for SP in public cloud for now


internal/signing/azure_keyvault.go
```diff
@@ -227,3 +243,45 @@ func convert1363ToAsn1(b []byte) ([]byte, error) {
 
 	return asn1.Marshal(rs)
 }
+
+func GetKeyvaultAuthorizer() (autorest.Authorizer, error) {
```
sethvargo, line 247: From reading the SDK, I was under the impression that this was handled automatically by NewAuthorizerFromEnvironment and you can set environment variables.

PR #570, opened 2020-06-08
Switch to git-sha based deployments

sethvargo:
This will make it easier to diff between deployments


sherifkozman:
@sethvargo scripts have been switched to git-sha based deployments , yet all documentation are still referencing timestamps based tags for container images (i.e YYYYMMDDhhmmss)

sethvargo:
GH-576

PR #571, opened 2020-06-08
Update envconfig to fix nested struct passing mutator funcs

sethvargo:


internal/database/connection.go
```diff
@@ -67,14 +66,6 @@ func dbConnectionString(config *Config) string {
 	return strings.Join(p, " ")
 }
 
-// dbURI builds a Postgres URI suitable for the lib/pq driver, which is used by
```
sethvargo, line 70: This isn't used anywhere in the codebase, and I lost a few minutes to it before realizing it's not used. Happy to revert this change, but I don't like dead code 😄

sethvargo:
Hmm looks like the container failed to start.

/retest

PR #574, opened 2020-06-08
Don't log passwords

sethvargo:


sethvargo:
/assign @mikehelmick 

This still isn't "perfect" as the password would be printed if someone called `Printf` with "%v" or "%#v", but it's a good start.

internal/setup/setup.go
```diff
@@ -120,7 +120,7 @@ func SetupWith(ctx context.Context, config interface{}, l envconfig.Lookuper) (*
 		// Update serverEnv setup.
 		serverEnvOpts = append(serverEnvOpts, serverenv.WithSecretManager(sm))
 
-		logger.Infof("secret manager configuration: %+v", smConfig)
+		logger.Infow("secret manager", "config", smConfig)
```
sethvargo, line 123: This gives us structured logging, yay

crwilcox:
lgtm, but will let @mikehelmick approve :)

internal/database/config.go
```diff
@@ -25,7 +25,7 @@ type Config struct {
 	Port               string        `env:"DB_PORT, default=5432"`
 	SSLMode            string        `env:"DB_SSLMODE, default=require"`
 	ConnectionTimeout  int           `env:"DB_CONNECT_TIMEOUT"`
-	Password           string        `env:"DB_PASSWORD"`
+	Password           string        `env:"DB_PASSWORD" json:"-"`
```
mikehelmick, line 28: so, the json:"-" is interpreted by the logging infrastructure?

mikehelmick: it's fine if so - but would deserving of a comment and maybe a link

sethvargo: Added comment.

sethvargo:
@mikehelmick re-review plz

PR #575, opened 2020-06-08
Remove unused db struct tags

sethvargo:
General cleanup, these aren't used

### DIFF
--- a/internal/export/model/export_model.go
+++ b/internal/export/model/export_model.go
@@ -32,15 +32,15 @@ const (
 )
 
 type ExportConfig struct {
-	ConfigID         int64         `db:"config_id"`
-	BucketName       string        `db:"bucket_name"`
-	FilenameRoot     string        `db:"filename_root"`
-	Period           time.Duration `db:"period_seconds"`
-	OutputRegion     string        `db:"output_region"`
-	InputRegions     []string      `db:"input_regions"`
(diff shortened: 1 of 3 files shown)


sherifkozman:
upgrading 1 week+ old commits break exisiting Cloud run deployments because deployment scripts do not take into account changing variables names in Cloud Run such as `dbPassword` to `db-password`

mikehelmick:
> upgrading 1 week+ old commits break exisiting Cloud run deployments because deployment scripts do not take into account changing variables names in Cloud Run such as `dbPassword` to `db-password` 

Is there a concern with the naming in this change?

Or there a problem somewhere else?

sherifkozman:
The problem occurs when a project proviosned with older terraform configs , the scripts dont reflects those changes and thus services fail to deploy. So the deployment isnt backward compatible and may require removing all services and deploying them again.

sethvargo:
Hi @sherifkozman. The project is iterating rapidly and we have made some non-BC changes in both the code and Terraform configurations. If you have specific issues, I'd be happy to help you resolve them.

sherifkozman:
@sethvargo Thanks, i have implemented the required fixes to our exisiting implementation but i was noting some breaking changes that would affect anyone who have previously deployed it and it might be noted somewhere if it wont be handled. 
Other than that all good

PR #576, opened 2020-06-09
Update deploy docs to note git-sha

sethvargo:


sethvargo:
Grumble. I fixed the data race in GH-577

PR #579, opened 2020-06-09
Adding missing cloudSQL client IAM Roles for Cloud Run Services

sherifkozman:
Adding the previously removed CloudSQL client IAM permissions for Cloud RUN Service accounts

### DIFF
--- a/terraform/service_cleanup_export.tf
+++ b/terraform/service_cleanup_export.tf
@@ -22,6 +22,12 @@ resource "google_service_account" "cleanup-export" {
   display_name = "Exposure Notification Cleanup Export"
 }
 
+resource "google_project_iam_member" "cleanup-export-cloudsql" {
+  project = data.google_project.project.project_id
+  role    = "roles/cloudsql.client"
+  member  = "serviceAccount:${google_service_account.export.email}"
+}
+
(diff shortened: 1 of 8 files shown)


sherifkozman:
@googlebot I signed it!

sethvargo:
Hi @sherifkozman 

The Cloud Run services should _not_ need this permission anymore. They communicate with Cloud SQL over a local VPN, and thus the Cloud SQL proxy (which is what requires the IAM permission) is no longer required for this setup.

sherifkozman:
@sethvargo Terraform isnt able to communicate this change for Cloud RUN service configs for anyone who has deployed services a week ago so it would break any exisiting deployment that utilized the service account provisioned prior enabling VPC access.
It needs to be noted or phased out gradually.

sethvargo:
Hi @sherifkozman - this is already documented [here in the deployment documentation](https://github.com/google/exposure-notifications-server/blob/master/docs/deploying.md#provisioning-infrastructure-with-terraform):

> You can use Terraform to provision the initial infrastructure, database, service accounts, and first deployment of the services on Cloud Run. **Terraform does not manage the Cloud Run services after their initial creation!**

Terraform ignores all changes to the Cloud Run services.

PR #584, opened 2020-06-09
Move test steps into make targets

sethvargo:
This makes the presubmit script a bit "dumber" and moves some of the logic into a shared Makefile, giving more flexibility for running locally. It also creates two targets for testing: `test` and `test-acc`. The first target runs the short tests. The second target runs all the integration tests and the race detector. CI uses the `test-acc`.

### DIFF
--- a/internal/cache/cache_test.go
+++ b/internal/cache/cache_test.go
@@ -153,29 +153,33 @@ func TestConcurrentReaders(t *testing.T) {
 	}
 
 	parallel := 10
-	done := make(chan bool, parallel)
+	done := make(chan error, parallel)
 	for i := 0; i < parallel; i++ {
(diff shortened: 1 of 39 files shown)


internal/cache/cache_test.go
```diff
@@ -153,29 +153,33 @@ func TestConcurrentReaders(t *testing.T) {
 	}
 
 	parallel := 10
-	done := make(chan bool, parallel)
+	done := make(chan error, parallel)
```
mikehelmick, line 156: good change

mikehelmick:
some test failures

sethvargo:
Looks like shell issues on the builder. Gonna have to dig in more.

PR #604, opened 2020-06-10
include end time in export filenames

gurayAlsac:
If we configure a "catch up file" at some factor as a more frequent
export configuration (e.g., 24 hr catch up, 4 hr regular period),
unless they use a different prefix/suffix in the name they'll clash
on naming. Easy way to make this foolproof is to have filenames
include start/end times.

This should not effect anybody that simply reads index for files to
download. This also shouldn't effect our sample app implementation
because its region is not configured for catch up files. But I will
also submit a corresponding change in the regex it uses to group
parts of a batch together:
https://github.com/google/exposure-notifications-android/blob/5eb8e99dc8ef0db4ee4754495d032e27094bcfe9/app/src/main/java/com/google/android/apps/exposurenotification/network/Uris.java#L66

### DIFF
--- a/internal/export/worker.go
+++ b/internal/export/worker.go
@@ -287,7 +287,7 @@ func (s *Server) createIndex(ctx context.Context, eb *model.ExportBatch, newObje
 }
 
 func exportFilename(eb *model.ExportBatch, batchNum int) string {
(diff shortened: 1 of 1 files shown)


gurayAlsac:
Sample app repo has picked up the change to support an optional file name:
https://github.com/google/exposure-notifications-android/blob/765dab0af327d92cc6f5b7f16f5ad80d29361370/app/src/main/java/com/google/android/apps/exposurenotification/network/Uris.java#L66

Proceeding with this change now.

/unhold

Issue #606, opened 2020-06-10
offer /debug endpoint surfacing some server config/state


gurayAlsac:
To facilitate end-to-end testing (where testers may not have access to the underlying project resources), we could offer some super simple `/debug` endpoints handled in, say, the publish and export servers. These would be only served in test instances, and could serve up a simple static page that surfaces:

- all environment variables
- dump all config tables (e.g., export config, authorized app, signature info)
- last n keys uploaded + timestamps (to check if server got a key from a test upload)
- state of export batcher/worker (last few batches/files created, # of keys, truncate window, etc)

Again, this would only be enabled in test, and would be used to get a quick snapshot of server state to pinpoint or rule out issues in testing with devices.

We could strip some PII and make this more of an operational /statusz type of page, but I think the greater need is in debugging E2E configuration/state, where seeing some of the keys uploaded could be helpful.

gurayAlsac:
Seth helped close this one out. We can reopen new issue if needs change.

PR #609, opened 2020-06-10
Add debug endpoints on export

sethvargo:
This PR is for discussion. It introduces a `/debug` endpoint that is disabled by default and can be enabled to provide unauthenticated debug information, useful for client or e2e testing. We can debate the information returned (and always change it later), but the larger question is **should we have this**?  

Git is being weird, so you might want to review the whitespace-ignored diff: https://github.com/google/exposure-notifications-server/pull/609/files?w=1

Refs GH-606.

@mikehelmick @gurayAlsac @peetsasaki

### DIFF
--- a/internal/export/batcher.go
+++ b/internal/export/batcher.go
@@ -22,84 +22,89 @@ import (
 	"time"
 
 	coredb "github.com/google/exposure-notifications-server/internal/database"
-	"github.com/google/exposure-notifications-server/internal/export/database"
+	exportdatabase "github.com/google/exposure-notifications-server/internal/export/database"
 	"github.com/google/exposure-notifications-server/internal/export/model"
 	publishmodel "github.com/google/exposure-notifications-server/internal/publish/model"
 
 	"github.com/google/exposure-notifications-server/internal/logging"
 )
 
-// CreateBatchesHandler is a handler to iterate the rows of ExportConfig and
+// handleCreateBatches is a handler to iterate the rows of ExportConfig and
 // create entries in ExportBatchJob as appropriate.
-func (s *Server) CreateBatchesHandler(w http.ResponseWriter, r *http.Request) {
-	ctx, cancel := context.WithTimeout(r.Context(), s.config.CreateTimeout)
-	defer cancel()
+func (s *Server) handleCreateBatches(ctx context.Context) http.HandlerFunc {
 	logger := logging.FromContext(ctx)
 	metrics := s.env.MetricsExporter(ctx)
+	db := s.env.Database()
 
-	// Obtain lock to make sure there are no other processes working to create batches.
-	lock := "create_batches"
-	unlockFn, err := s.db.Lock(ctx, lock, s.config.CreateTimeout)
-	if err != nil {
-		if errors.Is(err, coredb.ErrAlreadyLocked) {
-			metrics.WriteInt("export-batcher-lock-contention", true, 1)
-			msg := fmt.Sprintf("Lock %s already in use, no work will be performed", lock)
-			logger.Infof(msg)
-			fmt.Fprint(w, msg) // We return status 200 here so that Cloud Scheduler does not retry.
+	return func(w http.ResponseWriter, r *http.Request) {
+		ctx, cancel := context.WithTimeout(r.Context(), s.config.CreateTimeout)
+		defer cancel()
+
+		// Obtain lock to make sure there are no other processes working to create batches.
+		lock := "create_batches"
+		unlockFn, err := db.Lock(ctx, lock, s.config.CreateTimeout)
+		if err != nil {
+			if errors.Is(err, coredb.ErrAlreadyLocked) {
+				metrics.WriteInt("export-batcher-lock-contention", true, 1)
+				msg := fmt.Sprintf("Lock %s already in use, no work will be performed", lock)
+				logger.Infof(msg)
+				fmt.Fprint(w, msg) // We return status 200 here so that Cloud Scheduler does not retry.
+				return
+			}
+			logger.Errorf("Could not acquire lock %s: %v", lock, err)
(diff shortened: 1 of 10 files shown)


internal/export/worker.go
```diff
@@ -43,55 +43,60 @@ const (
...
 )
 
-// WorkerHandler is a handler to iterate the rows of ExportBatch, and creates
+// handleDoWork is a handler to iterate the rows of ExportBatch, and creates
 // export files.
-func (s *Server) WorkerHandler(w http.ResponseWriter, r *http.Request) {
-	ctx, cancel := context.WithTimeout(r.Context(), s.config.WorkerTimeout)
-	defer cancel()
+func (s *Server) handleDoWork(ctx context.Context) http.HandlerFunc {
 	logger := logging.FromContext(ctx)
```
sethvargo, line 51: This is actually a current bug in all our handlers. We pull the logger from the request context, but it should really come from the server context.

sethvargo:
I didn't create many configs to see, but here's an example of what the endpoint returns:

```json
{
  "Config": {
    "Database": {
      "Name": "en-server-db",
      "User": "en-server",
      "Host": "127.0.0.1",
      "Port": "5432",
      "SSLMode": "require"
    },
    "KeyManager": {
      "KeyManagerType": "GOOGLE_CLOUD_KMS"
    },
    "SecretManager": {
      "SecretManagerType": "GOOGLE_SECRET_MANAGER",
      "SecretsDir": "/var/run/secrets",
      "SecretCacheTTL": 300000000000
    },
    "Storage": {
      "BlobstoreType": "GOOGLE_CLOUD_STORAGE"
    },
    "Port": "8080",
    "CreateTimeout": 300000000000,
    "WorkerTimeout": 300000000000,
    "MinRecords": 1000,
    "PaddingRange": 100,
    "MaxRecords": 30000,
    "TruncateWindow": 3600000000000,
    "MinWindowAge": 7200000000000,
    "TTL": 1209600000000000,
    "DebugEndpoint": true
  },
  "ExportConfigs": [
    {
      "ConfigID": 1,
      "BucketName": "foobar",
      "FilenameRoot": "foobar",
      "Period": 1000000000,
      "OutputRegion": "US",
      "InputRegions": null,
      "From": "2020-06-08T20:00:00-04:00",
      "Thru": "0001-01-01T00:00:00Z",
      "SignatureInfoIDs": null
    }
  ],
  "ExportBatchEnds": {
    "1": "0001-01-01T00:00:00Z"
  },
  "ExportBatchFiles": null,
  "SignatureInfos": null
}
```

internal/export/debug.go
```diff
@@ -0,0 +1,97 @@
...
+			fmt.Fprint(w, http.StatusText(http.StatusInternalServerError))
+			return
+		}
+
+		resp := &response{
+			Config:           s.config,
+			ExportConfigs:    exportConfigs,
+			ExportBatchEnds:  exportBatchEnds,
+			ExportBatchFiles: exportBatchFiles,
+			SignatureInfos:   signatureInfos,
```
gurayAlsac, line 86: We could also include the last X uploaded TEKs

gurayAlsac, line 86: Were we planning a similar debug handler on the publish server? We could dump authorized apps and some of the config it has too.

internal/export/batcher.go
```diff
@@ -22,84 +22,89 @@ import (
...
 )
 
-// CreateBatchesHandler is a handler to iterate the rows of ExportConfig and
+// handleCreateBatches is a handler to iterate the rows of ExportConfig and
 // create entries in ExportBatchJob as appropriate.
-func (s *Server) CreateBatchesHandler(w http.ResponseWriter, r *http.Request) {
-	ctx, cancel := context.WithTimeout(r.Context(), s.config.CreateTimeout)
-	defer cancel()
+func (s *Server) handleCreateBatches(ctx context.Context) http.HandlerFunc {
 	logger := logging.FromContext(ctx)
```
mikehelmick, line 35: this is technically a change in where the logger comes from, but materially it won't matter since we haven't configured different loggers by context ever.

internal/export/config.go
```diff
@@ -48,6 +48,12 @@ type Config struct {
 	TruncateWindow time.Duration `env:"TRUNCATE_WINDOW, default=1h"`
 	MinWindowAge   time.Duration `env:"MIN_WINDOW_AGE, default=2h"`
 	TTL            time.Duration `env:"CLEANUP_TTL, default=336h"`
+
+	// Debugging flags follow. These should not be enabled in production
+	// environments.
+
+	// DebugEndpointEnabled controls whether the debug endpoint is enabled.
+	DebugEndpoint bool `env:"DEBUG_ENDPOINT"`
```
mikehelmick, line 56: so - I think we want the debug endpoint on the exposure service not the export service.

in our current config, the export service is not accessible without IAM

internal/export/debug.go
mikehelmick: I think we should only put on e on the exposure/publish server

internal/export/config.go
gurayAlsac: We may want to verify that both publish and export are using same truncate window. But if getting a /debug on publish/exposure service is lower hanging fruit that's totally fine.

sethvargo: My plan was to do this on all services, just starting with one for now

sethvargo: Looking more... `export` is guarded by auth already, so you'd need to have the right IAM permissions to invoke the /debug endpoint. `publish` is the only public-facing service where this makes sense, but the publisher won't have access to any of the export runtime configuration (unless we do something whacky like call the Cloud Run API, which seems like a bad ideA).

sethvargo:
Gonna abandon this in favor of a new strategy.

PR #612, opened 2020-06-11
point to new link for file format

gurayAlsac:


docs/server_functional_requirements.md
```diff
@@ -162,7 +162,7 @@ generate the files in a single operation rather than on-demand, and distribute
 the files using a CDN.
 
 For information on the format of the batch file, see
-[Exposure Key Export File Format and Verification](https://www.google.com/covid19/exposurenotifications/pdfs/Exposure-Key-File-Format-and-Verification.pdf).
+[Exposure Key Export File Format and Verification](https://developers.google.com/android/exposure-notifications/exposure-key-file-format).
```
mikehelmick, line 165: Can we also link to the proto definition in this repo?

gurayAlsac: How about linking to slightly friendlier export readme first, and then having that link to proto?

PR #614, opened 2020-06-11
Reduce function arity

mikehelmick:
/fixes 312

### DIFF
--- a/internal/federationin/federationin.go
+++ b/internal/federationin/federationin.go
@@ -189,13 +189,17 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	timeoutContext, cancel := context.WithTimeout(ctx, h.config.Timeout)
 	defer cancel()
 
-	deps := pullDependencies{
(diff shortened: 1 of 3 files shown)


internal/federationin/federationin.go
```diff
@@ -189,13 +189,17 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	timeoutContext, cancel := context.WithTimeout(ctx, h.config.Timeout)
 	defer cancel()
 
-	deps := pullDependencies{
-		fetch:               client.Fetch,
-		insertExposures:     h.publishdb.InsertExposures,
-		startFederationSync: h.db.StartFederationInSync,
+	opts := pullOptions{
```
sethvargo, line 192: Would it be more efficient if these were all pointers?

mikehelmick: I think this is OK as is.

the pullDependencies are function pointers

the query is already a pointer

and the whole pullOptions is passed to functions by reference, not by value..

PR #618, opened 2020-06-11
Avoid dividing by 0

sethvargo:
This is unlikely to happen in production, but in testing with a low export period, it can panic because anything%0 is a division by zero error.

Refs: https://github.com/google/exposure-notifications-server/pull/599/files#r438477087

### DIFF
--- a/internal/export/model/export_model.go
+++ b/internal/export/model/export_model.go
@@ -58,8 +58,8 @@ func (ec *ExportConfig) Validate() error {
 	if ec.Period > oneDay {
 		return errors.New("maximum period is 24h")
 	}
-	if ec.Period == 0 {
-		return errors.New("period must be non-zero")
+	if ec.Period < 1*time.Second {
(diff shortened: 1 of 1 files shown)


internal/export/model/export_model.go
```diff
@@ -58,8 +58,8 @@ func (ec *ExportConfig) Validate() error {
 	if ec.Period > oneDay {
 		return errors.New("maximum period is 24h")
 	}
-	if ec.Period == 0 {
-		return errors.New("period must be non-zero")
+	if ec.Period == 0 || ec.Period.Seconds() == 0 {
```
mikehelmick, line 61: if !(ec.Period > 0) ?

sethvargo: The period could be 5 nanoseconds. That's > 0, but not > 1 second.

mikehelmick: if !(ec.Period.Seconds() >= 1)

PR #621, opened 2020-06-12
Export handlers cleanup

sethvargo:
I pulled the non-debug changes out of GH-609 since I'm not sure we agree that the /debug endpoint should exist on export. However, these changes are still worth including (especially the database check and logger fixes).


sethvargo:
Once again, https://github.com/google/exposure-notifications-server/pull/621/files?w=1 is your friend.

Issue #247, opened 2020-05-14
e2e: create an automated e2e test


crwilcox:
We have done some manual runs, but it would be great to have a test that can do an automated e2e walkthrough of exposure keys getting uploaded and exposure keys getting downloaded

radisef347:
![alt text](https://cdni.pornpics.com/1280/7/51/81056079/81056079_141_f080.jpg)

johanbrandhorst:
@thatlkat Is this document publically available somewhere? I'd be interested in experimenting with this.

thatlkat:
Not yet, will add readme when we have SUT in place

nmiodice:
I'd also love to get my eyes on that document. If one were to implement such an end to end test, would a PR for it be welcome in the repo?

We are planning some work, and I want to better understand how we should prioritize efforts. Is this already being worked on in parallel? If so, when can we expect a true end to end test to be implemented -- one that is fully isolated from the deployed environment?

johanbrandhorst:
We should reopen this, right? Is this no longer relevant?

sethvargo:
I believe we have an e2e test now, @crwilcox @chaodaiG can you confirm?

chaodaiG:
We have integration tests against in memory DB, testing the following scenarios:
- Publish/export keys
- Cleanup published keys
- Cleanup exported keys

https://github.com/google/exposure-notifications-server/blob/main/internal/integration/publish_test.go

So I think we have the requested "upload" part done, but no test for "download" part yet

PR #625, opened 2020-06-12
Add debugger

sethvargo:
Add a new GCP-specific service, debugger, to assist client<->server debugging. This isn't the complete list from GH-606, but it's a start.

/assign @mikehelmick @gurayAlsac

### DIFF
--- a/internal/export/database/export.go
+++ b/internal/export/database/export.go
@@ -119,7 +119,8 @@ func (db *ExportDB) GetAllExportConfigs(ctx context.Context) ([]*model.ExportCon
 		SELECT
 			config_id, bucket_name, filename_root, period_seconds, output_region, from_timestamp, thru_timestamp, signature_info_ids, input_regions
 		FROM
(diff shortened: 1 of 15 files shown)


internal/export/database/export.go
```diff
@@ -358,13 +359,11 @@ func (db *ExportDB) LatestExportBatchEnd(ctx context.Context, ec *model.ExportCo
 
 	row := conn.QueryRow(ctx, `
 		SELECT
-			end_timestamp
+			MAX(end_timestamp)
```
sethvargo, line 362: This is roughly a 14x speed improvement over sorting + limit 1.

sethvargo:
/retest

failure seems unrelated

sethvargo:
@mikehelmick I commented out the flakey test (it was supposed to be a stub for GH-247 anyway) since, and should be a non-issue once GH-599 is in). Can you re-LGTM please?

PR #628, opened 2020-06-15
move cache and base64utils to pkg so they can be imported into verifi…

mikehelmick:
…cation server codebase


internal/signing/azure_keyvault.go
```diff
@@ -26,9 +26,10 @@ import (
 	"math/big"
 	"strings"
 
-	"github.com/Azure/azure-sdk-for-go/services/keyvault/v7.0/keyvault"
 	"github.com/google/exposure-notifications-server/internal/azurekeyvault"
-	"github.com/google/exposure-notifications-server/internal/base64util"
+	"github.com/google/exposure-notifications-server/pkg/base64util"
+
+	"github.com/Azure/azure-sdk-for-go/services/keyvault/v7.0/keyvault"
```
sethvargo, line 32: This seems to have gotten weird

Issue #632, opened 2020-06-15
Add documentation to help estimate deployment cost


crwilcox:
I have had questions about the cost to run this reference server. It would be good to add docs that show how you might calculate this cost.

sherifkozman:
Great, i have tried to run a sample simulation over a Google Cloud TCO based on the information from terraform and other configs but was missing some assumptions such as expected batch download sizes Vs Exposure keys volume.

PR #633, opened 2020-06-15
Docs for Estimating Costs

crwilcox:
#632

### DIFF
--- a/docs/_data/navigation.yml
+++ b/docs/_data/navigation.yml
@@ -11,6 +11,8 @@ docs:
           url: /getting-started/publishing-temporary-exposure-keys
         - title: How to Download Keys
           url: /getting-started/downloading-export-batches-keys
+        - title: Estimating the Cost of Deployment
+          url: /getting-started/estimating-costs
     - title: Design
       children:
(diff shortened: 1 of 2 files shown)


mikehelmick:
> It would be great to add a minimum configuration estimate as well.

I think it depends on what you mean by minimum configuration. Like a dev instance?
Or a small production instance?

size of production instances depends on the number of users and estimated infection rates.

sherifkozman:
@mikehelmick The updates @crwilcox has pushed gives a decent indication of how to calculate resources with some fine tuning such as using CDN and possible use of a non managed PostgreSQL will bring costs even more down. But this is the decision of the HA on how to manage and run infrastructure options.

crwilcox:
@sherifkozman I am going to leave the network egress estimate as $10k+. Assuming US only traffic it may be closer to $6k, but if to AU it is $11.5k. And I would rather err on the side of overestimating, especially since I have already assumed partial adoption.

@krishnadurai I can add the smaller sql instance we mention for dev environments but to @mikehelmick point it is hard to give a minimum deployment without knowing the space it is being deployed for.

PR #634, opened 2020-06-16
add ability to expose metrics from /metrics endpoint

### LINKED ISSUE #437: non-logs-based metrics
In GCP, and probably in some OSS solutions, logs-based metrics can be used solely to count the number of log lines matching a search. This doesn't work for things like latency distributions. We should use a real metrics framework like Prometheus.

frodopwns:
closes #437 

Creates a configuration option to expose /metrics endpoints for metrics scraping via Prometheus or Azure Monitor for Containers.

Didn't add this to Federation-Out since it is GRPC.

Let me know if you would prefer this be implemented somewhere else.

### DIFF
--- a/internal/observability/observability.go
+++ b/internal/observability/observability.go
@@ -71,7 +71,7 @@ func exporter() traceAndViewExporter {
 		// TODO: Add other trace and view exporters and break out of this.
 		return nil
 
-	case "OCAGENT":
+	case "OCAGENT", "PROMETHEUS":
 		oce, err := ocagent.NewExporter(ocagent.WithInsecure(), ocagent.WithAddress("localhost:55678"))
 		if err != nil {
 			panic(err)
--- a/docs/getting-started/deploying.md
+++ b/docs/getting-started/deploying.md
@@ -159,7 +159,7 @@ To enable distributed tracing, please ensure your environment has these variable
 
 Variable|Values|Comment
 ---|---|---
-OBSERVABILITY_EXPORTER|If unset, no exporting shall be done. Use any of "stackdriver" or "ocagent" otherwise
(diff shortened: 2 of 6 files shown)


sethvargo:
Hey @frodopwns this looks great, but it'd probably be better to move the wrapper into the `server` package instead. What do you think? That would remove all the dupe logic in the commands. Could you also make sure you run `go mod tidy`?

frodopwns:
@sethvargo reasonable request, I'm on it....also, I ran the tidy command but didn't see a change, will try again to be sure. Thanks!

internal/observability/observability.go
```diff
@@ -88,3 +91,18 @@ func exporter() traceAndViewExporter {
...
+// ExposeMetricsIfPrometheus adds the /metrics endpoint to a mux if the correct env var is set
+func ExposeMetricsIfPrometheus(mux *http.ServeMux) error {
+	exporter := os.Getenv("OBSERVABILITY_EXPORTER")
+	if strings.EqualFold(exporter, "prometheus") {
+		exporter, err := prometheus.NewExporter(prometheus.Options{})
+		if err != nil {
+			return fmt.Errorf("failed to create prometheus exporter: %v", err)
+		}
+
+		mux.Handle("/metrics", exporter)
```
mikehelmick, line 104: how is access controlled?

frodopwns: Seems like we may need to spin up a separate server in a go process so we can serve this on a different port.

mikehelmick: either that or front these servers w/ some kind of proxy that can apply ACLs based on path.

on cloud run, we can only listen on a single port, but also wouldn't make sense to enable prometheus in that environment anyway.

frodopwns: I had been thinking of using the kube-rbac-proxy project but am now thinking a different port would be nicer.

frodopwns: @mikehelmick I refactored as mentioned above. Could use another look when you get a chance.

mikehelmick: thanks! looking now

PR #641, opened 2020-06-17
feat(observability): use envconfig to manage configuration for the observability exporter

ultimateboy:
1. Switches the observability exporter configuration to use envconfig
2. Adds an interface for ObservabilityExporter with a single GenericExporter implementation
3. Adds a configuration option to the OCAGENT exporter to set the exporter endpoint and control insecure ssl.
4. Completed a "todo" in the code to make the sample probability rate configurable.
5. Added a NOOP observability exporter to be the default exporter if none are provided and to aid the testing in setup package (the default is still stackdriver due to the envconfig struct tag).

I started this PR by simply adding the single configuration option to set the exporter endpoint, then things kinda spiraled out of control from there. I wanted to keep it simple, but as I dug in to make things configurable, it started to make sense to treat the observability exporter similarly to other providers and interface it out. Happy to hear suggestions on how this could be improved.

This unblocks sending trace data from OpenCensus to Azure Application Insights by way of the "local forwarder". More information on that is available here: https://opencensus.io/exporters/supported-exporters/go/applicationinsights/ and https://github.com/microsoft/ApplicationInsights-LocalForwarder

### DIFF
--- a/internal/observability/config.go
+++ b/internal/observability/config.go
@@ -1,0 +1,51 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package observability sets up and configures observability tools.
+package observability
+
+// ExporterType represents a type of observability exporter.
+type ExporterType string
+
+const (
+	ExporterStackdriver ExporterType = "STACKDRIVER"
+	ExporterPrometheus  ExporterType = "PROMETHEUS"
+	ExporterOCAgent     ExporterType = "OCAGENT"
+	ExporterNoop        ExporterType = "NOOP"
+)
+
+// Config holds all of the configuration options for the observability exporter
+type Config struct {
+	ExporterType ExporterType `env:"OBSERVABILITY_EXPORTER, default=STACKDRIVER"`
+
+	OpenCensusConfig
+	StackdriverConfig
+	OCAgentConfig
+}
+
+// OpenCensusConfig holds the configuration options for the open census exporter
+type OpenCensusConfig struct {
+	TraceProbabilitySampleRate float64 `env:"TRACE_PROBABILITY, default=0.40"`
+}
+
+// StackdriverConfig holds the configuration options for the stackdriver exporter
+type StackdriverConfig struct {
+	ProjectID string `env:"PROJECT_ID, default:$GOOGLE_CLOUD_PROJECT"`
+}
+
+// OCAgentConfig holds the configuration options for the default opencensus exporter
+type OCAgentConfig struct {
+	Insecure bool   `env:"OCAGENT_INSECURE, default=true"`
+	Endpoint string `env:"OCAGENT_TRACE_EXPORTER_ENDPOINT"`
+}
--- a/internal/observability/generic.go
+++ b/internal/observability/generic.go
@@ -1,0 +1,68 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package observability sets up and configures observability tools.
+package observability
+
+import (
+	"sync"
+
+	"go.opencensus.io/plugin/ocgrpc"
+	"go.opencensus.io/plugin/ochttp"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/trace"
+)
+
+// Compile-time check to verify implements interface.
+var _ Exporter = (*GenericExporter)(nil)
+
+var initExporterOnce sync.Once
+
+type traceAndViewExporter interface {
+	trace.Exporter
+	view.Exporter
+}
+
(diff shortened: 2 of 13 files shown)


internal/observability/config.go
```diff
@@ -0,0 +1,51 @@
...
+	ExporterPrometheus  ExporterType = "PROMETHEUS"
+	ExporterOCAgent     ExporterType = "OCAGENT"
+	ExporterNoop        ExporterType = "NOOP"
+)
+
+// Config holds all of the configuration options for the observability exporter
+type Config struct {
+	ExporterType ExporterType `env:"OBSERVABILITY_EXPORTER, default=STACKDRIVER"`
+
+	*OpenCensusConfig
```
sethvargo, line 32: envconfig will initialize these so they aren't nil, but they could be nil in tests or when envconfig hasn't run yet (config.OpenCensusConfig.X would NPE).

sethvargo: We get around this elsewhere by embedding a non-pointer

internal/observability/config.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// StackdriverConfig holds the configuration options for the stackdriver exporter
+type StackdriverConfig struct {
+	ProjectID string `env:"PROJECT_ID"`
+}
+
+// OCAgentConfig holds the configuration options for the default opencensus exporter
+type OCAgentConfig struct {
+	Insecure bool   `env:"OCAGENT_INSECURE, default=true"`
```
sethvargo, line 49: Should the default be false?

internal/observability/config.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// OpenCensusConfig holds the configuration options for the open census exporter
+type OpenCensusConfig struct {
+	TraceProbabilitySampleRate float64 `env:"TRACE_PROBABILITY, default=0.40"`
+}
+
+// StackdriverConfig holds the configuration options for the stackdriver exporter
+type StackdriverConfig struct {
+	ProjectID string `env:"PROJECT_ID"`
```
sethvargo, line 44: Can you have this default to GOOGLE_CLOUD_PROJECT? `,default:$GOOGLE_CLOUD_PROJECT` _should_ work.

internal/observability/observability.go
```diff
@@ -16,75 +16,56 @@
...
-	"go.opencensus.io/stats/view"
-	"go.opencensus.io/trace"
 )
 
-func init() {
-	initExporterOnce.Do(initExporter)
+// Exporter defines the minimum shared functionality for an observability exporter
+// used by this application.
+type Exporter interface {
+	InitOnce()
```
sethvargo, line 29: Can we use a less common thing here like `InitExport()` just to avoid a collision with another iface?

internal/observability/observability.go
```diff
@@ -16,75 +16,56 @@
...
-	view.Exporter
-}
-
-func exporter() traceAndViewExporter {
-	switch strings.ToUpper(os.Getenv("OBSERVABILITY_EXPORTER")) {
+// NewFromEnv returns the observability exporter given the provided configuration, or an error
+// if it failed to be created.
+func NewFromEnv(ctx context.Context, config *Config) (Exporter, error) {
+	switch config.ExporterType {
 	default:
```
sethvargo, line 36: Should there be a default now?

internal/setup/setup.go
```diff
@@ -60,6 +61,12 @@ type SecretManagerConfigProvider interface {
 	SecretManagerConfig() *secrets.Config
 }
 
+// ObservabilityExporterConfigProvider signals that the config knows how to configure an
+// observability exporter.
+type ObservabilityExporterConfigProvider interface {
```
sethvargo, line 66: Nit - the config providers are alphabetized in this file

internal/setup/setup.go
```diff
@@ -203,5 +210,20 @@ func SetupWith(ctx context.Context, config interface{}, l envconfig.Lookuper) (*
 		}
 	}
 
+	// Configure and initialize the observability exporter.
```
sethvargo, line 213: This should probably happen after the secret manager and kms setup (before database)

internal/observability/observability.go
ultimateboy: Sure. Does `InitExportOnce()` make sense to you? I think the "once" part is important.

internal/observability/observability.go
ultimateboy: Could definitely just default it to NOOP

internal/observability/config.go
ultimateboy: Great question. I didnt want to change the default that was there previously, which is why I stuck with true.

sethvargo: Let's leave it true, but can you file an issue with the kind/security tag and we'll take a look at changing it.

internal/observability/observability.go
sethvargo: What would happen if it were invoked twice?

internal/observability/observability.go
sethvargo: Yea, lets do that

internal/observability/config.go
ultimateboy: Will do. Just going to wait for a day until either this PR merges or not so that I can link to the correct spot in the code in the issue.

internal/observability/generic.go
```diff
@@ -0,0 +1,68 @@
...
+	trace.ApplyConfig(trace.Config{
+		DefaultSampler: trace.ProbabilitySampler(g.sampleRate),
+	})
+	trace.RegisterExporter(g.exporter)
+	view.RegisterExporter(g.exporter)
+
+	// Record the various HTTP view to collect metrics.
+	httpViews := append(ochttp.DefaultServerViews, ochttp.DefaultClientViews...)
+	if err := view.Register(httpViews...); err != nil {
+		panic(err)
```
sethvargo, line 61: Sorry I missed this in the first pass. I don't think we should panic, especially since we control when `InitExportOnce` is called. Can we change the interface to return an error and catch that error in setup?

sethvargo: (Yes, this was the behavior before, but while we're refactoring... 😄)

ultimateboy: Thanks @sethvargo - thats not as simple as it seems because the `sync.Once`'s `.Do()` function wraps the initExporter() and does not return an error.

ultimateboy: Guess I'll just grab the err from the .Do and return it anyways. Should be fine. PR coming shortly.

ultimateboy: See https://github.com/google/exposure-notifications-server/pull/646

Issue #648, opened 2020-06-19
Remove error code flattening and latency injection


mikehelmick:
Now that API calls are authenticated via the health authority certificates, return useful error codes and stop injecting latency.

/assign

PR #649, opened 2020-06-19
Remove latency delay and return real responses

sethvargo:
Fixes GH-648


internal/publish/publish.go
```diff
@@ -120,9 +117,8 @@ func (h *publishHandler) handleRequest(w http.ResponseWriter, r *http.Request) r
 		// isn't transient.
 		// This message (and logging) will only contain the AppPkgName from the request
 		// and no other data from the request.
-		msg := fmt.Sprintf("no AuthorizedApp, dropping data: %v", err)
-		logger.Error(msg)
-		span.SetStatus(trace.Status{Code: trace.StatusCodeInternal, Message: msg})
+		message := fmt.Sprintf("no AuthorizedApp, dropping data: %v", err)
+		span.SetStatus(trace.Status{Code: trace.StatusCodeInternal, Message: message})
 		return response{
 			status:      http.StatusInternalServerError,
```
mikehelmick, line 123: should this be unauthorized now?

sethvargo: Done

internal/publish/publish.go
```diff
@@ -120,12 +117,11 @@ func (h *publishHandler) handleRequest(w http.ResponseWriter, r *http.Request) r
...
-		msg := fmt.Sprintf("no AuthorizedApp, dropping data: %v", err)
-		logger.Error(msg)
-		span.SetStatus(trace.Status{Code: trace.StatusCodeInternal, Message: msg})
+		message := fmt.Sprintf("no AuthorizedApp, dropping data: %v", err)
+		span.SetStatus(trace.Status{Code: trace.StatusCodeInternal, Message: message})
 		return response{
-			status:      http.StatusInternalServerError,
-			message:     http.StatusText(http.StatusInternalServerError),
+			status:      http.StatusUnauthorized,
+			message:     message,
```
mikehelmick, line 124: sorry - should have caught this - this should be the standard http.StatusText( as well and not say "NoAuthorizedApp found"

sethvargo: How would the app know why it was unauthorized then? We already return the real message for if the app posts to a bad region, for example.

mikehelmick: true.

PR #660, opened 2020-06-23
feat(secrets-manager): add ability to json-decode secret values and select the value of a map by specifying the key in the secret-name

ultimateboy:
Adds a configuration option "SECRET_EXPANSION" (defaults to false) which enables arbitrary json decoding of secret values to select the value from a map.

For example:
If a secret with a name of "psqlcreds" has a json value of `{"username":"gandalf", "password":"abc"}`
When `GetSecretValue(ctx, "psqlcreds")` is called, the raw json value will be returned.
When `GetSecretValue(ctx, "psqlcreds.username")` is called, only "gandalf" (without quotes) will be returned.

This also works for nested values. For example:
If a secret with a name of "psqlcreds" has a json value of `{"creds": {"username":"gandalf", "password":"abc"}}`
When `GetSecretValue(ctx, "psqlcreds.creds.username")` is called, only "gandalf" (without quotes) will be returned.

This simplifies the use of every other secrets-managers in that all of the connection details can be provided in a single secret. For the database, this is especially handy as there are a number of different fields needed. Attempting to use the cloud-provider secrets-managers without this requires a lot of individual secrets to be created in the store.

### DIFF
--- a/pkg/secrets/json_expander.go
+++ b/pkg/secrets/json_expander.go
@@ -1,0 +1,91 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package secrets
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+type JSONExpander struct {
+	sm SecretManager
+}
+
+// NewJSONExpander creates a new secret manager that allows secret values to be stored as json.
+// When resolving secrets, if "dot-notation" is provided via the secret-name, the secret value
+// will be json-decoded and the dot-notation will be used to resolve the secret value.
+func NewJSONExpander(ctx context.Context, f SecretManagerFunc) (SecretManager, error) {
+	sm, err := f(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new JSONExpander SecretManager: %w", err)
+	}
+
+	return WrapJSONExpander(ctx, sm)
+}
+
+// WrapJSONExpander wraps an existing SecretManager with json-expansion logic.
+func WrapJSONExpander(ctx context.Context, sm SecretManager) (SecretManager, error) {
+	return &JSONExpander{
+		sm: sm,
+	}, nil
+}
+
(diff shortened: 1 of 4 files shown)


pkg/secrets/json_expander.go
```diff
@@ -0,0 +1,85 @@
...
+		return "", fmt.Errorf("secret-name has more than one period - json expansion path undeterminable")
+	}
+
+	// Get the value from the embdedded secret-manager using what comes before the period.
+	val, err := sm.sm.GetSecretValue(ctx, parts[0])
+	if err != nil {
+		return "", err
+	}
+
+	json, err := simplejson.NewJson([]byte(val))
```
sethvargo, line 73: Could we  unmarshal into a `map[string]string` (or `map[string]interface{}`) instead of depending on a library? It'll be a few extra LOC, but isn't super complex. We could also move it into a helper for easier testing.

pkg/secrets/json_expander.go
```diff
@@ -0,0 +1,85 @@
...
+// If a secret with a name of "psqlcreds" has a value of `{"username":"gandalf", "password":"abc"}`
+// When GetSecretValue(ctx, "psqlcreds") is called, the raw json value will be returned.
+// When GetSecretValue(ctx, "psql.username") is called, only "gandalf" (without quotes) will be returned.
+func (sm *JSONExpander) GetSecretValue(ctx context.Context, name string) (string, error) {
+	parts := strings.Split(name, ".")
+	if len(parts) == 1 {
+		return sm.sm.GetSecretValue(ctx, name)
+	}
+
+	if len(parts) > 2 {
```
sethvargo, line 63: I think we can handle deep nesting if we use native json:

```golang
var m map[string]interface{}
if err := json.Unmarshal(val, &m); err != nil {
  return err
}

for _, p := range parts[:len(parts)-1] {
  v, ok := m[p]
  if !ok {
    return fmt.Errorf("missing key %v", p)
  }  

  m, ok = v.(map[string]interface{})
  if !ok {
    return fmt.Errorf("not a nested field: %v", p)
  }
}

val, ok := m[len(parts)]
if !ok {
  return fmt.Errorf("missing key %v", p)
}

return val, nil

```

ultimateboy: I like that idea. I kinda wanted to start with the go-simplejson route just to make sure the general idea is acceptable. I think putting a little work into not bringing in another import makes sense.

sethvargo:
Hey @ultimateboy - I like the idea. I made a suggestion about handling deep nesting.

pkg/secrets/json_expander.go
ultimateboy: Done. Thanks for the suggestion. I like the added ability to have nested values!

pkg/secrets/json_expander.go
ultimateboy: Done. Thanks for the code. I mostly was able to copy and paste it in place.

sethvargo:
@ultimateboy tag me when you want re-review

ultimateboy:
@sethvargo Should be ready for review at this time. Sorry about the multiple pushes.

PR #667, opened 2020-06-26
Allow overlapped keys in Debug configurations

crwilcox:
b/159869725

In testing, we sometimes don't publish keys immediately, to the server this looks like overlapped keys. This makes the server log an error, but it won't fail

### DIFF
--- a/internal/publish/model/exposure_model.go
+++ b/internal/publish/model/exposure_model.go
@@ -249,7 +249,8 @@ func (t *Transformer) TransformPublish(ctx context.Context, inData *verifyapi.Pu
 	nextInterval := entities[0].IntervalNumber
 	for _, ex := range entities {
 		if ex.IntervalNumber < nextInterval {
-			return nil, fmt.Errorf("exposure keys have overlapping intervals")
+			logging.FromContext(ctx).Errorf("Server detected keys with overlapping intervals")
+			break
 		}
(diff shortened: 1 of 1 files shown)


sethvargo:
/hold

It looks like there's some recent comments from @flagxor on the bug we should resolve first.

internal/publish/model/exposure_model.go
```diff
@@ -249,7 +249,8 @@ func (t *Transformer) TransformPublish(ctx context.Context, inData *verifyapi.Pu
 	nextInterval := entities[0].IntervalNumber
 	for _, ex := range entities {
 		if ex.IntervalNumber < nextInterval {
-			return nil, fmt.Errorf("exposure keys have overlapping intervals")
+			logging.FromContext(ctx).Errorf("Server detected keys with overlapping intervals")
```
gurayAlsac, line 252: We might base it on whether debugReleaseSameDay is true, since this should only arise in testing.

crwilcox: I pushed an update to gate this on the debug config.

flagxor:
Ok, updated the bug.
This matches API implementor's understanding of the intended behavior.
There may be some crossed wires on this, but for now server should probably tolerate this.

lgtm

mikehelmick:
I don't believe that overlapping keys is a valid production status. Will followup.

PR #675, opened 2020-06-30
Refactor integration test package to allow it consumed by others, like performance test

chaodaiG:
/cc @stati0n

### DIFF
--- a/internal/integration/integration.go
+++ b/internal/integration/integration.go
@@ -41,7 +41,8 @@ const (
 	exportDir = "my-bucket"
 )
 
-func TestServer(tb testing.TB, ctx context.Context, exportPeriod time.Duration) (*serverenv.ServerEnv, *EnServerClient, *database.DB) {
+// NewTestServer sets up clients used for integration tests
+func NewTestServer(tb testing.TB, ctx context.Context, exportPeriod time.Duration) (*serverenv.ServerEnv, *EnServerClient, *database.DB) {
 	env, client := testServer(tb)
 	db := env.Database()
 	enClient := &EnServerClient{client: client}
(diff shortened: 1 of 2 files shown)


internal/integration/integration.go
```diff
@@ -31,8 +31,30 @@ import (
...
+	authorizedappmodel "github.com/google/exposure-notifications-server/internal/authorizedapp/model"
+	exportdatabase "github.com/google/exposure-notifications-server/internal/export/database"
+	exportmodel "github.com/google/exposure-notifications-server/internal/export/model"
+)
+
+const (
+	exportDir = "my-bucket"
 )
 
+func TestServer(tb testing.TB, ctx context.Context, exportPeriod time.Duration) (*serverenv.ServerEnv, *EnServerClient, *database.DB) {
```
sethvargo, line 44: We should make this NewTestServer or something more explicit

chaodaiG: Agree and updated

internal/integration/integration.go
```diff
@@ -31,8 +31,31 @@ import (
...
+	exportmodel "github.com/google/exposure-notifications-server/internal/export/model"
+)
+
+const (
+	exportDir = "my-bucket"
 )
 
+// NewTestServer sets up clients used for integration tests
+func NewTestServer(tb testing.TB, ctx context.Context, exportPeriod time.Duration) (*serverenv.ServerEnv, *EnServerClient, *database.DB) {
+	env, client := testServer(tb)
```
stati0n, line 46: Does this need to be `TestServer`?

stati0n: Never mind.  Think I got something wrong when I checked out and merged your branch ;)

chaodaiG: :)

PR #681, opened 2020-07-02
Fixes to export index.txt file creation

### LINKED ISSUE #672: index.txt crosses export configs
### TL;DR

When there are multiple export configs on a single instance, the index.txt file seems to cross the streams as it were.

For example, this is a sample from an export that should only contain "CH" region TEKs

```
exposureKeyExport-CH/1593446400-1593446700-00001.zip
exposureKeyExport-US/1592240400-00001.zip
```

**Expected behavior**

The different export configs have separate index.txt files

**Observed behavior**

They were merged

mikehelmick:
* When generating the files for the index.txt, restrict it to the exportconfig.ID of the batch being processed
* When determining if an empty batch should refresh the index, do this per-configID instead of per /do-work
* Change index file locking to per-config id and not per-batch, better represents the race condition we are guarding against.

Fixes #672

### DIFF
--- a/internal/debugger/handler_debug.go
+++ b/internal/debugger/handler_debug.go
@@ -56,7 +56,7 @@ func (s *Server) handleDebug(ctx context.Context) http.HandlerFunc {
 
 		ExportConfigs    []*exportmodel.ExportConfig
 		ExportBatchEnds  map[int64]*time.Time
-		ExportBatchFiles []string
+		ExportBatchFiles map[int64][]string
 
 		SignatureInfos []*exportmodel.SignatureInfo
 	}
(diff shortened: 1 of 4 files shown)


internal/debugger/handler_debug.go
```diff
@@ -104,6 +105,17 @@ func (s *Server) handleDebug(ctx context.Context) http.HandlerFunc {
 				return fmt.Errorf("failed to get all export configs: %w", err)
 			}
 			resp.ExportConfigs = exportConfigs
+
+			queue(&wg, errCh, func() error {
+				for _, ec := range exportConfigs {
+					exportBatchFiles, err := exportDB.LookupExportFiles(ctx, ec.ConfigID, 4*time.Hour)
```
gurayAlsac, line 111: Not technically part of this PR but something I just noticed: does the "4" need to be configurable or based on some existing configuration?

PR #683, opened 2020-07-03
Fix Cache-Control header

zssz:
https://github.com/googleapis/google-cloud-go/commit/16dbcc9af7869d37fa7c8625733aab6628b7c2e4

`Metadata` sets the "x-goog-meta-Cache-Control" and _not_ the "Cache-Control" header.


sethvargo:
Hi @zssz - can you please run `go fmt` and make sure the presubmits are passing. Looks like the linter is sad about an extra newline character.

sethvargo:
/retest

@mikehelmick @crwilcox I think we have a flakey test

PR #691, opened 2020-07-06
gh-pages

crwilcox:


jixiangox:
Cat all files into single file in gh head; svg file format diagram is not viewable in GitHub App.

Issue #705, opened 2020-07-08
Posting Keys from previosly working clients results error 400


sherifkozman:
Recenetly submitting exposure Keys from clients that were succesffully working before started giving : 
Error 400 : unable to read request data: exposure keys have overlapping intervals 
Any pointers on how to trace that and the root cause ?

mikehelmick:
if you set [this flag](https://github.com/google/exposure-notifications-server/blob/main/internal/publish/config.go#L50) to true, it will allow that upload to work.

We're still in process for upgrading the server for the "v1.5" model, see #663 

In this new model - there will be 2 keys for the last day, one that ends earlier and one that continues for the rest of the day.

sherifkozman:
@mikehelmick This worked fine for now, thanks

sherifkozman:
Hello @mikehelmick even after lifting the embargo over the Keys and setting DEBUG_RELEASE_SAME_DAY_KEYS=true for the exposure service we are still getting error 400

mikehelmick:
Same error message?

Can you share the server logs?

And can you share the data that is being uploaded? You can redact the keys, just need to see the timing information.

sherifkozman:
So we are getting a bunch of errors of overlapping Keys through the mobile app that was previously working.
Yesterday they were generating Error 400 with "exposure keys have overlapping intervals" today they are being accepted with the same warning.

[keys.txt](https://github.com/google/exposure-notifications-server/files/4897612/keys.txt)

mikehelmick:
I pulled out just the timing information from the file.

```
    - rollingPeriod : 90
    - rollingStartNumber : 2657088
    - rollingPeriod : 89
    - rollingStartNumber : 2657088
    - rollingPeriod : 89
    - rollingStartNumber : 2657088
    - rollingPeriod : 88
    - rollingStartNumber : 2657088
    - rollingPeriod : 144
    - rollingStartNumber : 2656944
    - rollingPeriod : 133
    - rollingStartNumber : 2656944
    - rollingPeriod : 116
    - rollingStartNumber : 2656944
    - rollingPeriod : 113
    - rollingStartNumber : 2656944
    - rollingPeriod : 112
    - rollingStartNumber : 2656944
    - rollingPeriod : 112
    - rollingStartNumber : 2656944
    - rollingPeriod : 112
    - rollingStartNumber : 2656944
    - rollingPeriod : 112
    - rollingStartNumber : 2656944
    - rollingPeriod : 99
[...]

PR #708, opened 2020-07-10
additional logging on transform failures

mikehelmick:
Additional logging for #705


sherifkozman:
We will add more logging to the device as well to be able to capture responses

Issue #250, opened 2020-05-14
test: create load test for exposure key batch cleanup


crwilcox:

sethvargo:
Oh hey, stalebot works 😄

chaodaiG:
This was fixed in https://github.com/google/exposure-notifications-server/pull/713

Issue #243, opened 2020-05-14
cleanup: load test


crwilcox:
Should verify how long a run takes at scale. Also verify it doesn't adversely impact table performance during cleanup.

sethvargo:
@stati0n I think you mentioned you wanted to work on this.

chaodaiG:
The first half was done in https://github.com/google/exposure-notifications-server/pull/713

Will need to figure out how to test this `Also verify it doesn't adversely impact table performance during cleanup.`

stati0n:
@crwilcox is "table performance" in this case simultaneously inserting data while running cleanup?

MushuEE:
@stati0n that is what I understand from syncing with @chaodaiG, I was going to try to repopulate the DB with a blob or docker image. Currently cleanup takes such a short time that it is hard to determine impact.

MushuEE:
Finally have time to try and tackle this. 
Background for future me/others:

We use a postgres:12-alpine container to create the local integration/e2e test database. This is then populated in the test.

Plan is to pre-populate the database via either a standalone script or the test (I think I might want a standalone script to make the process more flexible and re-usable -if we want to increase / decrease database size-). Then store that blob backup either in source or gcs.  At the beginning of the tests we will restore from backup to pre-populate the db quickly. The current method of loading keys takes some time (~20m). This will hopefully both speed up the tests along with allowing much more data to be injected, enabling us to better gauge the cleanup impact.

Helpful commands
```
docker exec -it <name> pg_dump "dbname=en-server user=my-username password=abcd1234" > backup.sql
docker exec -it <name> psql "dbname=en-server user=my-username password=abcd1234" -c '\z'
```

MushuEE:
@chaodaiG I am not sure what the next step to take here is. I was not able to go far past 100k Publishers and cleanup time seemed to never be impacted. If we need many more publishers I might need to expand the memory space of the postgress because I start getting failures at high numbers and my only idea is there could be a space issue with say 200k Publishers

chaodaiG:
I think this either means the performance of cleanup is very great, or there were not enough exports generated. Could you try to log how many exports(exported files) were generated for 100k publishes?

MushuEE:
Sure, will do

MushuEE:
I believe this is verified. I ran a number of tests with up to 400,000 publishers or 5.6mil keys. This required me to change the export config timeout and max records:
```
	// Export
	exportConfig := &export.Config{
[...]

PR #713, opened 2020-07-13
Load test for cleanup exports

chaodaiG:
Part of https://github.com/google/exposure-notifications-server/issues/250 and https://github.com/google/exposure-notifications-server/issues/243

This PR adds load test for cleanup exports. Worth noting that:
- Exports were "generated" slightly differently from a standard procedure(a.k.a. `/export` endpoint)
- Only added a single test scenario of 10000 exports + 3 keys in each
- The new functions in `integration.go` were copy/pasted from their old locations

/cc @stati0n

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,3 +57,13 @@ test-acc:
 		-vet="${VETTERS}" \
 		./...
 .PHONY: test-acc
+
+performance-test:
+	@go test \
+		-count=1 \
+		-timeout 10m \
+		-vet="${VETTERS}" \
+		-v \
+		-tags=performance \
+		./internal/performance
+.PHONY: performance-test
--- a/Makefile
+++ b/Makefile
@@ -61,8 +61,7 @@ test-acc:
 performance-test:
 	@go test \
 		-count=1 \
-		-timeout 10m \
-		-vet="${VETTERS}" \
+		-timeout=10m \
 		-v \
 		-tags=performance \
 		./internal/performance
--- a/internal/integration/helpers.go
+++ b/internal/integration/helpers.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"path"
 	"testing"
 	"time"
 
@@ -35,6 +36,7 @@ import (
 	"github.com/google/exposure-notifications-server/pkg/keys"
 	"github.com/google/exposure-notifications-server/pkg/secrets"
 	"github.com/google/exposure-notifications-server/pkg/server"
+	"github.com/sethvargo/go-retry"
 
 	authorizedappmodel "github.com/google/exposure-notifications-server/internal/authorizedapp/model"
 	exportdatabase "github.com/google/exposure-notifications-server/internal/export/database"
@@ -258,7 +260,7 @@ func testClient(tb testing.TB, srv *server.Server) *http.Client {
 	}
 
 	return &http.Client{
-		Timeout:   5 * time.Second,
+		Timeout:   50 * time.Second,
 		Transport: prt,
 	}
 }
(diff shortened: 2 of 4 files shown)


chaodaiG:
This PR assumes the mako logic will be setup from the PR that @stati0n is working on

internal/integration/integration.go
```diff
@@ -255,3 +260,96 @@ func createSignatureInfo(ctx context.Context, db *database.DB, exportPeriod time
...
+	enClient.ExportBatches(tb)
+
+	tb.Logf("Waiting %v before starting workers", 500*time.Millisecond)
+	time.Sleep(500 * time.Millisecond)
+	enClient.StartExportWorkers(tb)
+}
+
+// GetKeysFromAllBatches gets all exported keys, unless the filename is
+// explicitly listed in excluded
+func GetKeysFromAllBatches(t *testing.T, ctx context.Context, env *serverenv.ServerEnv) []*pbexport.TemporaryExposureKeyExport {
```
sethvargo, line 289: Wanna make these TB too?

scripts/performance.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
sethvargo, line 1: Can we use a Makefile target instead like the other tests?

internal/integration/integration.go
chaodaiG: Good point, done

scripts/performance.sh
chaodaiG: If just a `go test` it would for sure be a good Makefile target. But performance test will have some setup steps for mako, which will be added by @stati0n in the upcoming PR. I'm afraid that the logic of mako setup will be too large to be fit in a Makefile

stati0n: Let's assume [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) for the moment and correct it in a future PR if we need to.

chaodaiG: Good idea, done

chaodaiG:
Failed a different test this time, flakiness?
/retest

chaodaiG:
@sethvargo , I have rebased on top of #733 , could you take a look again?

chaodaiG:
Seems like it went through a pretty bad rebase, let me figure it out first

chaodaiG:
Rebase completed, now it's ready for review

Makefile
```diff
@@ -57,3 +57,13 @@ test-acc:
 		-vet="${VETTERS}" \
 		./...
 .PHONY: test-acc
+
+performance-test:
+	@go test \
+		-count=1 \
+		-timeout 10m \
+		-vet="${VETTERS}" \
```
sethvargo, line 65: We probably don't need to run the vetters here, but it's probably fine

Makefile
```diff
@@ -57,3 +57,13 @@ test-acc:
 		-vet="${VETTERS}" \
 		./...
 .PHONY: test-acc
+
+performance-test:
+	@go test \
+		-count=1 \
+		-timeout 10m \
```
sethvargo, line 64: nit: can you make this consistent (we use flag=value elsewhere)

internal/integration/helpers.go
```diff
@@ -258,7 +260,7 @@ func testClient(tb testing.TB, srv *server.Server) *http.Client {
 	}
 
 	return &http.Client{
-		Timeout:   5 * time.Second,
+		Timeout:   50 * time.Second,
```
sethvargo, line 263: This seems like a really long timeout - can you explain more about the increase please?

internal/performance/cleanup_test.go
```diff
@@ -0,0 +1,226 @@
+// +build performance
```
sethvargo, line 1: This looks _very_ similar to the existing integration test. Is there a way we can merge them into a single test to reduce duplication?

internal/integration/helpers.go
chaodaiG: I have documented above this. For now it'll work, for long term, I can export *http.Client in *integration.Client so that the timeout can be customized for each test

internal/performance/cleanup_test.go
sethvargo: What is "reasonable time"? Are we talking hours or minutes? I think it's okay if the integration tests take a few minutes to be honest. Especially if we can reduce the maintenance overhead.

chaodaiG: I'm afraid it's hours, the exports has to be done in sequential order, and considering the db lock, each request takes at least 2 seconds in my past experiments, so we are talking about 2 * 10000 seconds, which is almost 6 hours of time here. 

I do agree with you that it's desired to reduce this the maintenance overhead, will spend some more time to see whether I can squeeze the time spent for each export

chaodaiG: Ack, issue created here https://github.com/google/exposure-notifications-server/issues/751, and the tests were fixed

PR #716, opened 2020-07-13
Fix README errors in terraform guide

chaodaiG:
Copy/paste works only when there is no indent

/cc @sethvargo

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -72,13 +72,13 @@ For full instructions on deploying, view the [deployment docs](../docs/getting-s
     Configure Terraform to store state in the bucket:
 
     ```text
-    $ cat <<EOF > ./state.tf
-    terraform {
-      backend "gcs" {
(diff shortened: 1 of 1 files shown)


sethvargo:
There isn't an indent on the [rendered markdown](https://github.com/google/exposure-notifications-server/blob/main/terraform/README.md). Your change breaks the markdown renderer:

screenshot-20200713-210146@2x

sethvargo:
/-1

jixiangox:
The copy/paste is ok when you use the terminal in browser; indent is not matched when your terminal is in you local computer.

chaodaiG:
> There isn't an indent on the [rendered markdown](https://github.com/google/exposure-notifications-server/blob/main/terraform/README.md). Your change breaks the markdown renderer:
> 
> screenshot-20200713-210146@2x
> 

That's right. My bad, was copy pasting from my IDE

/close

Issue #663, opened 2020-06-23
Add support for v1.5 API


flagxor:
### TL;DR
v1.5 adds some additional fields attached to keys that need to be handled by the Key server.
Limited support for change of key status + revocation is supported in the API shape.

### Design Goals

Additional fields need to be selected by the HA app:
* report_type - unambiguous source of the diagnosis
* days_since_onset_of_symptoms - offset to the date on which symptoms first appeared

The HA also has the opportunity to make a single revision to a previously published key:
* Changes are placed in revised_keys

Only some changes are allowed (enforced by the client, but server can also check):
Recursive -> Confirmed Test, Revoked, Clinical Diagnosis, Self-Report
Self-Report ->  Confirmed Test, Revoked, Clinical Diagnosis
Clinical Diagnosis -> Confirmed Test, Revoked

To support these fields well the Diagnosis Key server will need to:
* Allow the app to provide an optional days since first onset and attach this to keys
* Allow the app to suggest a reportType, but limit some report types to those certified by a verification server

Revision + revocation will require dispensing some secret to the client to allow later revalidation to change a previously released key.
Revision + revocation support in the reference server is of secondary priority to supporting the new key fields.

Transmission risk is deprecated, but can be co-present with the other fields to support v1.0 clients.

mikehelmick:
I'll take this on next week.

mikehelmick:
Is revoked in this case, a user initiated action or does that need to be run through a verification server?

mikehelmick:
Some high level notes

- In this reference sever we will not support uploads of "recursive" or "self-report" at this time, but instead continue to recommend the use of the verification server protocol.
- Symptom onset date will be accepted and then assigned by the server. This field can also be present on the verification certificate. Ideally that is where the field will come from.
- Report type __must__ come from a verification certificate, we won't let apps assign this on their own
- For revised keys:
  - the verification system itself doesn't know that it is revising a key. 
  - For HA verification there are 2 paths we can support
    - likely -> confirmed
    - likely -> negative
    - if a duplicate key is encountered and the verification certificate indicates this status change, it will be recorded
  - This is the 3rd allowable change only (see original comment)
  - revisions must come from the same health authority ID 

Execution plan

- [ ] Change verification certificate claims to support v1.5 fields
- [ ] Change the exposure table to add the following fields: health_authority_id, report_type, days_since_symptom_onset, revised_report_type, revised_at (timestamp)
[...]

PR #718, opened 2020-07-14
Implement ExposureKey revisions

mikehelmick:
* Publish API changes
 * Transform still validates the incoming request
 * If any of the uploaded keys have previously been seen, join and figure out changes
 * perform upsert over the new keys

More work on #663

### DIFF
--- a/internal/publish/database/exposure.go
+++ b/internal/publish/database/exposure.go
@@ -27,6 +27,7 @@ import (
 	"github.com/google/exposure-notifications-server/internal/database"
 	"github.com/google/exposure-notifications-server/internal/logging"
 	"github.com/google/exposure-notifications-server/internal/publish/model"
+	verifyapi "github.com/google/exposure-notifications-server/pkg/api/v1alpha1"
 	"github.com/google/exposure-notifications-server/pkg/base64util"
 
 	pgx "github.com/jackc/pgx/v4"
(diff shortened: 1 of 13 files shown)


internal/publish/database/exposure.go
```diff
@@ -261,40 +264,106 @@ func (db *PublishDB) ReadExposures(ctx context.Context, b64keys []string) (map[s
 	return result, nil
 }
 
+func prepareInsertExposure(ctx context.Context, tx pgx.Tx) (string, error) {
```
sethvargo, line 267: Is there a reason we don't use

```
INSERT INTO Exposure ...
ON CONFLICT (exposure_key)
DO
  UPDATE EXPOSURE SET ...
```

? Seems like it would simplify this and ensure atomicity? Right now there's a race where the exposure could be created between the time you look it up and the time you try to create it.

sethvargo: Stackover for posterity: https://stackoverflow.com/questions/1109061/insert-on-duplicate-update-in-postgresql

internal/publish/model/exposure_model.go
```diff
@@ -31,6 +32,13 @@ import (
 	verifyapi "github.com/google/exposure-notifications-server/pkg/api/v1alpha1"
 )
 
+var (
+	ErrorExposureKeyMismatch     = errors.New("attempted to revise a key with a different key")
```
sethvargo, line 36: Nit: using `fmt.Errorf` will avoid needing to import `errors`

internal/publish/model/exposure_model.go
```diff
@@ -45,12 +53,64 @@ type Exposure struct {
...
 	RevisedReportType            *string
 	RevisedAt                    *time.Time
-	RevisedDaysSinceSymptomOnset *int
+	RevisedDaysSinceSymptomOnset *int32
+	RevisedTransmissionRisk      *int
+}
+
+// Revise updates the Revised fields of a key
+func (e *Exposure) Revise(in *Exposure) (bool, error) {
+	if e.ExposureKeyBase64() != in.ExposureKeyBase64() {
```
sethvargo, line 69: Is `ExposureKeyBase64` cached on calculation? I feel like we're calling it an awful lot and encoding isn't cheap.

internal/publish/database/exposure.go
mikehelmick: moved revision entirely in a DB transaction.

mikehelmick:
/retest
sadly , we still have that flaky test

PR #722, opened 2020-07-15
V1.5export

mikehelmick:
Add new TEK metadata fields to generated exports
Add revised keys to output
Work on #663

this is a re-open as gh wasn't picking up my merge fixes.

### DIFF
--- a/internal/export/exportfile.go
+++ b/internal/export/exportfile.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/google/exposure-notifications-server/internal/export/model"
 	publishmodel "github.com/google/exposure-notifications-server/internal/publish/model"
+	verifyapi "github.com/google/exposure-notifications-server/pkg/api/v1alpha1"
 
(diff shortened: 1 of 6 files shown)


internal/export/exportfile.go
```diff
@@ -131,30 +132,66 @@ func unmarshalContent(file *zip.File) (*export.TemporaryExposureKeyExport, error
 	return message, nil
 }
 
-func marshalContents(eb *model.ExportBatch, exposures []*publishmodel.Exposure, batchNum int32, batchSize int32, signers []*Signer) ([]byte, error) {
+func sortExposures(exposures []*publishmodel.Exposure) {
+	sort.Slice(exposures, func(i, j int) bool {
```
sethvargo, line 136: Should we use `SliceStable`? Can there even be duplicates?

mikehelmick: no, there can't be duplicates in this slice.

PR #730, opened 2020-07-15
Add Managed Identity auth to Azure Blob Storage

### LINKED ISSUE #715: Azure Storage credentials cannot be resolved from a secrets store - they are assumed to be directly in the environment
### TL;DR
The credentials for Azure Storage accounts are not able to be pulled from KeyVault (using the `secret://` prefix) because they are pulled directly from the environment. The code can be found [here](https://github.com/google/exposure-notifications-server/blob/main/internal/storage/azure_blobstore.go#L39-L67).

### Design

**Proposal**
The values could be more configurable if resolved using the [resolver](https://github.com/google/exposure-notifications-server/blob/ad7e1e00d20858f2f8a40a85c48143bdd9baf113/internal/setup/setup.go#L126) strategy that is currently being used for other secrets, like `DB_PASS`

**Alternatives considered**

Yes; none of them seem as idiomatic as the proposed design

**Resources**
https://github.com/google/exposure-notifications-server/blob/main/internal/storage/azure_blobstore.go#L39-L67

**Additional information**

No

jpflueger:
Resolves #715 

Adds support for using a managed identity to access a blob storage container on Azure. 

Notes:
- Omitting the `AZURE_STORAGE_ACCOUNT_KEY` will result in the storage component using the managed identity endpoint to obtain an access token that can be used to access blob storage containers.
- In order to test, the code must be running on an Azure VM, VMSS or AKS with a user-assigned identity configured for the machine and/or pod. 

/cc @nmiodice @andrebriggs

### DIFF
--- a/internal/storage/azure_blobstore.go
+++ b/internal/storage/azure_blobstore.go
@@ -21,8 +21,10 @@ import (
 	"io"
 	"net/url"
 	"os"
+	"time"
 
 	"github.com/Azure/azure-storage-blob-go/azblob"
+	"github.com/Azure/go-autorest/autorest/adal"
 )
 
 // Compile-time check to verify implements interface.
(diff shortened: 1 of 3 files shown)


internal/storage/azure_blobstore.go
```diff
@@ -34,6 +36,42 @@ type AzureBlobstore struct {
...
+	spt, err := adal.NewServicePrincipalTokenFromMSI(msiEndpoint, blobstoreURL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get service principal token from msi %v: %v", msiEndpoint, err)
+	}
+
+	var tokenRefresher azblob.TokenRefresher
+	tokenRefresher = func(credential azblob.TokenCredential) time.Duration {
+		err := spt.Refresh()
+		if err != nil {
+			panic(err)
```
sethvargo, line 62: I'm not a huge fan of panicing here. Can we return an error or maybe just log instead?

internal/storage/azure_blobstore.go
```diff
@@ -34,6 +36,42 @@ type AzureBlobstore struct {
...
+		}
+
+		token := spt.Token()
+		credential.SetToken(token.AccessToken)
+
+		exp := token.Expires().Sub(time.Now().Add(2 * time.Minute))
```
sethvargo, line 68: UTC time please

internal/storage/azure_blobstore.go
jpflueger: I'm pretty new to Go so I wasn't sure if this was a good thing to do or not. @nmiodice pointed me to the documentation for this package [here](https://godoc.org/github.com/Azure/azure-storage-blob-go/azblob). 

>  Your tokenRefresher function must return a time.Duration indicating how long the TokenCredential object should wait before calling your tokenRefresher function again. **If your tokenRefresher callback fails to refresh the token, you can return a duration of 0 to stop your TokenCredential object from ever invoking tokenRefresher again. Also, oen way to deal with failing to refresh a token is to cancel a context.Context object used by requests that have the TokenCredential object in their pipeline.**

I changed this to log the error and return a duration of 0 which will prevent the token from refreshing.

sethvargo: Welcome to Go 😄. It's a fun language I promise

sethvargo:
@jpflueger can you do the CLA thing?

jpflueger:
@googlebot I signed it!

PR #731, opened 2020-07-16
Don't stale enhancements

sethvargo:
/cc @mikehelmick

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -21,8 +21,8 @@ jobs:
           This Pull Request is stale because it has been open for 14 days with
           no activity. It will automatically close after 7 more days of
           inactivity.
-        stale-pr-label: 'kind/stale,kind/enhancement'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           This Pull Request is stale because it has been open for 14 days with
           no activity. It will automatically close after 7 more days of
           inactivity.
-        stale-pr-label: 'kind/stale'
+        stale-pr-label: 'kind/stale,kind/enhancement'
```
mikehelmick, line 24: next line?

sethvargo: No, it wants a CSV

mikehelmick: actually, looking closer, shouldn't this go on line 18?

adding kind/enhancement to the exempt list. Not to the labels to add on to a stale PR

sethvargo: Ah yup

PR #733, opened 2020-07-16
Rewrite integration test to incorporate retries

sethvargo:
Hopefully this fixes the sometimes flakey tests


sethvargo:
:|

internal/integration/integration_test.go
```diff
@@ -0,0 +1,392 @@
...
+
+func TestIntegration(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+	env, client, db := NewTestServer(t, 2*time.Second)
+
+	// Set query criteria (used throughout)
+	criteria := publishdb.IterateExposuresCriteria{
+		OnlyLocalProvenance: false,
```
mikehelmick, line 52: can add a TODO / file an issue- this only gets non-revised keys.
we'll also want to add the ability to revise keys in the test and iterate over the revised keys

internal/integration/integration_test.go
```diff
@@ -0,0 +1,392 @@
...
+
+			return retry.RetryableError(fmt.Errorf("export file still exists"))
+		}
+
+		return nil
+	})
+}
+
+// getExposures finds the exposures that match the given criteria.
+func getExposures(db *database.DB, criteria publishdb.IterateExposuresCriteria) ([]*publishmodel.Exposure, error) {
```
mikehelmick, line 333: I recently added a function in publishdb to read back exposures by key - so if we know what we're looking for we _could_ read it directly

mikehelmick:
> :|

Prow only supports 1️⃣ command at a ⌛

PR #736, opened 2020-07-17
Integration tests use real GCS

chaodaiG:
- Switch between using real GCS or local storage with env vars
- Configurable bucket name and file name root, so that prow jobs can test in the same bucket

/cc @sethvargo @stati0n

### DIFF
--- a/internal/integration/helpers.go
+++ b/internal/integration/helpers.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"net/http"
 	"os"
-	"strconv"
 	"testing"
 	"time"
 
@@ -109,16 +108,13 @@ func testServer(tb testing.TB) (*serverenv.ServerEnv, *http.Client) {
 		tb.Fatal(err)
 	}
 
-	var bs storage.Blobstore
 	if FileNameRoot, err = randomString(); err != nil {
 		tb.Fatal(err)
 	}
-	skipGcsTest, _ := strconv.ParseBool(os.Getenv("SKIP_GOOGLE_CLOUD_STORAGE_TESTS"))
-	if skipGcsTest || testing.Short() || os.Getenv("GOOGLE_CLOUD_BUCKET") == "" {
-		bs, err = storage.NewMemory(ctx)
-	} else {
+	bs, err := storage.NewMemory(ctx)
+	if v := os.Getenv("GOOGLE_CLOUD_BUCKET"); v != "" && !testing.Short() {
 		ExportDir = os.Getenv("GOOGLE_CLOUD_BUCKET")
-		bs, err = storage.BlobstoreFor(ctx, storage.BlobstoreTypeGoogleCloudStorage)
+		bs, err = storage.NewGoogleCloudStorage(ctx)
 	}
 	if err != nil {
 		tb.Fatal(err)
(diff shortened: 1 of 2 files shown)


internal/integration/helpers.go
```diff
@@ -37,10 +42,27 @@ import (
...
 )
 
-const (
-	exportDir = "my-bucket"
+var (
+	exportDir    = "my-bucket"
+	fileNameRoot = "/"
 )
 
+func GetExportDir() string {
```
sethvargo, line 50: I don't think you need to make getters or setters for this. Once it's a variable, it's globally accessible anyway to all the tests.

internal/integration/helpers.go
```diff
@@ -102,7 +125,20 @@ func testServer(tb testing.TB) (*serverenv.ServerEnv, *http.Client) {
 		tb.Fatal(err)
 	}
 
-	bs, err := storage.NewMemory(ctx)
+	var bs storage.Blobstore
+	var b [512]byte
```
sethvargo, line 129: Can you make the random generation a helper please?

internal/integration/helpers.go
chaodaiG: sg, done

internal/integration/helpers.go
chaodaiG: done

internal/integration/helpers.go
```diff
@@ -102,7 +109,17 @@ func testServer(tb testing.TB) (*serverenv.ServerEnv, *http.Client) {
...
+	var bs storage.Blobstore
+	if FileNameRoot, err = randomString(); err != nil {
+		tb.Fatal(err)
+	}
+	skipGcsTest, _ := strconv.ParseBool(os.Getenv("SKIP_GOOGLE_CLOUD_STORAGE_TESTS"))
+	if skipGcsTest || testing.Short() || os.Getenv("GOOGLE_CLOUD_BUCKET") == "" {
+		bs, err = storage.NewMemory(ctx)
+	} else {
+		ExportDir = os.Getenv("GOOGLE_CLOUD_BUCKET")
+		bs, err = storage.BlobstoreFor(ctx, storage.BlobstoreTypeGoogleCloudStorage)
```
sethvargo, line 121: Use https://github.com/google/exposure-notifications-server/blob/c3b76f3bc0cfbaf54a86f5c9412543bccc07cf19/internal/storage/google_cloud_storage.go#L38 instead

internal/integration/helpers.go
```diff
@@ -102,7 +109,17 @@ func testServer(tb testing.TB) (*serverenv.ServerEnv, *http.Client) {
 		tb.Fatal(err)
 	}
 
-	bs, err := storage.NewMemory(ctx)
```
sethvargo, line 105: Go-style wise, it's probably better to keep this as is, then overwrite `bs` if the envvar exists:

```golang
bs, err := storage.NewMemory()
if err != nil {
  tb.Fatal(err)
}

if v := os.Getenv("GOOGLE_CLOUD_BUCKET"; v != "" && !testing.Short() {
  bs, err = storage.NewGoogleCloudStorage(ctx)
  if err != nil {
    tb.Fatal(err)
  }
}
```

internal/integration/helpers.go
chaodaiG: good call, done

internal/integration/helpers.go
chaodaiG: done

Issue #734, opened 2020-07-16
Implement and accept "revocation/revision" certificates


mikehelmick:
### TL;DR

As part of the [verification flow](https://developers.google.com/android/exposure-notifications/verification-system)

When keys are accepted, an encrypted token needs to be returned as the revocation (or revision) certificate.

This certificate needs to be passed back in order to revise the keys at a later date. This ensures that a key can't be revised by a different device.

/kind enhancement
/assign
/priority important-soon

mikehelmick:
This is done - I Want to add one more test before I close it out.

mikehelmick:
last thing to add - filtering out "too old" keys

mikehelmick:
and.. that is already done.

/close

PR #740, opened 2020-07-17
Add encrypt/decrypt operations to key management interface

mikehelmick:
* add implementation in all supported key management services
  * AEAD is missing from azure keyvault, have reached out for help
* In support of #734

### DIFF
--- a/pkg/api/v1alpha1/exposure_types.go
+++ b/pkg/api/v1alpha1/exposure_types.go
@@ -56,6 +56,10 @@ const (
 //  - Does not have to be within range of any of the provided keys (i.e. future
 //    key uploads)
 //
+// revisionToken: An opaque string that must be passed in-tact from on additional
+//   publish requests from the same device, there the same TEKs may be published
+//   again.
+//
 // Padding: random base64 encoded data to obscure the request size.
 //
 // The following fields are deprecated, but accepted for backwards-compatibility:
@@ -68,11 +72,22 @@ type Publish struct {
 	VerificationPayload  string        `json:"verificationPayload"`
 	HMACKey              string        `json:"hmackey"`
 	SymptomOnsetInterval int32         `json:"symptomOnsetInterval"`
+	RevisionToken        string        `json:"revisionToken"`
 
 	Padding string `json:"padding"`
 
 	Platform                  string `json:"platform"`                  // DEPRECATED
 	DeviceVerificationPayload string `json:"deviceVerificationPayload"` // DEPRECATED
+}
+
+// PublishResponse is sent back to the client on a publish request.
+// If successful, the revisionToken indicates an opaque string that must be
+// passed back if the same devices wishes to publish TEKs again.
+//
+// On error, the error field will contain the error details.
+type PublishResponse struct {
+	RevisionToken string `json:"revisionToken"`
+	Error         string `json:"error"`
 }
 
 // ExposureKey is the 16 byte key, the start time of the key and the
--- a/pkg/keys/aws_kms.go
+++ b/pkg/keys/aws_kms.go
@@ -49,3 +49,35 @@ func NewAWSKMS(ctx context.Context) (KeyManager, error) {
 func (s *AWSKMS) NewSigner(ctx context.Context, keyID string) (crypto.Signer, error) {
 	return awskms.NewSigner(ctx, s.svc, keyID)
 }
+
+func buildAAD(aad string) map[string]*string {
+	m := make(map[string]*string)
+	m["aad"] = &aad
+	return m
+}
+
+func (s *AWSKMS) Encrypt(ctx context.Context, keyID string, plaintext []byte, aad string) ([]byte, error) {
+	input := kms.EncryptInput{
+		KeyId:             &keyID,
+		EncryptionContext: buildAAD(aad),
+		Plaintext:         plaintext,
+	}
+	output, err := s.svc.EncryptWithContext(ctx, &input)
+	if err != nil {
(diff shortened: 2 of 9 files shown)


pkg/api/v1alpha1/exposure_types.go
```diff
@@ -68,13 +72,24 @@ type Publish struct {
 	VerificationPayload  string        `json:"verificationPayload"`
 	HMACKey              string        `json:"hmackey"`
 	SymptomOnsetInterval int32         `json:"symptomOnsetInterval"`
+	RevisionToken        string        `json:"revisionToken"`
```
mikehelmick, line 75: not implemented yet, showing API shape

pkg/keys/aws_kms.go
```diff
@@ -49,3 +49,35 @@ func NewAWSKMS(ctx context.Context) (KeyManager, error) {
...
+func buildAAD(aad string) map[string]*string {
+	m := make(map[string]*string)
+	m["aad"] = &aad
+	return m
+}
+
+func (s *AWSKMS) Encrypt(ctx context.Context, keyID string, plaintext []byte, aad string) ([]byte, error) {
+	input := kms.EncryptInput{
+		KeyId:             &keyID,
+		EncryptionContext: buildAAD(aad),
```
sethvargo, line 62: I think the AWS-ey Go way to do this is:

```go
map[string]*string{
  "aad": aws.String(aad),
}
```

pkg/keys/aws_kms.go
```diff
@@ -49,3 +49,35 @@ func NewAWSKMS(ctx context.Context) (KeyManager, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to encrypt: %w", err)
+	}
+	return output.CiphertextBlob, nil
+}
+
+func (s *AWSKMS) Decrypt(ctx context.Context, keyID string, ciphertext []byte, aad string) ([]byte, error) {
+	input := kms.DecryptInput{
+		KeyId:             &keyID,
+		EncryptionContext: buildAAD(aad),
```
sethvargo, line 75: Same thing

pkg/keys/hashicorp_vault.go
```diff
@@ -54,20 +54,97 @@ func NewHashiCorpVault(ctx context.Context) (KeyManager, error) {
...
+
+func (v *HashiCorpVault) Encrypt(ctx context.Context, keyID string, plaintext []byte, aad string) ([]byte, error) {
+	kid, err := NewHCValueKeyID(keyID)
+	if err != nil {
+		return nil, err
+	}
+	pth := fmt.Sprintf("transit/encrypt/%s", kid.Name)
+	params := map[string]interface{}{
+		"plaintext": base64.StdEncoding.EncodeToString(plaintext),
+		"context":   base64.StdEncoding.EncodeToString([]byte(aad)),
```
sethvargo, line 94: We should document that this key needs to be created with `derived=true`

pkg/keys/hashicorp_vault.go
```diff
@@ -54,20 +54,97 @@ func NewHashiCorpVault(ctx context.Context) (KeyManager, error) {
...
+		"type":      "aes256-gcm96",
+	}
+
+	result, err := v.client.Logical().Write(pth, params)
+	if err != nil {
+		return nil, fmt.Errorf("unable to encrypt: %w", err)
 	}
+
+	ciphertext, ok := result.Data["ciphertext"]
```
sethvargo, line 103: It's generally recommended that we store the entire ciphertext, including the `vault:vx:` prefix

pkg/keys/hashicorp_vault.go
```diff
@@ -54,20 +54,97 @@ func NewHashiCorpVault(ctx context.Context) (KeyManager, error) {
...
+}
+
+func (v *HashiCorpVault) Decrypt(ctx context.Context, keyID string, ciphertext []byte, aad string) ([]byte, error) {
+	kid, err := NewHCValueKeyID(keyID)
+	if err != nil {
+		return nil, err
+	}
+	pth := fmt.Sprintf("transit/decrypt/%s", kid.Name)
+	params := map[string]interface{}{
+		"ciphertext": fmt.Sprintf("vault:v1:%s", base64.StdEncoding.EncodeToString(ciphertext)),
```
sethvargo, line 128: It won't always be v1 - that's why we should store the full blob that's returned.

pkg/keys/keys.go
```diff
@@ -31,6 +31,9 @@ import (
 // KeyManager implementations must be able to return a crypto.Signer.
 type KeyManager interface {
 	NewSigner(ctx context.Context, keyID string) (crypto.Signer, error)
+
+	Encrypt(ctx context.Context, keyID string, plaintext []byte, aad string) ([]byte, error)
```
sethvargo, line 35: Can we doc these? Is it valid for aad to be ""?

sethvargo: Also, should aad be []byte?

pkg/keys/hashicorp_vault.go
mikehelmick: added comments on the HashiCorpVault struct.

pkg/keys/hashicorp_vault.go
mikehelmick: thanks for the tip.

pkg/keys/keys.go
mikehelmick: went back and forth on it - agree []byte makes more sense.

practically, we'll probably take this in as a base64 env var

mikehelmick:
rebased on main

PR #754, opened 2020-07-20
Creating load test workload using /export endpoint

chaodaiG:
Creating this as a draft PR only for showing how much time is needed for creating exports with `/export` endpoint.

Due to design, it's not possible to generate exports using go routines, as it would generate non-deterministic number of keys in each batch of exports, also there could be export calls results in no exports. 

This PR demonstrates how much time it takes for generating 100 exports, which is roughly 210 seconds on my development machine

### DIFF
--- a/internal/performance/cleanup_test.go
+++ b/internal/performance/cleanup_test.go
@@ -27,11 +27,11 @@ import (
 
 	exportapi "github.com/google/exposure-notifications-server/internal/export"
 	exportdb "github.com/google/exposure-notifications-server/internal/export/database"
-	exportmodel "github.com/google/exposure-notifications-server/internal/export/model"
 	"github.com/google/exposure-notifications-server/internal/integration"
 	"github.com/google/exposure-notifications-server/internal/storage"
 	"github.com/google/exposure-notifications-server/internal/util"
 	verifyapi "github.com/google/exposure-notifications-server/pkg/api/v1alpha1"
+	pgx "github.com/jackc/pgx/v4"
 	"github.com/sethvargo/go-retry"
 )
 
@@ -41,7 +41,7 @@ func TestBatchCleanupBenchmark(t *testing.T) {
 		keysPerBatch = 3
 		// If consider batching every 10 seconds, 10000 batches is 100000
 		// seconds ~= 27 hours worth of exports
-		batchCount = 10000
+		batchCount = 100
 	)
 	var (
 		err          error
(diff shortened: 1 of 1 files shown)


sethvargo:
> Due to design, it's not possible to generate exports using go routines, as it would generate non-deterministic number of keys in each batch of exports...

Can you share more about this?

chaodaiG:
> > Due to design, it's not possible to generate exports using go routines, as it would generate non-deterministic number of keys in each batch of exports...
> 
> Can you share more about this?

The next level of optimization I could imagine, is building a pipeline of go routines, doing `publish`, `mark publish old`, `export`, and `mark export old`.
During my experiments, I have noticed that the `/export` endpoint scoop up all unexported keys and export them. So if we use the pipeline I mentioned above, it's pretty hard to control how many keys are there when `/export` is invoked, if the number is 0 then there is no export

sethvargo:
We bound exports between a start and stop time. I wonder if we could leverage that to make the exports more deterministic? @mikehelmick @gurayAlsac any ideas?

mikehelmick:
if we want to limit the number of keys in the export, we can assign specific created_at/revised_at timestamps to be within the export batch bounds.

gurayAlsac:
Is the idea that you want to slowly increase the number of keys in the export and gradually ramp up the load? Or fix the number of keys in the export so that latency measurements can be meaningfully compared over time?

Does the export load test need to exercise the publish path? I.e., are we trying to get a sense of throughput of the system overall, or just for the export server itself? If the latter, you might do a direct insert into the db within the appropriate time ranges to precisely control the number of keys.

chaodaiG:
> if we want to limit the number of keys in the export, we can assign specific created_at/revised_at timestamps to be within the export batch bounds.

I didn't see `revised_at` being used for exports. Did use `created_at`(which is what this PR does as well), but the exported keys in each batch are not consistent either

chaodaiG:
> Is the idea that you want to slowly increase the number of keys in the export and gradually ramp up the load? Or fix the number of keys in the export so that latency measurements can be meaningfully compared over time?
> 
> Does the export load test need to exercise the publish path? I.e., are we trying to get a sense of throughput of the system overall, or just for the export server itself? If the latter, you might do a direct insert into the db within the appropriate time ranges to precisely control the number of keys.

The idea is to generate exports in reasonable amount of time, i.e. minutes instead of hours. The implementation of this test in master branch is already inserting directly into db, and it worked pretty well. The purpose of this PR is trying to use standard endpoints for generating exports instead

PR #761, opened 2020-07-21
Internal structure for revision token

mikehelmick:
* Add proto definition for revision token
* Add ability to encrypt/decrype the revision token data

Towards #734

/cc @tlepoint


internal/revision/revision.go
```diff
@@ -0,0 +1,137 @@
...
+			IntervalCount:        k.IntervalCount,
+		}
+		copy(pbKey.TemporaryExposureKey, k.ExposureKey)
+		tokenData.RevisableKeys = append(tokenData.RevisableKeys, &pbKey)
+	}
+	return &tokenData
+}
+
+// MakeRevisionToken turns the TEK data from a given publish request
+// into an encrypted protocol buffer revision token.
```
sethvargo, line 54: Maybe doc this uses envelope encryption?

PR #764, opened 2020-07-22
Load test for exporting keys at scale

chaodaiG:
- Add a new test replacing the old load test for cleaning up, this new test measures performance for both export and cleanup
- Implemented a slightly better way of generating exports
- There is still a workaround of generating exports, which is also hard to be done using `/publish` endpoint for scaled workloads. So it's still blocked from being merged with integration test

The test runs 660 seconds on my development machine, took 2.2 minutes export and 265 milliseconds cleaning up.

The mako set up will be added in a separate PR


internal/performance/export_test.go
```diff
@@ -0,0 +1,259 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		if updated != keysPerPublish {
+			t.Fatalf("Want updated: %d, got %d", keysPerPublish, updated)
+		}
+	}
+
+	// Ensure all keys published
+	integration.Eventually(t, 30, func() error {
```
mikehelmick, line 146: is this step needed? doesn't finding all the keys in and export tell us that?

chaodaiG: Good point, it's not needed. Deleted

PR #765, opened 2020-07-23
Managed revision keys

mikehelmick:
* add "in memory" implementation of the keys.KeyManager interface
  * used in these tests, but will be useful in other places as well
* create new RevisionKeys database table
* revision keys are generated AES keys that are encrypted using the configured KMS before storing in the databse
* revision tokens are encrypted/decrypted using envelope encryption based on these keys
* modify revision token to use a token manager
  * allowed keys are cached
  * the keys are only unwrapped when necessary
  * key rotation is enabled by minting a new revision key
  * a key can be destroyed when it should no longer be accepeted

Part of #734 - just a bit more to go after this.


internal/pb/revision.proto
```diff
@@ -17,7 +17,7 @@ syntax = "proto3";
 option go_package = "github.com/google/exposure-notifications-server/internal/pb;pb";
 
 message RevisionToken {
-    string kid = 1;
+    int64 kid = 1;
```
sethvargo, line 20: Hmm - so this change restricts us to integer-based key ids - is that okay? Do we ever expect UUID or other identifiers?

internal/revision/database/revision.go
```diff
@@ -0,0 +1,292 @@
...
+	CreatedAt     time.Time
+	Allowed       bool
+
+	// The unwrapped cipher.
+	DEK []byte
+}
+
+// KeyIDString returns the keyID as a string that can be used in the encoded revision tokens.
+func (r *RevisionKey) KeyIDString() string {
+	return fmt.Sprintf("%d", r.KeyID)
```
sethvargo, line 46: If we're already printing it as a string, why not keep it as a string in the proto?

Also strconv.Itoa(r.KeyID) is much faster.

internal/revision/database/revision.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+// RevisionDB wraps a database connection and provides functions for interacting with revision keys.
+type RevisionDB struct {
+	db           *database.DB
+	wrapperKeyID string
```
sethvargo, line 52: Should we consider making a `KeyParams` type that is the KeyID, AAD, and other future things?

internal/revision/database/revision.go
```diff
@@ -0,0 +1,292 @@
...
+		wrapperAAD:   wrapperAAD,
+		keyManager:   kms,
+	}, nil
+}
+
+// DestroyKey zeros out the wrapped key and marks the key as allowed=false.
+func (rdb *RevisionDB) DestroyKey(ctx context.Context, keyID int64) error {
+	logger := logging.FromContext(ctx)
+	logger.Warnf("destroying key material for revision key ID %v", keyID)
+	zero := []byte{0}
```
sethvargo, line 80: Constant?

migrations/000033_AddRevisionTokenKeys.up.sql
```diff
@@ -0,0 +1,27 @@
...
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+CREATE TABLE RevisionKeys (
+    kid SERIAL PRIMARY KEY,
+    aad bytea NOT NULL,
```
sethvargo, line 19: If we're zeroing out these, why not make them nullable and avoid the mysterious zero-byte sentinel value?

internal/revision/database/revision.go
```diff
@@ -0,0 +1,292 @@
...
+		for rows.Next() {
+			if err := rows.Err(); err != nil {
+				return fmt.Errorf("failed to iterate: %w", err)
+			}
+			var id int64
+			if err := rows.Scan(&id); err != nil {
+				return err
+			}
+			keys[id] = struct{}{}
+			if effectiveID == 0 {
```
sethvargo, line 132: I'm not sure what's going on here? Can you add a comment explaining `effectiveID`?

internal/revision/database/revision.go
```diff
@@ -0,0 +1,292 @@
...
+	keys := make([]*RevisionKey, 0)
+	if err := rdb.db.InTx(ctx, pgx.ReadCommitted, func(tx pgx.Tx) error {
+		rows, err := tx.Query(ctx, `
+			SELECT
+				kid, aad, wrapped_cipher, created_at, allowed
+			FROM
+				revisionkeys
+			WHERE
+				allowed=$1
+			ORDER BY created_at DESC`, true)
```
sethvargo, line 159: This is probably worth commenting or highlighting somewhere - the []*RevisionKey will be ordered chronologically.

internal/revision/database/revision.go
```diff
@@ -0,0 +1,292 @@
...
+	}); err != nil {
+		return 0, nil, fmt.Errorf("unable to read keys: %w1", err)
+	}
+
+	unwrappedKeys := make([]*RevisionKey, 0, len(keys))
+	// Attempt to unwrap all of the keys
+	for _, wk := range keys {
+		unwrapped, err := rdb.keyManager.Decrypt(ctx, rdb.wrapperKeyID, wk.WrappedCipher, rdb.wrapperAAD)
+		if err != nil {
+			logger.Errorf("still allowed revision key that can't be unwrapped: kid: %v error: %v", wk.KeyID, err)
```
sethvargo, line 190: I think we should return immediately if any decryption fails.

internal/revision/database/revision.go
```diff
@@ -0,0 +1,292 @@
...
+		if err != nil {
+			return err
+		}
+		defer rows.Close()
+
+		if rows.Next() {
+			if err := rows.Err(); err != nil {
+				return fmt.Errorf("failed to iterate: %w", err)
+			}
+			r := RevisionKey{}
```
sethvargo, line 224: Style nit: `var r RevisionKey`

internal/revision/database/revision.go
```diff
@@ -0,0 +1,292 @@
...
+		return nil
+	}); err != nil {
+		return nil, fmt.Errorf("unable to get revision key: %w", err)
+	}
+	if revKey == nil {
+		return nil, fmt.Errorf("no effective key found")
+	}
+
+	// Unwrap the DEK w/ the KeyManager.
+	unwrapped, err := rdb.keyManager.Decrypt(ctx, rdb.wrapperKeyID, revKey.WrappedCipher, rdb.wrapperAAD)
```
sethvargo, line 239: I'm seeing this pattern a few times now, but keyID and ADD are properties of rdb. You could make this a helper for readability:

```go
func (rdb *RevisionDB) decrypt(ctx context.Context, ciphertext []byte) ([]byte, error) {
 // ...
}
```

internal/revision/revision.go
```diff
@@ -17,18 +17,132 @@
...
+
+// New creates a new TokenManager that uses a database handle to manage a cache
+// of allowed revision keys.
+func New(ctx context.Context, db *database.RevisionDB, cacheDuration time.Duration) (*TokenManager, error) {
+	now := time.Now()
+	tm := &TokenManager{
+		db:                db,
+		allowed:           make(map[int64]*database.RevisionKey),
+		cacheDuration:     cacheDuration,
+		cacheRefreshAfter: now.Add(-2 * cacheDuration),
```
sethvargo, line 57: If cacheDuration is (2 << 64)-1, I think this overflows?

internal/revision/revision.go
```diff
@@ -17,18 +17,132 @@
...
+
+	reload, err := tm.isReloadNeeded(ctx)
+	if err != nil {
+		return fmt.Errorf("unable to read revsion keys: %w", err)
+	}
+	if !reload {
+		return nil
+	}
+
+	// At this point reload is needed. Trash the
```
sethvargo, line 87: Finish comment

internal/revision/revision.go
```diff
@@ -17,18 +17,132 @@
...
 
+// TokenManager is responsible for creating and unlocking revision tokens.
+type TokenManager struct {
+	db *database.RevisionDB
+
+	mu        sync.RWMutex
+	allowed   map[int64]*database.RevisionKey
+	effective *database.RevisionKey
+
+	cacheDuration     time.Duration
```
sethvargo, line 45: Is our existing cache implementation not sufficient?

pkg/keys/in_memory.go
```diff
@@ -0,0 +1,157 @@
...
+	"crypto/cipher"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"fmt"
+	"io"
+	"sync"
+)
+
+// InMemoryKMS is useful for testing. Do NOT use in a running system as all
```
sethvargo, line 30: Nit: we should probably just name it `InMemory` since it's `keys.InMemory`

sethvargo, line 30: We should also consider dropping the Noop one and encourage using this instead.

internal/revision/database/revision.go
```diff
@@ -29,6 +39,10 @@ import (
 	"github.com/jackc/pgx/v4"
 )
 
+var (
+	zero = []byte{}
```
sethvargo, line 43: Do we still need this is the field is nullable now?

mikehelmick: yeah, made it nil.

internal/revision/database/revision.go
sethvargo: Thanks!

PR #769, opened 2020-07-23
Setup mako for uploading load test metrics to server

chaodaiG:
This PR injects performance test metrics to https://mako.dev, so that the results can be visualized there as well. 

This had been tested locally and performance metrics was uploaded to mako server, which could be visualized at https://mako.dev/benchmark?benchmark_key=5698752440434688. (Note: for faster turnaround time, the data for this was from 1000 publishes instead of 100000 in real test)

/assign @sethvargo @mikehelmick 
/cc @stati0n

### DIFF
--- a/internal/performance/helper.go
+++ b/internal/performance/helper.go
@@ -1,3 +1,5 @@
+// +build performance
+
 // Copyright 2020 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -23,6 +25,10 @@ import (
 	"testing"
 	"time"
 
+	// Warning: package github.com/golang/protobuf/proto is deprecated: Use the
+	// "google.golang.org/protobuf/proto" package instead.  (SA1019).
+	// Use deprecated package as mako isn't compatible with
+	// github.com/golang/protobuf/proto yet
 	"github.com/golang/protobuf/proto"
 	"github.com/google/mako/go/quickstore"
 
--- a/internal/performance/helper.go
+++ b/internal/performance/helper.go
@@ -42,10 +42,7 @@ const (
 
 // setup sets up client used for performance test
 func setup(tb testing.TB) (*quickstore.Quickstore, func(context.Context)) {
-	var (
-		microservice string
-	)
-
+	var microservice string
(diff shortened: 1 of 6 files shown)


internal/performance/helper.go
```diff
@@ -0,0 +1,101 @@
...
+	mpb "github.com/google/mako/spec/proto/mako_go_proto"
+)
+
+const (
+	benchmarkConfigFile = "export_test_benchmark.config"
+)
+
+// setup sets up client used for performance test
+func setup(tb testing.TB) (*quickstore.Quickstore, func(context.Context)) {
+	var (
```
mikehelmick, line 45: just
```
var microservice string
```

chaodaiG: done

internal/performance/helper.go
```diff
@@ -0,0 +1,101 @@
...
+		microservice string
+	)
+
+	benchmarkConfig := &mpb.BenchmarkInfo{}
+
+	{
+		p := os.Getenv("MAKO_PORT")
```
mikehelmick, line 49: env var configuration should use a config struct
https://github.com/google/exposure-notifications-server/blob/main/internal/publish/config.go#L39

and then we use a library to parse the environment

internal/performance/helper.go
```diff
@@ -0,0 +1,101 @@
...
+	benchmarkConfigFile = "export_test_benchmark.config"
+)
+
+// setup sets up client used for performance test
+func setup(tb testing.TB) (*quickstore.Quickstore, func(context.Context)) {
+	var (
+		microservice string
+	)
+
+	benchmarkConfig := &mpb.BenchmarkInfo{}
```
mikehelmick, line 46: declare this closer to where it us used

internal/performance/helper.go
chaodaiG: sg, done

internal/performance/helper.go
chaodaiG: After using envconfig the closures don't seem to be necessary, removed closures and reordered a little bit

PR #771, opened 2020-07-25
Hook up revision tokens

### LINKED ISSUE #767: delete keys.Noop
remove the keys.Noop KeyManager implementation and replace all uses w/ the new InMemory implementation.

Can also use this to write additional tests that actually sign and verify exports in memory.

/kind enhancement

mikehelmick:
* revsion tokens are always generated
 * Added internal padding to a minimum size to prevent deriving # of keys from token size.
* if present, and if enabled for the app, they are used to restrict which keys can be revised
* publish test verifies contents of the token

More work on #734
Fixes #767


mikehelmick:
/hold
until I get the terraform ready - this changes dependencies on the exposure server and we need to generate byte on setup.

mikehelmick:
/unhold

@sethvargo - please double check my terraform
/shrug

mikehelmick:
terraform deploy verified on fresh project

jixiangox:
Deploy partial components to different regions is not OK combined with standard db instance changed

mikehelmick:
> Deploy partial components to different regions is not OK combined with standard db instance changed

Hello - is that in relation to this PR or is there an issue we need to file?

Issue #773, opened 2020-07-27
Better formulation of chaff requests


mikehelmick:
- should not be a dedicated API
- allow for client to explicitly signal a chaff request
- utilize http://github.com/mikehelmick/go-chaff

PR #775, opened 2020-07-27
Support chaff requests for exposures

sethvargo:
Fixes GH-773


mikehelmick:
looks like you need to run go mod verify locally and repush

PR #776, opened 2020-07-29
Fix UI-model connections

zssz:


mikehelmick:
/lgtm
/approve

Thanks for catching that!

PR #788, opened 2020-07-30
Handle slices of secrets

sethvargo:
/assign @mikehelmick


mikehelmick:
check test failures

sethvargo:
🐼

sethvargo:
@mikehelmick PTAnotherL

pkg/secrets/in_memory.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"context"
+	"fmt"
+)
+
+// Compile-time check to verify implements interface.
+var _ SecretManager = (*InMemory)(nil)
+
+// InMemory is an in-memory secret manager, primarily used for testing.
+type InMemory struct {
```
mikehelmick, line 26: do you want to add the ability to add new secrets after create?
I did this for the InMemory KMS

pkg/secrets/config.go
```diff
@@ -24,9 +24,12 @@ type SecretManagerType string
...
 	SecretManagerTypeAWSSecretsManager    SecretManagerType = "AWS_SECRETS_MANAGER"
 	SecretManagerTypeAzureKeyVault        SecretManagerType = "AZURE_KEY_VAULT"
-	SecretManagerTypeGoogleSecretManager  SecretManagerType = "GOOGLE_SECRET_MANAGER"
 	SecretManagerTypeGoogleHashiCorpVault SecretManagerType = "HASHICORP_VAULT"
-	SecretManagerTypeNoop                 SecretManagerType = "NOOP"
+	SecretManagerTypeGoogleSecretManager  SecretManagerType = "GOOGLE_SECRET_MANAGER"
+	SecretManagerTypeInMemory             SecretManagerType = "IN_MEMORY"
+
+	// SecretManagerTypeNoop is deprecated. Use SecretManagerTypeInMemory.
+	SecretManagerTypeNoop SecretManagerType = "NOOP"
```
mikehelmick, line 32: I think it would be best to do a hard drop here since we made it impossible to create one.

then anyone using this as a lib would get a compile error instead of a runtime error

pkg/secrets/in_memory.go
sethvargo: I did that originally, but it requires a type assertion since we return an interface. I added the  `NewInMemoryFromMap` so you could create one with values preset. I'm open to allowing it to mutate state, just didn't see a use case for it.

pkg/secrets/config.go
sethvargo: Done.

sethvargo:
@mikehelmick again plz

sethvargo:
Looks like a module proxy error?

PR #789, opened 2020-07-30
Allow integration test against SQL server from GCP

chaodaiG:
- Create a presubmit setup script for setting up db for test
- By default, integration test still uses local DB as of today
- Use GCP SQL server when an environment variable is set
- Allow reuse existing instance, user, and database
- Teardown can be optionally enabled by passing environment variables

A couple of other options had been thought about:

1. Provision DB using terraform. The pros is no duplicated logic, and the cons is not quite flexible on reusing existing instance and databases
1. Use Golang setup/teardown for provisioning. The cons is the same as current approach, with many more additional work

Feedbacks and comments are welcome, would be super happy to hear about better solutions

### DIFF
--- a/internal/integration/helpers.go
+++ b/internal/integration/helpers.go
@@ -38,6 +38,7 @@ import (
 	"github.com/google/exposure-notifications-server/pkg/keys"
 	"github.com/google/exposure-notifications-server/pkg/secrets"
 	"github.com/google/exposure-notifications-server/pkg/server"
+	"github.com/sethvargo/go-envconfig"
 	"github.com/sethvargo/go-retry"
 
 	authorizedappmodel "github.com/google/exposure-notifications-server/internal/authorizedapp/model"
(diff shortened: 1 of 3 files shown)


sethvargo:
Just to be clear, this is a Postgres server, not a Microsoft SQL server, right?

sethvargo:
My $0.02 is that we should use the Terraform configs to stand up a dedicated test instance on-demand instead of the shell script.

chaodaiG:
That's correct, it's postgres server not sql, GCP names them SQL overall
```
DB_VERSION="${DB_VERSION:-POSTGRES_11}"
```

chaodaiG:
> My $0.02 is that we should use the Terraform configs to stand up a dedicated test instance on-demand instead of the shell script.

I would prefer this as well if it doesn't take more than 10 minutes to spin up the entire stack. There are also some other limitations such as org policy doesn't like `allUsers` such as https://github.com/google/exposure-notifications-server/blob/d61a3f3e91bef00ad5308bae26210fa0de5a21a8/terraform/storage.tf#L46

chaodaiG:
Close this PR for now, will try to deploy with terraform
/close

PR #790, opened 2020-07-30
Fix presubmit test bug to surface error message better

chaodaiG:
presubmit.sh script already has errexit set, so the script exits right away when the exit code of a command is non-zero. This makes checking exit code such as `$?` afterwards being noop, thus the error message would never get hit. Fixing this so that the error messages are surfaced

An example of this problem: https://oss-prow.knative.dev/view/gs/oss-prow/pr-logs/pull/google_exposure-notifications-server/789/pull-en-server-release-unit/1288924239646167040

### DIFF
--- a/scripts/presubmit.sh
+++ b/scripts/presubmit.sh
@@ -32,12 +32,11 @@ ${ROOT}/scripts/dev protoc
 
 
 echo "📚 Fetch dependencies"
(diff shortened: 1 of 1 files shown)


sethvargo:
Can you make the same changes on the verification server please?

PR #796, opened 2020-07-31
Add traveler bit

### LINKED ISSUE #792: Roaming: Implement a "traveler" bit
On publish and in the database, store a traveller bool next to the TEKs.

/kind enhancement
/kind good-first-issue

mikehelmick:
Indicates that this TEK has travelled beyond it's assigned region(s).
Add ability to include or exclusively get travellers in the exposure iterator.

Fixes #792

### DIFF
--- a/internal/publish/database/exposure.go
+++ b/internal/publish/database/exposure.go
@@ -51,12 +51,14 @@ func New(db *database.DB) *PublishDB {
 
 // IterateExposuresCriteria is criteria to iterate exposures.
 type IterateExposuresCriteria struct {
-	IncludeRegions []string
-	ExcludeRegions []string
-	SinceTimestamp time.Time
-	UntilTimestamp time.Time
-	LastCursor     string
-	RevisedKeys    bool // If true, only revised keys that match will be selected.
+	IncludeRegions   []string
+	IncludeTravelers bool // Include records in the IncludeRegions OR travallers
+	OnlyTravelers    bool // Only includes records marked as travelers.
+	ExcludeRegions   []string
+	SinceTimestamp   time.Time
+	UntilTimestamp   time.Time
+	LastCursor       string
+	RevisedKeys      bool // If true, only revised keys that match will be selected.
 
 	// OnlyLocalProvenance indicates that only exposures with LocalProvenance=true will be returned.
 	OnlyLocalProvenance bool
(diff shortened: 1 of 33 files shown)


flagxor:
lgtm

migrations/000035_AddTravelerBit.up.sql
```diff
@@ -0,0 +1,22 @@
...
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+-- Traveler column is not indexed because we always iterate over keys by their created_at
+-- or revised_at time which is indexed and do secondary filtering in the server.
+ALTER TABLE Exposure
+    ADD COLUMN traveler BOOL NOT NULL DEFAULT false;
```
sethvargo, line 20: Nit: it might be better to make this `is_traveler` to indicate its a boolean.

internal/publish/database/exposure_test.go
```diff
@@ -140,6 +143,7 @@ func TestExposures(t *testing.T) {
 			IntervalCount:   3,
 			CreatedAt:       batchTime.Add(3 * time.Hour),
 			Regions:         []string{"US"},
+			Traveler:        true,
```
sethvargo, line 146: Can we add another test where Travler is false?

internal/publish/database/exposure.go
```diff
@@ -154,8 +157,14 @@ func generateExposureQuery(criteria IterateExposuresCriteria) (string, []interfa
 	`
 
 	if len(criteria.IncludeRegions) == 1 {
-		args = append(args, criteria.IncludeRegions)
-		q += fmt.Sprintf(" AND (regions && $%d)", len(args)) // Operation "&&" means "array overlaps / intersects"
+		if criteria.IncludeTravelers {
+			args = append(args, criteria.IncludeRegions)
+			args = append(args, true)
+			q += fmt.Sprintf(" AND ((regions && $%d) OR traveler = $%d)", len(args)-1, len(args)) // Operation "&&" means "array overlaps / intersects"
```
sethvargo, line 163: I'm pretty confused about what's happening here in this SQL. Could we break this up or add a comment?

Also, I'm hesitant about passing the raw SQL string instead of binding these values. I understand they are lengths that we calculate, but it's still derived from user-supplied input.

internal/publish/database/exposure.go
```diff
@@ -189,7 +198,16 @@ func generateExposureQuery(criteria IterateExposuresCriteria) (string, []interfa
 		q += fmt.Sprintf(" AND local_provenance = $%d", len(args))
 	}
 
-	q += " ORDER BY created_at"
+	if criteria.OnlyTravelers {
+		args = append(args, true)
+		q += fmt.Sprintf(" AND traveler = $%d", len(args))
```
sethvargo, line 203: Isn't this a boolean? Why would it be equal to an int?

internal/publish/database/exposure.go
mikehelmick: We're not passing the raw SQL - it's setting values for bind variables and simultaneously building up the list.

internal/publish/database/exposure.go
mikehelmick: it's not, it putting in a bind for $XX where XX is the current length of args.
args itself is bound as query params

migrations/000035_AddTravelerBit.up.sql
mikehelmick: our other bool fields don't have that prefix, but doesn't matter much to me.

internal/publish/database/exposure.go
mikehelmick: added comment

migrations/000035_AddTravelerBit.up.sql
sethvargo: Ok nvm

Issue #800, opened 2020-08-01
Allow configuring custom domain mapping via Terraform


sethvargo:
https://www.terraform.io/docs/providers/google/r/cloud_run_domain_mapping.html

/assign

PR #803, opened 2020-08-01
Support custom domains

sethvargo:
Fixes GH-800

### DIFF
--- a/terraform/vars.tf
+++ b/terraform/vars.tf
@@ -139,7 +139,7 @@ variable "federationout_custom_domain" {
   type    = string
   default = ""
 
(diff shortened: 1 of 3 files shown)


terraform/vars.tf
```diff
@@ -128,6 +128,20 @@ variable "service_environment" {
   description = "Per-service environment overrides."
 }
 
+variable "exposure_custom_domain" {
+  type    = string
+  default = ""
+
+  description = "Custom domain to map for exposures. This domain must already be verified by Google. If not provided, no domain mapping is created."
```
mikehelmick, line 135: same note as verification server about also pre-setting up DNS

sethvargo:
@mikehelmick P(re)TAL

Issue #791, opened 2020-07-30
Revision token, key rotation


mikehelmick:
### TL;DR

Revision token encryption keys are managed at the DB level.
It would be good to periodically rotate these keys and then age out old keys (thus rendering useless old revision tokens).

Currently thinking to rotate keys ever 8 days and keep the most recent 2 active.

The necessary plumbing for key rotation is already built in
https://github.com/google/exposure-notifications-server/blob/main/internal/revision/database/revision.go

### Design

**Proposal**

Build this into an existing periodic job, i.e. the cleanup job, or create a new period job specifically for key rotation.
In either case, the job will need encryptor/decryptor permissions to wrap the newly generated key for storage.

The existing locking mechanism should be used so that in a race condition, two jobs to not simultaneously rotate the key. Doing so wouldn't be the worst thing in the world, but you would have more keys active than necessary.

Keys that have been active for more than 15 days should be destroyed. This will make it so the sever can no longer decrypt these older revision keys.

**Alternatives considered**

Considered not rotating keys, but certainly not a best practice.

Can't use auto KMS rotation (or could on top of this) because this is done w/ envelope encryption.

/kind enhancement
/kind good-first-issue

PR #808, opened 2020-08-04
Boilerplate endpoint for key rotation

whaught:

## Proposed Changes
- Create a main entrypoint for checking keys and performing a rotation
   - We'll hook this up to a terraform config for cloud scheduler later
- Create a boilerplate handler that does nothing
- Plumb in a Config that includes the RevisionDB-important things

### Todo
- A few of the config things will likely be eliminated or modified
- Next we'll fill out the logic for rotation


internal/publish/config.go
```diff
@@ -55,20 +57,25 @@ type Config struct {
...
 	// Normally "still valid" keys can be accepted, but are embargoed.
 	ReleaseSameDayKeys bool `env:"DEBUG_RELEASE_SAME_DAY_KEYS"`
 }
 
+// RevisionTokenConfig represents the configuration and associated environment variables
+// for handling revision tokens.
+type RevisionTokenConfig struct {
+	// Crypto key to use for wrapping/unwrapping the revision token cipher blocks.
+	KeyID     string `env:"REVISION_TOKEN_KEY_ID"`
+	AAD       string `env:"REVISION_TOKEN_AAD"` // must be base64 encoded, may come from secret://
```
sethvargo, line 73: We do [this](https://github.com/google/exposure-notifications-verification-server/blob/77c550f2db7d5a9c01ab0ec364f6027cec257a25/pkg/config/server_config.go#L140-L162) over in the verification server which will handle a base64-encoded byte slice. Then you can drop `RevisionTokenADDBytes()`.

sethvargo: Otherwise you have to base64 decode it at runtime and worry about caching the result.

sethvargo: (I realize this came from the existing thing above, but while we're moving it around...)

internal/publish/config.go
```diff
@@ -47,6 +47,8 @@ type Config struct {
 	Verification          verification.Config
 	ObservabilityExporter observability.Config
 
+	RevisionTokenConfig
```
sethvargo, line 50: Nit, can you name instead of embeddeing?

sethvargo, line 50:
```suggestion
	RevisionToken RevisionTokenConfig
```

sethvargo: We also generally don't put `Config` at the end since it's part of the `Config` struct.

internal/keyrotation/keyrotation.go
```diff
@@ -0,0 +1,77 @@
...
+	"go.opencensus.io/trace"
+
+	"github.com/google/exposure-notifications-server/internal/logging"
+	revisiondb "github.com/google/exposure-notifications-server/internal/revision/database"
+	"github.com/google/exposure-notifications-server/internal/serverenv"
+)
+
+// NewRotationHandler creates a http.Handler that manages deletion of
+// old export files that are no longer needed by clients for download.
+func NewRotationHandler(config *Config, env *serverenv.ServerEnv) (http.Handler, error) {
```
sethvargo, line 31: We don't have a lot of consistency right now, but I'd like us to start moving to the pattern in `internal/export` where we return a server, the handlers hang off the server, and the server produces its own routes ([example](https://github.com/google/exposure-notifications-server/blob/main/internal/export/server.go#L48)). Then `handler` becomes `server` and `ServeHTTP` becomes `HandleRotate` and returns an `http.Handler`. It makes testing and composing things a bit easier.

internal/keyrotation/keyrotation.go
```diff
@@ -0,0 +1,77 @@
...
+	}, nil
+}
+
+type handler struct {
+	config     *Config
+	env        *serverenv.ServerEnv
+	revisionDB *revisiondb.RevisionDB
+}
+
+func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
sethvargo, line 61: Please use a pointer receiver.

internal/keyrotation/keyrotation.go
```diff
@@ -0,0 +1,77 @@
...
+	ctx, span := trace.StartSpan(r.Context(), "(*keyrotation.handler).ServeHTTP")
+	defer span.End()
+
+	logger := logging.FromContext(ctx)
```
sethvargo, line 65: I started doing it in the verification server - we should name these loggers.

sethvargo, line 65:
```suggestion
	logger := logging.FromContext(ctx).Named("keyrotation.HandleRotate")
```

internal/keyrotation/keyrotation_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+			got, err := NewRotationHandler(&Config{}, tc.env)
+			if tc.err != nil {
+				if err.Error() != tc.err.Error() {
+					t.Fatalf("got %+v: want %v", err, tc.err)
+				}
+			} else if err != nil {
+				t.Fatalf("got unexpected error: %v", err)
+			} else {
+				handler, ok := got.(*handler)
```
sethvargo, line 70: Yea, if you switch to the pattern I described above, this gets much easier to test 😄

internal/keyrotation/keyrotation.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+type handler struct {
+	config     *Config
+	env        *serverenv.ServerEnv
+	revisionDB *revisiondb.RevisionDB
+}
+
+func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx, span := trace.StartSpan(r.Context(), "(*keyrotation.handler).ServeHTTP")
```
sethvargo, line 62: This is probably fine for now. We should investigate whether it makes sense to make this a middleware so we don't litter handlers with unnecessary tracing logic.

internal/keyrotation/config.go
```diff
@@ -0,0 +1,64 @@
...
+	// TODO(whaught): Copied. Not sure if we will actually need all of these.
+	Database              database.Config
+	SecretManager         secrets.Config
+	Storage               storage.Config
+	ObservabilityExporter observability.Config
+
+	RevisionToken publish.RevisionTokenConfig
+
+	Port           string        `env:"PORT, default=8080"`
+	RotationPeriod time.Duration `env:"ROTATION_PERIOD, default=14d"`
```
sethvargo, line 46: Can you doc RotationPeriod and MinKeys?

cmd/key-rotation/main.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/google/exposure-notifications-server/internal/logging"
+	_ "github.com/google/exposure-notifications-server/internal/observability"
+	"github.com/google/exposure-notifications-server/internal/setup"
+	"github.com/google/exposure-notifications-server/pkg/server"
+	"github.com/sethvargo/go-signalcontext"
+)
+
+func main() {
+	ctx, done := signalcontext.OnInterrupt()
+	defer done()
```
sethvargo, line 33: This will never actually run. We have to do something like this instead:

```go
func main() {
	ctx, done := signalcontext.OnInterrupt()

	err := realMain(ctx)
	done()

	if err != nil {
		logger.Fatal(err)
	}
	logger.Info("successful shutdown")
}
```

(yea, I know all the other ones are like this, I'll submit a PR shortly to update).

sethvargo: https://github.com/google/exposure-notifications-server/pull/809

internal/keyrotation/config.go
```diff
@@ -0,0 +1,64 @@
...
+var _ setup.SecretManagerConfigProvider = (*Config)(nil)
+var _ setup.ObservabilityExporterConfigProvider = (*Config)(nil)
+
+// Config represents the configuration and associated environment variables for
+// the key rotation components.
+type Config struct {
+	// TODO(whaught): Copied. Not sure if we will actually need all of these.
+	Database              database.Config
+	SecretManager         secrets.Config
+	Storage               storage.Config
```
mikehelmick, line 40: this one shouldn't need blobstore

does need KeyManager though

internal/keyrotation/config.go
```diff
@@ -0,0 +1,64 @@
...
+	ObservabilityExporter observability.Config
+
+	RevisionToken publish.RevisionTokenConfig
+
+	Port           string        `env:"PORT, default=8080"`
+	RotationPeriod time.Duration `env:"ROTATION_PERIOD, default=14d"`
+	MinKeys        int           `env:"MIN_KEYS, default=2"`
```
mikehelmick, line 47: it's no Min # of keys that's needed - it's a min effective time period. 

We need to make sure we have 15 days of coverage - theoretically 14d, but timezones are hard, so we usually round up to 15d.

internal/keyrotation/keyrotation.go
mikehelmick: if we're going to name them, we should probably install them in the propagated context

internal/keyrotation/keyrotation.go
```diff
@@ -0,0 +1,77 @@
...
+	defer span.End()
+
+	logger := logging.FromContext(ctx)
+
+	// TODO(whaught):
+	// 1. Retrieve keys
+	// 2. Early exit if the newest is new (configurable)
+	// 2. Otherwise generate a key and store
+	// 3. delete oldest keys, but always keep recent and min 2
```
mikehelmick, line 71: retain anything created in last 15d or that is still primary key

internal/keyrotation/keyrotation.go
```diff
@@ -0,0 +1,77 @@
...
+	revisionDB *revisiondb.RevisionDB
+}
+
+func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx, span := trace.StartSpan(r.Context(), "(*keyrotation.handler).ServeHTTP")
+	defer span.End()
+
+	logger := logging.FromContext(ctx)
+
+	// TODO(whaught):
```
mikehelmick, line 67: also - lock incase this job gets invoked in parallel (I know, unlikely)

internal/keyrotation/keyrotation_test.go
```diff
@@ -0,0 +1,79 @@
...
+	testDB := database.NewTestDatabase(t)
+	emptyKMS := &keys.GoogleCloudKMS{}
+
+	testCases := []struct {
+		name string
+		env  *serverenv.ServerEnv
+		err  error
+	}{
+		{
+			name: "nil Database",
```
mikehelmick, line 40: it's best to have this be

`nil_database` because that is what go test will output on failure - a lesson I learned too late, trying to fix test case names as I go

internal/publish/config.go
whaught: But here there are two Configs in this package. I could split it out elsewhere?

cmd/key-rotation/main.go
```diff
@@ -0,0 +1,69 @@
...
+	var config keyrotation.Config
+	env, err := setup.Setup(ctx, &config)
+	if err != nil {
+		return fmt.Errorf("setup.Setup: %w", err)
+	}
+	defer env.Close(ctx)
+
+	handler, err := keyrotation.NewRotationHandler(&config, env)
+	if err != nil {
+		return fmt.Errorf("cleanup.NewRotationHandler: %w", err)
```
MushuEE, line 56: should this be `keyrotation.NewRotationHandler`?

whaught: ah my copy/paste is showing

internal/publish/config.go
whaught: Done!

PR #809, opened 2020-08-04
Ensure main's context's are closed

sethvargo:


cmd/cleanup-export/main.go
```diff
@@ -30,12 +30,15 @@ import (
...
 func main() {
 	ctx, done := signalcontext.OnInterrupt()
-	defer done()
 
-	if err := realMain(ctx); err != nil {
-		logger := logging.FromContext(ctx)
+	err := realMain(ctx)
+	done()
+
+	logger := logging.FromContext(ctx)
```
mikehelmick, line 37: I think this may work, but breaks the contract around context. The logging.FromContext call should come before the done() call.

sethvargo: Done.

sethvargo:
@mikehelmick PT(another)L

PR #813, opened 2020-08-05
make publish test like way faster

mikehelmick:
reuse database for each test

### DIFF
--- a/internal/publish/publish_test.go
+++ b/internal/publish/publish_test.go
@@ -24,6 +24,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"encoding/pem"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"net/http"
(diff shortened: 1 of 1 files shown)


internal/publish/publish_test.go
```diff
@@ -129,13 +130,40 @@ func issueJWT(t *testing.T, cfg jwtConfig) (jwtText, hmacKey string) {
...
+	}
+}
+
+func (n *nameAssigner) next() string {
+	n.modifier++
+	n.lastAssigned = fmt.Sprintf("%s.%d", n.baseAPK, n.modifier)
+	return n.lastAssigned
+}
+
+func (n *nameAssigner) last() string {
```
whaught, line 151: I might call it "peek()" or "current()" given that it's the only one kept around.

mikehelmick: renamed to current

PR #814, opened 2020-08-05
use specific errors for revision token issues

mikehelmick:
This will enable fixing the 500 level errors on revision token issues, but will join up after #806 merges in

### DIFF
--- a/internal/publish/database/exposure.go
+++ b/internal/publish/database/exposure.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"database/sql"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


internal/publish/database/exposure.go
```diff
@@ -39,6 +40,18 @@ const (
 	InsertExposuresBatchSize = 500
 )
 
+var (
+	// ErrExistingKeyNotInToken is returned when attmping to present an exposure that already exists, but
```
whaught, line 44:
```suggestion
	// ErrExistingKeyNotInToken is returned when attempting to present an exposure that already exists, but
```

internal/publish/database/exposure_test.go
```diff
@@ -286,11 +287,42 @@ func TestReviseExposures(t *testing.T) {
 
 	revisions := []*model.Exposure{revisedExp, newExp}
 	// In the first pass - try to revise the key without the necessary revision token.
-	wantError := "attempt to revise key not in revision token"
 	if _, err := pubDB.InsertAndReviseExposures(ctx, revisions, nil, true); err == nil {
 		t.Fatalf("expected error revising without token data")
-	} else if !strings.Contains(err.Error(), wantError) {
-		t.Fatalf("wrong error, want: '%v' got: %v", wantError, err)
+	} else if !errors.Is(err, ErrNoRevisionToken) {
+		t.Fatalf("wrong error, want: '%v' got: %v", ErrNoRevisionToken, err)
```
whaught, line 293: got before want (here and below)

Issue #812, opened 2020-08-04
Terraform broken on fresh project


mikehelmick:
service_cleanup_export.tf line 88
  for_each= merge(

local.common_cloudrun_env_vars is a tuple with 11 elements
var.service_environment is empty map of map of string

call to function "merge" failed: arguments must be maps or objects, got "tuple"

/kind bug
/assign @sethvargo

PR #816, opened 2020-08-05
Fix terraform maps -> tuples

sethvargo:
Fixes GH-812


terraform/service_cleanup_export.tf
```diff
@@ -79,7 +79,7 @@ resource "google_cloud_run_service" "cleanup-export" {
 
         resources {
           limits = {
-            cpu    = "2"
+            cpu    = "2000m"
```
sethvargo, line 82: Not sure if this is a recent Cloud Run API change or a Terraform bug, but I was getting a consistent diff on this.

PR #817, opened 2020-08-05
Logic for key rotation

whaught:

## Proposed Changes
- logic for key rotation
   - added a mutex for read/write
   - retrieves keys
   - creates new keys
   - deletes old ones

### DIFF
--- a/internal/cleanup/cleanup_test.go
+++ b/internal/cleanup/cleanup_test.go
@@ -49,6 +49,7 @@ func TestNewExposureHandler(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
+		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 
@@ -101,6 +102,7 @@ func TestNewExportHandler(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
+		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 
--- a/internal/cleanup/cleanup_test.go
+++ b/internal/cleanup/cleanup_test.go
@@ -49,8 +49,6 @@ func TestNewExposureHandler(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		tc := tc
-
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 
@@ -86,25 +84,23 @@ func TestNewExportHandler(t *testing.T) {
 		err  error
 	}{
 		{
-			name: "nil Database",
+			name: "nil_database",
 			env:  serverenv.New(ctx),
 			err:  fmt.Errorf("missing database in server environment"),
 		},
 		{
-			name: "nil Blobstore",
+			name: "nil_blobstore",
 			env:  serverenv.New(ctx, serverenv.WithDatabase(testDB)),
 			err:  fmt.Errorf("missing blobstore in server environment"),
 		},
 		{
-			name: "Fully Specified",
+			name: "fully_specified",
 			env:  serverenv.New(ctx, serverenv.WithBlobStorage(noopBlobstore), serverenv.WithDatabase(testDB)),
 			err:  nil,
 		},
 	}
 
 	for _, tc := range testCases {
-		tc := tc
-
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 
--- a/internal/keyrotation/rotate.go
+++ b/internal/keyrotation/rotate.go
@@ -36,7 +36,7 @@ func (s *Server) handleRotateKeys(ctx context.Context) http.HandlerFunc {
 		defer s.mu.Unlock()
 
 		if err := s.doRotate(ctx); err != nil {
-			logger.Error(err)
+			logger.Errorw("failed to rotate", "error", err)
 			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
--- a/internal/keyrotation/rotate.go
+++ b/internal/keyrotation/rotate.go
@@ -33,6 +33,7 @@ func (s *Server) handleRotateKeys(ctx context.Context) http.HandlerFunc {
 		_, span := trace.StartSpan(r.Context(), "(*keyrotation.handler).ServeHTTP")
 		defer span.End()
 
+		// TODO(whaught): This mutex should be a DB lock. Doesn't help for many instances.
 		s.mu.Lock()
 		defer s.mu.Unlock()
 
--- a/internal/keyrotation/rotate.go
+++ b/internal/keyrotation/rotate.go
@@ -35,7 +35,7 @@ func (s *Server) handleRotateKeys(ctx context.Context) http.HandlerFunc {
 		s.mu.Lock()
 		defer s.mu.Unlock()
 
-		if err := s.handleRotation(ctx); err != nil {
+		if err := s.doRotate(ctx); err != nil {
 			logger.Error(err)
 			w.WriteHeader(http.StatusInternalServerError)
 			return
@@ -46,7 +46,7 @@ func (s *Server) handleRotateKeys(ctx context.Context) http.HandlerFunc {
 	}
 }
 
-func (s *Server) handleRotation(ctx context.Context) error {
+func (s *Server) doRotate(ctx context.Context) error {
 	metrics := s.env.MetricsExporter(ctx)
 
 	_, allowed, err := s.revisionDB.GetAllowedRevisionKeys(ctx)
--- a/internal/keyrotation/rotate.go
+++ b/internal/keyrotation/rotate.go
@@ -57,7 +57,7 @@ func (s *Server) doRotate(ctx context.Context) error {
 	// First allowed is newest due to sql orderby.
 	if len(allowed) == 0 || time.Since(allowed[0].CreatedAt) >= s.config.NewKeyPeriod {
 		if _, err := s.revisionDB.CreateRevisionKey(ctx); err != nil {
-			return err
+			return fmt.Errorf("failed to create revision key: %w", err)
 		}
 		metrics.WriteInt("revision-keys-created", true, 1)
 	}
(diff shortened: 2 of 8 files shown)


internal/cleanup/cleanup_test.go
```diff
@@ -49,8 +49,6 @@ func TestNewExposureHandler(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		tc := tc
```
sethvargo, line 52: You still need this. Otherwise all tests will run as the last entry in `testCases` because of how `t.Parallel()` works.

internal/cleanup/cleanup_test.go
```diff
@@ -86,25 +84,23 @@ func TestNewExportHandler(t *testing.T) {
...
 		{
-			name: "Fully Specified",
+			name: "fully_specified",
 			env:  serverenv.New(ctx, serverenv.WithBlobStorage(noopBlobstore), serverenv.WithDatabase(testDB)),
 			err:  nil,
 		},
 	}
 
 	for _, tc := range testCases {
-		tc := tc
```
sethvargo, line 106: Still need this

internal/keyrotation/rotate.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+func (s *Server) handleRotateKeys(ctx context.Context) http.HandlerFunc {
+	logger := logging.FromContext(ctx).Named("keyrotation.HandleRotate")
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		_, span := trace.StartSpan(r.Context(), "(*keyrotation.handler).ServeHTTP")
+		defer span.End()
+
+		s.mu.Lock()
```
sethvargo, line 35: I guess we can cheat and set the configuration for this service to be `maxInstances=1`, but we should consider a database-level lock or something.

internal/keyrotation/rotate.go
```diff
@@ -0,0 +1,84 @@
...
+		defer s.mu.Unlock()
+
+		if err := s.doRotate(ctx); err != nil {
+			logger.Error(err)
```
sethvargo, line 39:
```suggestion
			logger.Errorw("failed to rotate", "error", err)
```

internal/keyrotation/rotate.go
```diff
@@ -0,0 +1,84 @@
...
+		w.WriteHeader(http.StatusOK)
+	}
+}
+
+func (s *Server) doRotate(ctx context.Context) error {
+	metrics := s.env.MetricsExporter(ctx)
+
+	_, allowed, err := s.revisionDB.GetAllowedRevisionKeys(ctx)
+	if err != nil {
+		return fmt.Errorf("rotate-keys unable to read revision keys: %v", err)
```
sethvargo, line 54:
```suggestion
		return fmt.Errorf("rotate-keys unable to read revision keys: %w", err)
```

internal/keyrotation/rotate.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	// First allowed is newest due to sql orderby.
+	if len(allowed) == 0 || time.Since(allowed[0].CreatedAt) >= s.config.NewKeyPeriod {
+		if _, err := s.revisionDB.CreateRevisionKey(ctx); err != nil {
+			return err
```
sethvargo, line 60:
```suggestion
			return fmt.Errorf("failed to create revision key: %w", err)
```

internal/keyrotation/rotate.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		metrics.WriteInt("revision-keys-created", true, 1)
+	}
+
+	if len(allowed) < 2 {
+		return nil
```
sethvargo, line 66: Should we log or produce a metric here? Shouldn't 2 come from the config instead of being hard-coded?

internal/keyrotation/rotate.go
```diff
@@ -0,0 +1,84 @@
...
+	deleted := 0
+	defer metrics.WriteInt("revision-keys-deleted", true, deleted)
+
+	for _, key := range allowed[2:] {
+		if time.Since(key.CreatedAt) < s.config.DeleteOldKeyPeriod {
+			continue
+		}
+
+		if err := s.revisionDB.DestroyKey(ctx, key.KeyID); err != nil {
+			return err
```
sethvargo, line 78: - Wrap error
- We might want to consider a multi-error here. Since we're deleting multiple keys.

internal/keyrotation/rotate.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		metrics.WriteInt("revision-keys-created", true, 1)
+	}
+
+	if len(allowed) < 2 {
+		return nil
+	}
+
+	deleted := 0
+	defer metrics.WriteInt("revision-keys-deleted", true, deleted)
```
sethvargo, line 70: If you're willing to switch to multierror, we can move this to after the loop and save the level of indirect.

internal/keyrotation/rotate_test.go
```diff
@@ -0,0 +1,89 @@
...
+		name           string
+		expectKeyCount int64
+	}{
+		{
+			name:           "empty_db",
+			expectKeyCount: 1,
+		},
+	}
+
+	for _, tc := range testCases {
```
sethvargo, line 48: Need `tc := tc`

internal/keyrotation/rotate_test.go
```diff
@@ -0,0 +1,89 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			keyID := "test" + t.Name()
```
sethvargo, line 50: `t.Parallel()`

internal/keyrotation/rotate_test.go
```diff
@@ -0,0 +1,89 @@
...
+			}
+
+			if count != tc.expectKeyCount {
+				t.Errorf("Purged %d keys, wanted %d", count, tc.expectKeyCount)
+			}
+		})
+	}
+}
+
+func purgeAllKeys(ctx context.Context, t *testing.T, db *database.DB) (int64, error) {
```
sethvargo, line 78: If this is a test helper, please prefix it with `test` (lowercase t). Please also accept `testing.TB` (interface) instead of `*testing.T`.

sethvargo: There's also ongoing debate as both testing.TB and context.Context are stylistically required to be "first".

internal/keyrotation/rotate_test.go
```diff
@@ -0,0 +1,89 @@
...
+	t.Helper()
+	var count int64 = 0
+	if err := db.InTx(ctx, pgx.Serializable, func(tx pgx.Tx) error {
```
sethvargo, line 81: This shouldn't be serializable if you truly want to delete everything.

internal/keyrotation/server.go
```diff
@@ -19,14 +19,22 @@ import (
...
 )
 
+// Server hosts end points to manage key rotation
+type Server struct {
+	config     *Config
+	env        *serverenv.ServerEnv
+	revisionDB *revisiondb.RevisionDB
+
+	// mutex lock to be held over read/write operations when rotating keys.
+	mu sync.RWMutex
```
sethvargo, line 35: I noted above, but I think we should acquire a table-level lock on revisionDB instead of relying on a mutex. A mutex isn't shared across instances.

internal/cleanup/cleanup_test.go
whaught: Ah thanks. Sorry was just poking at something that looked funny to me.

internal/keyrotation/rotate_test.go
whaught: I was worried about that since these are sharing the database and does a blind write to clear at the end of the test. Need to consider parallel safety.

This test needs work!

sethvargo: You can move the database creation into the parallelization too

internal/keyrotation/rotate.go
mikehelmick: yeah - db level lock is preferred

internal/keyrotation/rotate.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// First allowed is newest due to sql orderby.
+	if len(allowed) == 0 || time.Since(allowed[0].CreatedAt) >= s.config.NewKeyPeriod {
+		if _, err := s.revisionDB.CreateRevisionKey(ctx); err != nil {
+			return fmt.Errorf("failed to create revision key: %w", err)
+		}
+		metrics.WriteInt("revision-keys-created", true, 1)
+	}
+
+	if len(allowed) < 2 {
```
mikehelmick, line 65: shouldn't be 2 - needs to be time based

A key needs to be allowed for a specific min duration (config, but I think 15d is default)
Once a key is no longer the primary and was created < 15d ago, it can be destroyed.

whaught: It only deletes if older than s.config.DeleteOldKeyPeriod... It also always keeps at least two. We can drop that if we don't need any kind of minimum?

mikehelmick: ah, the minimum number of keys needed is 1.

We just need to not destroy the key material if the revision token might be accepted (minted less than 15d ago). Once a token is that old, we don't need to be able to unlock it any longer (data is useless).

internal/keyrotation/rotate.go
```diff
@@ -0,0 +1,84 @@
...
+		return nil
+	}
+
+	deleted := 0
+	defer metrics.WriteInt("revision-keys-deleted", true, deleted)
+
+	for _, key := range allowed[2:] {
```
mikehelmick, line 72: so here the slice doesn't need to be a sub-slice

internal/keyrotation/rotate.go
whaught: I'm saying suppose all keys are older than two weeks. Should we destroy all of them or is there any use in keeping around the one active one? That should never happen, but trying to make the code safe.

internal/keyrotation/rotate.go
whaught: I like this multi-error. Reads better than the defer metric too.

PR #826, opened 2020-08-06
Add traveler bit to exports

### LINKED ISSUE #794: Travellers: Allow export config to include traveller keys
Once #792 is in

Allow for the exportconfig to include any "traveller" keys in addition to matching region keys

To support federation and international roaming

/kind enhancement

mikehelmick:
Fixes #794 

## Proposed Changes

* Add include_travlers to all export models
* Set that bit on the iterate exposures when generating exports (using existing functionality
* Update admin console so this can be configured

**Release Note**

```release-note
ExportConfig supports the v1 API traveler bit, allowing exports to include "home" region and keys of federated-in travelers.
```

### DIFF
--- a/internal/admin/exports/form.go
+++ b/internal/admin/exports/form.go
@@ -24,16 +24,17 @@ import (
 )
 
 type formData struct {
-	OutputRegion string        `form:"OutputRegion"`
-	InputRegions string        `form:"InputRegions"`
-	BucketName   string        `form:"BucketName"`
-	FilenameRoot string        `form:"FilenameRoot"`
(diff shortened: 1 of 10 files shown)


sherifkozman:
@mikehelmick how does that affect deployments with multiple regions/HAs being deployed within the same sever setup Vs federated deployments ?

mikehelmick:
should have no change to existing deployments. The traveller bit was introduced in the V1 API.

If multiple HAs are sharing the same server, my recommendation is to have them write to the same "region" or at least have their exports pull in all the regions to a single export file. This will give you maximum coverage.

PR #828, opened 2020-08-06
Key Rotation E2E Test

whaught:
Fixes https://github.com/google/exposure-notifications-server/issues/791

## Proposed Changes

* Start the key rotation endpoint in the testserver
    * Set it to rotate ~seconds
* Create an e2e test that asserts new keys created
* Change the error message if a revision token is passed, but doesn't decrypt

**Release Note**

```release-note
E2E test for key rotation
```


internal/integration/integration_test.go
```diff
@@ -225,6 +225,12 @@ func TestIntegration(t *testing.T) {
 		}
 	}
 
+	// Rotate Keys
+	time.Sleep(2 * time.Second) // Ensure DeleteOldKeyPeriod is elapsed
```
mikehelmick, line 227: should probably do this before we publish keys and then verify that we god a revision token back from the publish.

Actually using that revision token would be part of the upcoming revision test.

whaught: sure we can do that. Really we want to rotate them and then assert that an old one is rejected

whaught: but at the least this adds coverage from a client-request perspective. We can also see from the logger that the thing happens.

mikehelmick: yeah, that would be great

I think that the new one isn't "rejected" but it acts as if you didn't pass one at all

PR #829, opened 2020-08-06
Enabling the verification flow by creating a dummy health authority and a valid verification certificate

argetlam-cs:
Fixes # 712

## Proposed Changes

* Enables the verification flow in the integration test
     * Creates a dummy health authority in the setup
     * Issues a valid verification certificate for the dummy health authority

**Release Note**

```release-note
Integration test that enables the verification flow
```

### DIFF
--- a/internal/integration/helpers.go
+++ b/internal/integration/helpers.go
@@ -21,7 +21,6 @@ import (
 	"crypto/rand"
 	"crypto/sha256"
 	"crypto/x509"
-	"encoding/base64"
 	"encoding/pem"
 	"fmt"
 	"net/http"
@@ -30,7 +29,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/dgrijalva/jwt-go"
 	"github.com/google/exposure-notifications-server/internal/authorizedapp"
 	"github.com/google/exposure-notifications-server/internal/cleanup"
 	"github.com/google/exposure-notifications-server/internal/database"
@@ -58,24 +56,13 @@ import (
 var (
 	ExportDir    = "my-bucket"
 	FileNameRoot = "/"
-	jwtCfg       = jwtConfig{}
+	jwtCfg       = utils.JwtConfig{}
 )
 
 // Holds a single signing key and the PEM public key.
 type signingKey struct {
 	Key       *ecdsa.PrivateKey
 	PublicKey string
-}
-
-// Config to fetch a verification jwt certificate
-type jwtConfig struct {
-	HealthAuthority      *vm.HealthAuthority
-	HealthAuthorityKey   *vm.HealthAuthorityKey
-	ExposureKeys         []verifyapi.ExposureKey
-	Key                  *ecdsa.PrivateKey
-	JWTWarp              time.Duration
-	ReportType           string
-	SymptomOnsetInterval uint32
 }
 
 // NewTestServer sets up clients used for integration tests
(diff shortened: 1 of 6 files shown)


internal/integration/helpers.go
```diff
@@ -319,3 +375,63 @@ func Eventually(tb testing.TB, times uint64, f func() error) {
...
+
+	return &signingKey{
+		Key:       privateKey,
+		PublicKey: pemPublicKey,
+	}
+}
+
+// Based on the publish request, generate a JWT as if it came from the
+// authorized health authority.
+func issueJWT(t *testing.T, cfg jwtConfig) (jwtText, hmacKey string) {
```
mikehelmick, line 402: I Think this came from another test - can we put it in a centrally located place and have them both depend on it?

can't be here because I think that would be a circular dependency

chaodaiG:
cc @MushuEE , this PR handles the TODO that we just talked about

internal/integration/helpers.go
argetlam-cs: Moved it to the verification util lib

pkg/verification/utils.go
```diff
@@ -26,16 +26,34 @@
...
+	"time"
 
+	"github.com/dgrijalva/jwt-go"
+	vm "github.com/google/exposure-notifications-server/internal/verification/model"
 	verifyapi "github.com/google/exposure-notifications-server/pkg/api/v1"
 	"github.com/google/exposure-notifications-server/pkg/api/v1alpha1"
 )
 
+// JwtConfig stores the config used to fetch a verification jwt certificate
+type JwtConfig struct {
```
mikehelmick, line 47: this shouldn't be in PKG since it requires that you pass in things from internal. 

needs to live in internal somewhere

argetlam-cs: Done

mikehelmick: the move looks good, but this should be removed - I think just revert this file.

argetlam-cs: Done.

internal/testing/utils.go
```diff
@@ -0,0 +1,88 @@
+// Copyright 2020 Google LLC
```
sethvargo, line 1: I'd discourage naming this "testing". That will conflict with `testing.T`

argetlam-cs: renamed it to utils

internal/integration/helpers.go
```diff
@@ -319,3 +362,24 @@ func Eventually(tb testing.TB, times uint64, f func() error) {
 func IndexFilePath() string {
 	return path.Join(FileNameRoot, "index.txt")
 }
+
+func newSigningKey(tb testing.TB) *signingKey {
```
sethvargo, line 366: Does it make sense to move this to the utils package too?

internal/utils/utils.go
```diff
@@ -0,0 +1,88 @@
...
+	"testing"
+	"time"
+
+	"github.com/dgrijalva/jwt-go"
+	vm "github.com/google/exposure-notifications-server/internal/verification/model"
+	verifyapi "github.com/google/exposure-notifications-server/pkg/api/v1"
+	vutil "github.com/google/exposure-notifications-server/pkg/verification"
+)
+
+// JwtConfig stores the config used to fetch a verification jwt certificate
```
sethvargo, line 40: Nit `JWTConfig`

internal/integration/helpers.go
argetlam-cs: Moved it

internal/utils/utils.go
argetlam-cs: Done

PR #830, opened 2020-08-06
Deletion based off of createdAt of next key

whaught:

## Proposed Changes

* Only rotate keys if they are 15d older than the current effective key
    * Current impl compares to now

### DIFF
--- a/internal/keyrotation/rotate.go
+++ b/internal/keyrotation/rotate.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/google/exposure-notifications-server/internal/logging"
+	"github.com/google/exposure-notifications-server/internal/revision/database"
 	"github.com/hashicorp/go-multierror"
 	"go.opencensus.io/trace"
 )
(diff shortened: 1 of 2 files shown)


internal/keyrotation/rotate.go
```diff
@@ -68,17 +68,22 @@ func (s *Server) doRotate(ctx context.Context) error {
...
 		metrics.WriteInt("revision-keys-created", true, 1)
+	} else {
+		effectiveCreated = allowed[0].CreatedAt
 	}
 
 	var result error
 	deleted := 0
 	for _, key := range allowed {
-		if time.Since(key.CreatedAt) < s.config.DeleteOldKeyPeriod {
+		if effectiveCreated.Sub(key.CreatedAt) < s.config.DeleteOldKeyPeriod {
```
mikehelmick, line 86: I think effectiveCreated needs to get updated in this loop

internal/keyrotation/rotate.go
```diff
@@ -62,29 +62,39 @@ func (s *Server) handleRotateKeys(ctx context.Context) http.HandlerFunc {
...
 
 	var result error
 	deleted := 0
 	for _, key := range allowed {
-		if time.Since(key.CreatedAt) < s.config.DeleteOldKeyPeriod {
+		if key.KeyID == effectiveID {
 			continue
 		}
+		if time.Since(previousCreated) < s.config.DeleteOldKeyPeriod {
+			continue // A key is not safe to delete until the newer one was effective for the period.
```
mikehelmick, line 91: I think in this path we still want to update previousCreated

this key isn't ready to be deleted, but it can still push the timestamp down for the next in line.

internal/keyrotation/rotate.go
```diff
@@ -62,29 +62,39 @@ func (s *Server) handleRotateKeys(ctx context.Context) http.HandlerFunc {
...
+		}
 		if err := s.revisionDB.DestroyKey(ctx, key.KeyID); err != nil {
 			result = multierror.Append(result, err)
```
mikehelmick, line 94: here too?

internal/keyrotation/rotate_test.go
```diff
@@ -46,7 +46,9 @@ func TestRotateKeys(t *testing.T) {
 	key, aad, wrapped := testMakeKey(ctx, t, kms, keyID)
 
 	neg20Days, _ := time.ParseDuration("-480h")
+	neg40Days, _ := time.ParseDuration("-480h")
```
mikehelmick, line 49: neg20 and neg40 have the same value

whaught: interestingly it still passes (and should) even if the gap is 0. The previous key was still effective for more than two weeks.

Issue #723, opened 2020-07-15
Smoke test terraform deployment


chaodaiG:
### TL;DR

Create smoke tests that could test terraform deployment, as well as automated tests for ensuring terraform deployment work

### Design

**Proposal**

chaodaiG:
@sethvargo  and @stati0n, I have created this issue tracking the discussion of this topic. I could be wrong but I'm considering it being a prerequisite for testing Cloud Run

stati0n:
It shouldn't be a hard blocker for Cloud Run.  We could orchestrate a deployment to Cloud Run ourselves via a Cloud Build.

chaodaiG:
> It shouldn't be a hard blocker for Cloud Run. We could orchestrate a deployment to Cloud Run ourselves via Cloud Build.

That's true for services, but for other resources like database, gcs etc. will need to be configured as what's been done by terraform. Or alternatively, we could deploy services using in mem databases?

stati0n:
We could.  I'm willing to say it's a prereq if we think it's enough work to work around it.

chaodaiG:
> We could. I'm willing to say it's a prereq if we think it's enough work to work around it.

Makes sense. I will create another issue gathering requirement for testing Cloud Run

chaodaiG:
Created #727 for discussing Cloud Run tests

chaodaiG:
Done in https://github.com/google/exposure-notifications-server/pull/948

/close

PR #840, opened 2020-08-07
Integration test works with existing GCP PostgresSQL

chaodaiG:
- Allow working with existing DB
- Add script bridging terraform with integration test

This PR has been tested with the database provisioned by terraform. Follow up would be figuring out how to test servers provisioned by terraform

Part of: #723

### DIFF
--- a/internal/integration/helpers.go
+++ b/internal/integration/helpers.go
@@ -40,6 +40,7 @@ import (
 	"github.com/google/exposure-notifications-server/pkg/keys"
 	"github.com/google/exposure-notifications-server/pkg/secrets"
 	"github.com/google/exposure-notifications-server/pkg/server"
+	"github.com/sethvargo/go-envconfig"
 	"github.com/sethvargo/go-retry"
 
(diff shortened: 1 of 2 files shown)


whaught:
@assign @sethvargo

internal/integration/helpers.go
```diff
@@ -130,6 +133,22 @@ func testServer(tb testing.TB) (*serverenv.ServerEnv, *http.Client) {
 	}
 
 	db := database.NewTestDatabase(tb)
+	if v := os.Getenv("DB_NAME"); v != "" && !testing.Short() {
```
sethvargo, line 136: Why is this needed? I think the database util already sets up the secret manager and stuff, no?

chaodaiG: https://github.com/google/exposure-notifications-server/blob/2dfbe2f7cee56b5f0d6b5b66911d03f2ab29de1c/internal/database/database_util.go#L48
I don't think database_util.go sets up secret manager, not database from environment either

PR #844, opened 2020-08-07
Fix performance test

chaodaiG:
Integrating verification bits from #829  into performance test so that it doesn't fail

### DIFF
--- a/internal/integration/helpers.go
+++ b/internal/integration/helpers.go
@@ -52,13 +52,12 @@ import (
 var (
 	ExportDir    = "my-bucket"
 	FileNameRoot = "/"
-	jwtCfg       = testutil.JWTConfig{}
 )
 
 // NewTestServer sets up clients used for integration tests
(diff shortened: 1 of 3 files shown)


internal/integration/helpers.go
```diff
@@ -103,14 +102,17 @@ func NewTestServer(tb testing.TB, exportPeriod time.Duration) (*serverenv.Server
...
+	return env, enClient, db, jwtCfg
 }
 
 // testServer sets up mocked local servers for running tests
-func testServer(tb testing.TB) (*serverenv.ServerEnv, *http.Client) {
+func testServer(tb testing.TB) (*serverenv.ServerEnv, *http.Client, testutil.JWTConfig) {
 	tb.Helper()
 
-	ctx := context.Background()
+	var (
```
sethvargo, line 112: It's more "go" to do:

```go
ctx := context.Background()
jwtCfg = testutil.JWTConfig{}
```

PR #847, opened 2020-08-07
Add interfaces for various key handlers

sethvargo:
* Add new interfaces for in-memory key management (used for testing)

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/keys/in_memory.go
+++ b/pkg/keys/in_memory.go
@@ -27,6 +27,12 @@ import (
 	"sync"
 )
 
+var _ KeyManager = (*InMemory)(nil)
+var _ SigningKeyCreator = (*InMemory)(nil)
+var _ SigningKeyAdder = (*InMemory)(nil)
+var _ EncryptionKeyCreator = (*InMemory)(nil)
+var _ EncryptionKeyAdder = (*InMemory)(nil)
+
 // InMemory is useful for testing. Do NOT use in a running system as all
 // keys are only kept in memory and will be lost across server reboots.
 type InMemory struct {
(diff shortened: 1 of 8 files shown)


pkg/keys/in_memory.go
```diff
@@ -43,42 +49,68 @@ func NewInMemory(ctx context.Context) (*InMemory, error) {
...
+	return key, nil
+}
+
+// AddEncryptionKey stores the key on the system.
+func (k *InMemory) AddEncryptionKey(keyID string, key []byte) error {
+	k.mu.Lock()
+	defer k.mu.Unlock()
 
+	if _, ok := k.cryptoKeys[keyID]; ok {
+		return fmt.Errorf("key already exists: %v", keyID)
```
chaodaiG, line 110: %q?

pkg/keys/in_memory.go
```diff
@@ -43,42 +49,68 @@ func NewInMemory(ctx context.Context) (*InMemory, error) {
...
-func (k *InMemory) AddEncryptionKey(keyID string) error {
+// CreateEncryptionKey generates and stores new encryption key identified by the
+// provided keyID.
+func (k *InMemory) CreateEncryptionKey(keyID string) ([]byte, error) {
 	k.mu.Lock()
 	defer k.mu.Unlock()
 
 	if _, ok := k.cryptoKeys[keyID]; ok {
-		return fmt.Errorf("key already exists: %v", keyID)
+		return nil, fmt.Errorf("key already exists: %v", keyID)
```
chaodaiG, line 92: %q?

pkg/keys/in_memory.go
sethvargo: `%v` is "best for the thing (which is just the string value). `%q` would quote the string, which isn't desirable here. You could argue for `%s`, which is more technically correct, but this is only used for testing anyway, so I'm not super worried 😄

chaodaiG: make sense
/unhold

PR #848, opened 2020-08-07
Reorganize JWT across perf/integration tests

MushuEE:
This organization cleans up some imports and groups like things
It also make mock JWT more availible and aligns with SRP

Was dupe fix of #844 (lost internet so was not able to coordinate well)
I think this is a bit cleaner and more flexible. 

## Proposed Changes

* move config mock to utils where JTWIssue lives
* build JTW Config within each test based on the keys and db
*


MushuEE:
Didn't mean to open this from main. Replaying from real branch and fixing merge

PR #850, opened 2020-08-08
Always configure a logger on the context

sethvargo:
Always create and include a logger on the context, make logging package public

**Release Note**

```release-note
Always include a logger object and make logging package public
```


pkg/logging/logger.go
```diff
@@ -0,0 +1,156 @@
+// Copyright 2020 Google LLC
```
sethvargo, line 1: Sigh. I git `mv`ed and everything...

PR #851, opened 2020-08-09
Enable perf tests to be run in prod/dev modes more easily

MushuEE:
Add some docs
Split benchmark config + add test flags to make runs more streamline

Link to results dashboards

## Proposed Changes

* Enable dev mode on Mako with `--dev` option to the performance script 
* Add some Docs for perf
* remove need to flip flop comments when making / testing changes in perf

/cc @chaodaiG

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,3 +66,15 @@ performance-test:
 		-tags=performance \
 		./internal/performance
 .PHONY: performance-test
+
+performance-dev-test:
+	@go test \
+		-count=1 \
+		-timeout=30m \
+		-v \
+		-tags=performance \
+		./internal/performance \
+		-args \
+		-dev \
+		-publishes=1000
+.PHONY: performance-dev-test
--- a/Makefile
+++ b/Makefile
@@ -66,15 +66,3 @@ performance-test:
 		-tags=performance \
 		./internal/performance
 .PHONY: performance-test
-
-performance-dev-test:
-	@go test \
-		-count=1 \
-		-timeout=30m \
-		-v \
-		-tags=performance \
-		./internal/performance \
-		-args \
-		-dev \
-		-publishes=1000
-.PHONY: performance-dev-test
--- a/internal/performance/export_test.go
+++ b/internal/performance/export_test.go
@@ -47,17 +47,17 @@ type testConfig struct {
 func TestExport(t *testing.T) {
 	const (
 		keysPerPublish = 14
-		// Consider the above publishes are evenly distributed in 24 hours, and
-		// period is 10 minutes
-		exportPeriod = 10 * time.Minute
-		totalBatches = 24 * 6
+		exportPeriod   = 10 * time.Minute
+		totalBatches   = 24 * 6
 	)
 	var (
 		ctx      = context.Background()
 		criteria = publishdb.IterateExposuresCriteria{
 			OnlyLocalProvenance: false,
 		}
-		numPublishes   = 100000
+		numPublishes = 100000
+		// Consider the above publishes are evenly distributed in 24 hours, and
+		// period is 10 minutes
 		batchStartTime = time.Now().Add(time.Duration(-totalBatches-10) * exportPeriod)
 	)
 	c := testConfig{}
--- a/internal/performance/export_test.go
+++ b/internal/performance/export_test.go
@@ -20,9 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os"
 	"path"
-	"strconv"
 	"strings"
 	"testing"
 	"time"
@@ -37,8 +35,14 @@ import (
 	"github.com/google/exposure-notifications-server/pkg/base64util"
 	"github.com/google/exposure-notifications-server/pkg/util"
 	pgx "github.com/jackc/pgx/v4"
+	"github.com/sethvargo/go-envconfig"
 	"github.com/sethvargo/go-retry"
 )
+
+type testConfig struct {
+	Publishes int  `env:"PUBLISHES,required"`
+	Dev       bool `env:"PERFORMANCE_DEV"`
+}
 
 func TestExport(t *testing.T) {
 	const (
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -66,3 +66,15 @@ performance-test:
...
+	@go test \
+		-count=1 \
+		-timeout=30m \
+		-v \
+		-tags=performance \
+		./internal/performance \
+		-args \
+		-dev \
+		-publishes=1000
+.PHONY: performance-dev-test
```
chaodaiG, line 80: Add a new line at the end of file to make git diff happier

scripts/performance.sh
```diff
@@ -14,9 +14,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+DEV=0
+
+for arg in "$@"
```
chaodaiG, line 19: Nit, since only `--dev` is supported, it might be easier to 
```suggestion
if [[ "$1" == "--dev" || "$1" == "-d" ]]; then
```

MushuEE:
@chaodaiG yep I like that idea! Will make the requisite changes. Should also clean up the makefile and avoid the nasty need for `--args` in go test line.

internal/performance/export_test.go
```diff
@@ -55,12 +53,19 @@ func TestExport(t *testing.T) {
...
-		want           = keysPerPublish * *numPublishes
+		numPublishes   = 100000
+		want           = keysPerPublish * numPublishes
 		batchStartTime = time.Now().Add(time.Duration(-totalBatches-10) * exportPeriod)
-		roughPerBatch  = *numPublishes/totalBatches + 1
+		roughPerBatch  = numPublishes/totalBatches + 1
 	)
 
-	makoQuickstore, cancel := setup(t, *makoDev)
+	if p, err := strconv.Atoi(os.Getenv("PUBLISHES")); err == nil {
```
chaodaiG, line 62: I'm a bit hesitate to allow `PUBLISHES` to be overrode in `prod` env, can we somehow couple this with `dev` env so that we can better prevent unexpected data from being pushed to prod benchmark.

internal/performance/helper.go
```diff
@@ -55,7 +56,7 @@ func setup(tb testing.TB, dev bool) (*quickstore.Quickstore, func(context.Contex
 	defer cancel()
 
 	benchmarkConfig := &mpb.BenchmarkInfo{}
-	if dev {
+	if os.Getenv("PERFORMANCE_DEV") == "1" {
```
chaodaiG, line 59: Can this be part of config as line 45 above, the env vars will be processed by github.com/sethvargo/go-envconfig

internal/performance/export_test.go
MushuEE: I see, I had placed it within the `dev` condition of the `.sh` script so that it could not be set without `--dev` from the command. But someone could still manually set it. I can also check for `PERFORMACE_DEV` 👍

internal/performance/export_test.go
```diff
@@ -55,12 +53,19 @@ func TestExport(t *testing.T) {
...
+		numPublishes = p
```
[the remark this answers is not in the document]
MushuEE: I moved these into the `var` block so I could mutate, is there a better way to handle this?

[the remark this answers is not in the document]
chaodaiG: Yeah that'll work. Also, feel free initialize `want` and `roughPerBatch` lower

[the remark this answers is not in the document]
MushuEE: yep 👌 they are in there.

internal/performance/helper.go
MushuEE: cool, thanks. Trying it out.

MushuEE: Done, I made a new config for both Mako and the tests, in the event that we want to tune more params in the future.

internal/performance/export_test.go
```diff
@@ -35,28 +35,43 @@ import (
...
+		totalBatches = 24 * 6
 	)
 	var (
 		ctx      = context.Background()
 		criteria = publishdb.IterateExposuresCriteria{
 			OnlyLocalProvenance: false,
 		}
+		numPublishes   = 100000
 		want           = keysPerPublish * numPublishes
 		batchStartTime = time.Now().Add(time.Duration(-totalBatches-10) * exportPeriod)
```
chaodaiG, line 62: Move this down to right above line 76 `makoQuickstore, cancel := setup(t)`

internal/performance/export_test.go
```diff
@@ -35,28 +35,43 @@ import (
 	"github.com/google/exposure-notifications-server/pkg/base64util"
 	"github.com/google/exposure-notifications-server/pkg/util"
 	pgx "github.com/jackc/pgx/v4"
+	"github.com/sethvargo/go-envconfig"
 	"github.com/sethvargo/go-retry"
 )
 
+type testConfig struct {
+	Publishes int  `env:"PUBLISHES,required"`
```
chaodaiG, line 43: FYI defaults can be set here

scripts/performance.sh
```diff
@@ -14,6 +14,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+if [[ "$1" == "--dev" || "$1" == "-d" ]]; then
+ export PERFORMANCE_DEV=1
+ pubs = $2
+ echo $2
```
chaodaiG, line 20: Debug?

scripts/performance.sh
```diff
@@ -14,6 +14,13 @@
...
+ export PUBLISHES=${2:-1000}
```
chaodaiG, line 21: Did you mean:
```suggestion
 export PUBLISHES=${pubs:-1000}
```

scripts/performance.sh
MushuEE: yes, thanks

internal/performance/export_test.go
MushuEE: good to know! will update tomorrow morning

scripts/performance.sh
MushuEE: fixed

scripts/performance.sh
MushuEE: I meant `2`, just to take the second argument

internal/performance/export_test.go
MushuEE: done

chaodaiG:
Looks good!
/lgtm
/approve

PR #854, opened 2020-08-09
do basic validation in the export analyzer

mikehelmick:
Adds basic data validation when printing export file.

```release-note
NONE
```

### DIFF
--- a/tools/export-analyzer/main.go
+++ b/tools/export-analyzer/main.go
@@ -50,7 +50,9 @@ func main() {
 	}
 
 	// Do some basic data validation.
-	checkExportFile(keyExport)
+	if !*quiet {
+		checkExportFile(keyExport)
+	}
 
 	prettyJSON, err := json.MarshalIndent(keyExport, "", "  ")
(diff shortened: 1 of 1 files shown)


tools/export-analyzer/main.go
```diff
@@ -45,9 +49,44 @@ func main() {
...
 	prettyJSON, err := json.MarshalIndent(keyExport, "", "  ")
 	if err != nil {
 		log.Fatalf("error pretty printing export: %v", err)
 	}
 	log.Printf("%v", string(prettyJSON))
 }
+
+func checkExportFile(export *exportpb.TemporaryExposureKeyExport) {
+	now := time.Now().UTC()
+	floor := model.IntervalNumber(now.Add(-14 * 24 * time.Hour))
```
gurayAlsac, line 66: nit: could define 14 and 16 (below) as self-describing constants up above?

mikehelmick: Will do in follow up

PR #863, opened 2020-08-12
allow publish to return a partial success/failure

### LINKED ISSUE #862: Keyset validation will drop all keys if start interval of any key is out of bounds
Our app is not using this project, but we ran into a similar issue as the one I'm reporting here and I thought I'd flag it.

### TL;DR
We've observed that mobile clients (iOS 13.5 confirmed) can return TEKs with a start interval that is way older than 14 days, mixed in with valid keys. The validation in https://github.com/google/exposure-notifications-server/blob/702c0796ecf6ab1a6816aaa2dfe422fed39fce9d/internal/publish/model/exposure_model.go#L299 seems to abort the transformation if any key is invalid, hereby rejecting the entire batch.

I'm not an expert at go or this project so I might be wrong 🙏 

**Expected behavior**
The invalid key is filtered out, but processing is continuing

mikehelmick:
Fixes #862

* If individual TEKs in a publish are invalid, allow the rest of the publish to commit
* introduce partial success error code

**Release Note**

```release-note
If just some of the exposure keys in a publish request are bad, drop those but accept the rest. Convey a partial success message back in the publish response.
```

### DIFF
--- a/internal/publish/model/exposure_model.go
+++ b/internal/publish/model/exposure_model.go
@@ -442,7 +442,7 @@ func (t *Transformer) TransformPublish(ctx context.Context, inData *verifyapi.Pu
 		upcaseRegions[i] = strings.ToUpper(r)
 	}
 
-	var transformErrors error
+	var transformErrors *multierror.Error
 	for i, exposureKey := range inData.Keys {
 		exposure, err := TransformExposureKey(exposureKey, inData.HealthAuthorityID, upcaseRegions, &settings)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


internal/publish/model/exposure_model.go
```diff
@@ -441,11 +442,13 @@ func (t *Transformer) TransformPublish(ctx context.Context, inData *verifyapi.Pu
 		upcaseRegions[i] = strings.ToUpper(r)
 	}
 
-	for _, exposureKey := range inData.Keys {
+	var transformErrors error
```
sethvargo, line 445:
```go
var transformErrors *multierror.Error
```

Then use 

```go
transformErrors.ErrorOrNil()
```

Issue #864, opened 2020-08-12
Revision token issues


mikehelmick:
2 potential issues discovered, but shouldn’t occurr in production environments

1) When validating a revision token, the token metadata for a TEK is checked against the database. The incoming publish metadata for a repeate TEK should also be checked agasint one of theres (either the token or the database). I.e. we need a==b==c, right now we’re checking a==b

2) When minting a new revision token on subsiquent publish. The existing token’s metadata should be taken first and then new uploads should be added second.

/kind bug
/assign @sethvargo

PR #865, opened 2020-08-12
Fix revision token validation

sethvargo:
Previously when validating a revision token, we would accept a full payload of TEK+metadata, and then immediately perform a database lookup on the TEK, ignoring any metadata. As a result, users could make requests with arbitrarily changed metadata, and the server would never throw a validation error. The server would not actually _accept_ those changed metadata values, it would just happily ignore them and use the persisted values in the database. This commit adds validation to ensure the incoming TEK metadata matches the database TEK metadata matches the revision token metadata. It also improves the logging around these cases.

In a similar vein, the only way the above is possible is due to a bug in how we construct revision token. This commit also inverts the logic to always include ALL previous revision tokens (and their metadata) FIRST in the revision token, then include any newly uploaded TEKs. This ensures that a partial upload of new keys would still return a revision token that could be used to revise any of the previous 14 days of keys, not just the uploaded ones.

Fixes GH-864

**Release Note**

```release-note
Fix revision token validation and ensure revision tokens are valid for the previous 14 days of keys, not just the keys uploaded in the revision request.
```

/assign @mikehelmick

### DIFF
--- a/internal/revision/revision.go
+++ b/internal/revision/revision.go
@@ -186,24 +186,28 @@ func buildTokenBufer(previous *pb.RevisionTokenData, eKeys []*model.Exposure) *p
 		RevisableKeys: make([]*pb.RevisableKey, 0, len(eKeys)),
 	}
 	got := make(map[string]struct{})
-	for _, k := range eKeys {
-		got[k.ExposureKeyBase64()] = struct{}{}
-		pbKey := pb.RevisableKey{
-			TemporaryExposureKey: make([]byte, len(k.ExposureKey)),
-			IntervalNumber:       k.IntervalNumber,
-			IntervalCount:        k.IntervalCount,
-		}
-		copy(pbKey.TemporaryExposureKey, k.ExposureKey)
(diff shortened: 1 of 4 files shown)


internal/revision/revision.go
```diff
@@ -186,24 +186,28 @@ func buildTokenBufer(previous *pb.RevisionTokenData, eKeys []*model.Exposure) *p
...
-			TemporaryExposureKey: make([]byte, len(k.ExposureKey)),
-			IntervalNumber:       k.IntervalNumber,
-			IntervalCount:        k.IntervalCount,
-		}
-		copy(pbKey.TemporaryExposureKey, k.ExposureKey)
-		tokenData.RevisableKeys = append(tokenData.RevisableKeys, &pbKey)
-	}
-	// Add in previous keys that weren't also in the new exposures
+
+	// Add in previous keys that weren't also in the new exposures. This needs to
```
mikehelmick, line 190: comment is wrong, it’s just all previous keys from the previous revision token.

sethvargo: DOne

PR #867, opened 2020-08-13
Improve wording

ithinkihaveacat:
">=14days ago" suggests 16 days would be valid. Reworded to eliminate the confusion between ">=" and "ago".


mikehelmick:
<jack sparrow>those are more like guidelines</jack sparrow> - everything is configurable

sethvargo:
/approve

but why is the rum gone?</jack-sparrow>

PR #868, opened 2020-08-14
Adding integration test scenario for exposure not yet verified

argetlam-cs:
Fixes #

## Proposed Changes

*  Restructured the integration test into test cases
*  Added test case for exposures not yet verified

**Release Note**

```release-note

```

### DIFF
--- a/internal/integration/integration_test.go
+++ b/internal/integration/integration_test.go
@@ -93,7 +93,7 @@ func TestIntegration(t *testing.T) {
 				if err == nil || !strings.Contains(err.Error(), tc.Error) {
 					t.Fatalf("expected error: %v, got: %v", tc.Error, err)
 				}
-				t.Skip()
+				return
 			}
 			if err != nil {
 				t.Fatal(err)
--- a/internal/integration/integration_test.go
+++ b/internal/integration/integration_test.go
@@ -42,8 +42,6 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-const INVALID_ERR_CODE = "health_authority_verification_certificate_invalid"
-
 func TestIntegration(t *testing.T) {
 	t.Parallel()
 
--- a/internal/integration/integration_test.go
+++ b/internal/integration/integration_test.go
@@ -71,7 +71,9 @@ func TestIntegration(t *testing.T) {
 	}
 
 	for _, tc := range cases {
+		tc := tc // Capture test case var for parallel runs
 		t.Run(tc.Name, func(t *testing.T) {
+			t.Parallel()
 			keys := util.GenerateExposureKeys(3, -1, false)
 
 			// Publish 3 keys
@@ -260,6 +262,7 @@ func TestIntegration(t *testing.T) {
 			keys = util.GenerateExposureKeys(3, -1, false)
 			payload.Keys = keys
 			jwtCfg.ExposureKeys = keys
+			jwtCfg.JWTWarp = tc.JWTWrap
 			verification, salt = testutil.IssueJWT(t, jwtCfg)
 			payload.VerificationPayload = verification
 			payload.HMACKey = salt
--- a/internal/integration/integration_test.go
+++ b/internal/integration/integration_test.go
@@ -49,6 +49,7 @@ func TestIntegration(t *testing.T) {
 		Name    string
 		JWTWrap time.Duration
 		Error   string
+		WantErr bool
 	}{
 		{
 			Name:    "exposure_verified",
@@ -57,10 +58,10 @@ func TestIntegration(t *testing.T) {
 		{
 			Name:    "exposure_not_verified",
 			JWTWrap: time.Hour,
-			Error:   "unable to validate diagnosis verification: Token used before issued",
+			Error:   `"error":"unable to validate diagnosis verification: Token used before issued","code":"health_authority_verification_certificate_invalid"`,
+			WantErr: true,
 		},
 	}
-
 	ctx := context.Background()
 	env, client, db, jwtCfg := NewTestServer(t, 2*time.Second)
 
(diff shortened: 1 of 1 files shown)


internal/integration/integration_test.go
```diff
@@ -53,316 +69,323 @@ func TestIntegration(t *testing.T) {
 		OnlyLocalProvenance: false,
 	}
 
-	keys := util.GenerateExposureKeys(3, -1, false)
+	for _, tc := range cases {
```
chaodaiG, line 72: For a table driven test, `t.Run` might be better, as it would make debug much easier

internal/integration/integration_test.go
```diff
@@ -53,316 +69,323 @@ func TestIntegration(t *testing.T) {
...
-		if got, want := len(exposures), 3; got != want {
-			t.Fatalf("expected %v to be %v", got, want)
+		jwtCfg.ExposureKeys = keys
+		jwtCfg.JWTWarp = tc.JWTWrap
+		verification, salt := testutil.IssueJWT(t, jwtCfg)
+		payload.VerificationPayload = verification
+		payload.HMACKey = salt
+		var revisionToken string
+		if resp, err := client.PublishKeys(payload); err != nil {
+			if !strings.Contains(err.Error(), tc.Error) {
```
chaodaiG, line 88: I would prefer an error comparison instead of strings.Contains

internal/integration/integration_test.go
```diff
@@ -53,316 +69,323 @@ func TestIntegration(t *testing.T) {
...
 		}
-		if got, want := len(exposures), 3; got != want {
-			t.Fatalf("expected %v to be %v", got, want)
+		jwtCfg.ExposureKeys = keys
+		jwtCfg.JWTWarp = tc.JWTWrap
+		verification, salt := testutil.IssueJWT(t, jwtCfg)
+		payload.VerificationPayload = verification
+		payload.HMACKey = salt
+		var revisionToken string
+		if resp, err := client.PublishKeys(payload); err != nil {
```
chaodaiG, line 87: This doesn't cover the case where error is expected but `PublishKeys` return nil

internal/integration/integration_test.go
sethvargo: Also drop a t.Parallel() in the t.Run() so the subtests run in parallel too

internal/integration/integration_test.go
sethvargo: Yes, please use `errors.Is` if possible.

internal/integration/integration_test.go
```diff
@@ -53,316 +69,323 @@ func TestIntegration(t *testing.T) {
...
-		if err := client.ExportBatches(); err != nil {
-			return err
+		// Assert there are 3 exposures in the database
+		{
+			exposures, err := getExposures(db, criteria)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if got, want := len(exposures), 3; got != want {
+				t.Fatalf("expected %v to be %v", got, want)
```
sethvargo, line 107: %#v

internal/integration/integration_test.go
argetlam-cs: For some reason the error comparison always fails

This is what I get
```
expected error: failed to POST /publish: response was not 200 OK: {"error":"unable to validate diagnosis verification: Token used before issued","code":"health_authority_verification_certificate_invalid"}: , got: failed to POST /publish: response was not 200 OK: {"error":"unable to validate diagnosis verification: Token used before issued","code":"health_authority_verification_certificate_invalid"}: 
```

I am guessing it's because the way the error is generated by the client.go file.
I am not sure what exactly is happening here

chaodaiG: `client.go` uses `%w` wrapping error here https://github.com/google/exposure-notifications-server/blob/c5f5cf67733a61676de4d046a39df40ee2c91dd1/internal/integration/client.go#L47, can you try using `errors.Is` with wanted error below?
```
wantErr := errors.New("unable to validate diagnosis verification: Token used before issued")
```

internal/integration/integration_test.go
```diff
@@ -53,316 +69,323 @@ func TestIntegration(t *testing.T) {
...
+
+			// Find the latest file in the index
+			lines := strings.Split(string(index), "\n")
+			latest := ""
+			for _, entry := range lines {
+				if strings.HasSuffix(entry, "zip") {
+					if latest == "" {
 						latest = entry
+					} else {
+						if entry > latest {
```
MushuEE, line 141: I know this was carried over from prior, but if "" is the smallest lexicographically. Is there really a need for the  `latest == ""` condition? I think `if entry > latest` covers all cases

internal/integration/integration_test.go
argetlam-cs: I have tried that too.
Tried again. Still the same result.

```
wantErr := fmt.Errorf(`response was not 200 OK: {"error":"unable to validate diagnosis verification: Token used before issued","code":"health_authority_verification_certificate_invalid"}`)
				if !errors.Is(err, wantErr) {
					t.Fatalf("expected error: %s, got: %s,", err, wantErr)
				}
```

argetlam-cs: https://stackoverflow.com/questions/62441960/golang-error-wrap-unwrap-type-checking-with-errors-is

Looks like that errors.Is does an actual instance match and this will definitely not work in this case.

Shall we stay with the error message check or are there other ways to do this?

argetlam-cs: the other option could be having an exported var in client.go that stores the error and we use that to compare.
But I don't like this option

internal/integration/integration_test.go
argetlam-cs: done

internal/integration/integration_test.go
argetlam-cs: done

internal/integration/integration_test.go
argetlam-cs: Done

internal/integration/integration_test.go
argetlam-cs: Done

internal/integration/integration_test.go
sethvargo: Why is that a problem?

argetlam-cs: Added it in the latest commit. Somehow I had misinterpreted the error I was seeing

internal/integration/integration_test.go
```diff
@@ -53,316 +70,327 @@ func TestIntegration(t *testing.T) {
...
+			payload.VerificationPayload = verification
+			payload.HMACKey = salt
+			var revisionToken string
+			resp, err := client.PublishKeys(payload)
+
+			if tc.WantErr {
+				if err == nil || !strings.Contains(err.Error(), tc.Error) {
+					t.Fatalf("expected error: %v, got: %v", tc.Error, err)
+				}
+				t.Skip()
```
chaodaiG, line 96: Not sure, but can you use `return` instead of `t.Skip`?

argetlam-cs: Done

PR #894, opened 2020-08-20
Fix Stackdriver Managed Resource

icco:
## Proposed Changes

* Attempts to fix the errors from stackdriver by giving a better task_id
* adds lots of debug logging
* Fixes a spelling error.
* Adds context to stackdriver requests instead of the default context.Background().

**Release Note**

```release-note
 - Metrics in stackdriver will now be under the correct resource type if running on GKE.
 - Stackdriver should no longer issue errors every minute.
```


pkg/observability/resource.go
```diff
@@ -15,56 +15,103 @@
...
 	}
+
 	labels["namespace"] = c.Namespace
 
+	// Are on Cloud Run Managed?
+	//
+	// https://cloud.google.com/monitoring/api/resources#tag_cloud_run_revision
+	// https://cloud.google.com/run/docs/reference/container-contract#env-vars
+	if c.Service != "" && c.Revision != "" {
+		resource = "cloud_run_revision"
```
mikehelmick, line 98: Did you test this? I don't think we can write to that namespace

pkg/observability/resource.go
```diff
@@ -15,56 +15,103 @@
...
 		labels["location"] = "unknown"
 	}
+
 	labels["namespace"] = c.Namespace
 
+	// Are on Cloud Run Managed?
+	//
+	// https://cloud.google.com/monitoring/api/resources#tag_cloud_run_revision
+	// https://cloud.google.com/run/docs/reference/container-contract#env-vars
+	if c.Service != "" && c.Revision != "" {
```
mikehelmick, line 97: these will also be set under knative as well

icco:
I still don't have a solid test environment for this @mikehelmick, but this is the results of talking w/ Monarch SRE. The main thing I want here is the logging to see what's going on, but the changes to task_id might help (I'm skeptical tho).

mikehelmick:
> I still don't have a solid test environment for this @mikehelmick, but this is the results of talking w/ Monarch SRE. The main thing I want here is the logging to see what's going on, but the changes to task_id might help (I'm skeptical tho).

I was running locally when I was testing it out. 
We can merge this in and make changes later if it's not quite right

pkg/observability/resource.go
```diff
@@ -15,45 +15,83 @@
...
+	if labels["task_id"] == "" {
+		iid, err := metadata.InstanceID()
+		if err == nil {
+			labels["task_id"] = iid
+		}
+	}
+
+	// Worse case task_id
+	if labels["task_id"] == "" {
 		labels["task_id"] = base64.StdEncoding.EncodeToString(uuid.NodeID())
```
mikehelmick, line 95: let's change this to use  https://godoc.org/github.com/google/uuid#New as the fallback instead of the NodeID

icco: done.

icco:
> > I still don't have a solid test environment for this @mikehelmick, but this is the results of talking w/ Monarch SRE. The main thing I want here is the logging to see what's going on, but the changes to task_id might help (I'm skeptical tho).
> 
> I was running locally when I was testing it out.
> We can merge this in and make changes later if it's not quite right

oh, I didn't think it'd work locally. Interesting. Well lets merge, but I'll keep that in mind.

PR #912, opened 2020-08-24
Make Terraform more configurabler

sethvargo:
Make Terraform more configurabler

**Release Note**

```release-note
Make Terraform more configurabler
```

### DIFF
--- a/terraform/database.tf
+++ b/terraform/database.tf
@@ -12,18 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-resource "random_string" "db-name" {
-  length  = 5
-  special = false
-  number  = false
-  upper   = false
-}
-
 resource "google_sql_database_instance" "db-inst" {
   project          = data.google_project.project.project_id
   region           = var.db_location
   database_version = "POSTGRES_11"
-  name             = "en-${random_string.db-name.result}"
+  name             = var.db_name
 
   settings {
     tier              = var.cloudsql_tier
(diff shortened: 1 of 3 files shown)


terraform/database.tf
```diff
@@ -12,18 +12,11 @@
...
-  number  = false
-  upper   = false
-}
-
 resource "google_sql_database_instance" "db-inst" {
   project          = data.google_project.project.project_id
   region           = var.db_location
   database_version = "POSTGRES_11"
-  name             = "en-${random_string.db-name.result}"
+  name             = var.db_name
```
sethvargo, line 19: Note that "name" is ignored on changes, so this won't cause a re-create

terraform/vars.tf
```diff
@@ -42,6 +48,18 @@ variable "kms_location" {
   default = "us-central1"
 }
 
+# Name of the key ring for export signing keys.
+variable "kms_export_signing_key_ring_name" {
```
chaodaiG, line 52: Do we also want to update their usage in this PR? https://github.com/google/exposure-notifications-server/blob/c1975364369fe6d8b02485656bd3b08c9549d62f/terraform/key_management.tf#L17

terraform/database.tf
icco: Is that documented anywhere? Didn't see it in https://www.terraform.io/docs/providers/google/r/sql_database_instance.html#argument-reference

terraform/vars.tf
sethvargo: Done

chaodaiG:
lgtm from my end, so
/lgtm
/hold
In case you want to address @icco 's comment

icco:
Given it was currently a random string, doesn't matter beyond my curiosity.

/lgtm

PR #924, opened 2020-08-27
Ignore locally built binaries.

yegle:


sethvargo:
Hi @yegle 

It's unclear why we need to ignore these. Can you just use `go run` instead?

yegle:
Without this, it's a bit annoying to have to remove the binary before running the build script.

sethvargo:
I'm still confused why you're building the binary though?

yegle:
It's probably just me that will prefer running a built binary than using `go run`.

Anyway, I think this is cosmetic change. I'll revert and switch to use `go run`.

Issue #915, opened 2020-08-24
keys.HashiCorpVault should implement keys.SigningKeyManagement


mikehelmick:
/kind enhancement

PR #926, opened 2020-08-27
Add signing key manager to Vault

sethvargo:
Fixes https://github.com/google/exposure-notifications-server/issues/915

⚠️ Vault doesn't support destroying a single key version, so it returns an error.

**Release Note**

```release-note
Add support for SigningKeyManager interface with Vault
```

### DIFF
--- a/pkg/keys/hashicorp_vault.go
+++ b/pkg/keys/hashicorp_vault.go
@@ -18,20 +18,23 @@ import (
 	"context"
 	"crypto"
 	"crypto/ecdsa"
-	"crypto/x509"
 	"encoding/base64"
-	"encoding/pem"
 	"fmt"
 	"io"
+	"sort"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/google/exposure-notifications-server/pkg/base64util"
+	"github.com/mitchellh/mapstructure"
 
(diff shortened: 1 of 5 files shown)


pkg/keys/hashicorp_vault.go
```diff
@@ -72,6 +75,144 @@ func NewHCValueKeyID(keyID string) (*HCValueKeyID, error) {
...
 }
 
+type readKeyResponse struct {
+	Versions map[string]struct {
+		CreationTime time.Time `ms:"creation_time"`
+		PublicKeyPEM string    `ms:"public_key"`
+	} `ms:"keys"`
+}
+
+type vaultKeyVersion struct {
```
mikehelmick, line 85: compile time cast assert?

mikehelmick: i know it’s not necessary - but we normally do it.. so up to you

sethvargo: I didn't originally add it because the return type from the method enforces it at compile time. Added.

mikehelmick: yeah - just trying to be internally consistent with what we've done elsewhere. Thanks!

PR #931, opened 2020-08-29
Add trace logging function

icco:
## Proposed Changes

* Adds a function to add trace information to log lines, as requested in https://github.com/google/exposure-notifications-verification-server/issues/416

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/logging/logger.go
+++ b/pkg/logging/logger.go
@@ -148,6 +148,26 @@ func levelEncoder() zapcore.LevelEncoder {
 	}
 }
 
+// TraceFromContext adds the correct Stackdriver trace fields.
+//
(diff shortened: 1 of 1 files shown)


pkg/logging/logger.go
```diff
@@ -148,6 +150,26 @@ func levelEncoder() zapcore.LevelEncoder {
...
+	span := trace.FromContext(ctx)
+
+	if span == nil {
+		return nil
+	}
+
+	sc := span.SpanContext()
+
+	return []zap.Field{
+		// TODO(icco): Figure out how to add project ID to this.
```
sethvargo, line 166: Primitively, you could just call os.Getenv for now

Issue #929, opened 2020-08-28
Include git SHA/tag in builds and logs


sethvargo:
When debugging things, it'd be helpful to know _which_ version of the code a log corresponds to. We can use build tags and the namespaced logger to always include this information.

/assign

mikehelmick:
+1 - would be great.

PR #937, opened 2020-09-01
New build process, yay!

sethvargo:
Fixes GH-929

**Release Note**

```release-note
Switch to new build process, bundle migrations with deployment
```

### DIFF
--- a/builders/deploy.yaml
+++ b/builders/deploy.yaml
@@ -18,22 +18,195 @@
 
 substitutions:
   _REGION:
-  _SERVICE:
+  _REPO: 'github.com/google/exposure-notifications-server'
   _TAG:
 
 steps:
-- id: 'deploy'
-  name: 'gcr.io/google.com/cloudsdktool/cloud-sdk:303.0.0-alpine'
+#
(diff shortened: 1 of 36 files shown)


builders/deploy.yaml
```diff
@@ -18,22 +18,195 @@
...
+      --quiet \
+      --project "${PROJECT_ID}" \
+      --platform "managed" \
+      --region "${_REGION}" \
+      --image "gcr.io/${PROJECT_ID}/${_REPO}/cleanup-exposure:${_TAG}" \
+      --no-traffic
+  waitFor:
+  - '-'
+
+#
```
sethvargo, line 67: Maybe we don't want to deploy this all the time? We don't have an easy way to do conditionals though...

cmd/migrate/main.go
```diff
@@ -0,0 +1,101 @@
+// Copyright 2020 Google LLC
```
sethvargo, line 1: It's a bit of code, but this lets us build migrations into a container, same as the verification server.

PR #943, opened 2020-09-01
Fix dropping all labels

icco:
## Proposed Changes

* We were accidentally dropping all stackdriver labels due to a bug.

**Release Note**

```release-note
Reenables sending of stackdriver metrics.
```


sethvargo:
/lgtm
/approve

Do we need to bump and cut a release of the verification server too then?

icco:
> /lgtm
> /approve
> 
> Do we need to bump and cut a release of the verification server too then?

sadly yes.

PR #949, opened 2020-09-02
Simplify Stackdriver Code

icco:
## Proposed Changes

* Remove any magic and be very explicit about where our labels are coming from.
* Query region instead of zone (https://github.com/googleapis/google-cloud-go/issues/2810) 
* Improved debugging by exporting fields

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/observability/resource.go
+++ b/pkg/observability/resource.go
@@ -37,6 +37,9 @@ type stackdriverMonitoredResource struct {
 // required labels filled out. This needs to be the correct resource type so we
 // can compared the default stackdriver metrics with the custom metrics we're
 // generating.
+//
+// NOTE: This code is focused on support GCP Cloud Run Managed. If you are
+// runniing in a different environment, you may see weird results.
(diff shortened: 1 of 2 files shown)


icco:
I've been testing this in encv-test, and it doesn't fix the issue, but it does make debugging easier.

pkg/observability/resource.go
```diff
@@ -30,86 +29,61 @@ var (
...
+	Labels   map[string]string
 }
 
 // NewStackdriverMonitoredResource returns a monitored resource with the
 // required labels filled out. This needs to be the correct resource type so we
 // can compared the default stackdriver metrics with the custom metrics we're
 // generating.
+//
+// NOTE: This code is focused on support GCP Cloud Run Managed. If you are
+// runniing in a different environment, you may see weird results.
```
sethvargo, line 42:
```suggestion
// running in a different environment, you may see weird results.
```

PR #951, opened 2020-09-02
Fix some stackdriver bugs

icco:
## Proposed Changes

* Continuing on #949, this fixes the missing ochttp metrics.
* Fixes location, which comes back in a bad format from the metadata server.
* Increases the metric reporting time (doesn't fix the errors, but makes them less frequent)

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/observability/resource.go
+++ b/pkg/observability/resource.go
@@ -37,6 +37,9 @@ type stackdriverMonitoredResource struct {
 // required labels filled out. This needs to be the correct resource type so we
 // can compared the default stackdriver metrics with the custom metrics we're
 // generating.
+//
+// NOTE: This code is focused on support GCP Cloud Run Managed. If you are
+// runniing in a different environment, you may see weird results.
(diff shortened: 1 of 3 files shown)


pkg/observability/resource.go
```diff
@@ -72,6 +73,12 @@ func NewStackdriverMonitoredResource(ctx context.Context, c *StackdriverConfig)
 	}
 	labels["location"] = region
 
+	// Metadata often returns the following: "projects/111111111111/regions/us-central1"
+	pieces := strings.Split(region, "/")
```
sethvargo, line 77: This feels fragile and could have subtle bugs if that format changes

icco: Yeah. I'm not sure what to do otherwise. I feel like a regex would be more fragile?

sethvargo: SplitN and checking against the other lengths? Or maybe returning/logging an error on mismatch?

icco: good idea, added error message.

mikehelmick (approved):

PR #952, opened 2020-09-02
add base64 decode tool for inspecting hmac candidates

mikehelmick:
## Proposed Changes

* small tool to print info about b64 encoded hmac values

**Release Note**

```release-note
NONE
```

### DIFF
--- a/tools/b64decode/main.go
+++ b/tools/b64decode/main.go
@@ -1,0 +1,40 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


tools/b64decode/main.go
```diff
@@ -0,0 +1,40 @@
...
+	"log"
+	"os"
+
+	"github.com/google/exposure-notifications-server/pkg/base64util"
+)
+
+// Small to base64 decode data and print some info
+func main() {
+	if len(os.Args) != 2 {
+		log.Fatal("requires 1 argument which will be interpreted as a hex byte array")
```
sethvargo, line 28: This isn't true? It will be interpreted as a base64-encoded byte array, right?

mikehelmick: copy paste

sethvargo (approved):
/lgtm

Issue #914, opened 2020-08-24
keys.InMemory should implement keys.SigningKeyManagement


mikehelmick:
/kind enhancement

PR #957, opened 2020-09-07
Switch InMemory KMS to filesystem based

sethvargo:
The challenge with the in-memory implementation is that keys go away if the process dies. This doesn't mesh well with the verification server, which creates/destroys keys and requires persistence.

This also exposes test helpers for managing keys.

This added 1s to the overall test suite time, so I'm really not concerned with the additional I/O.

Fixes GH-914

**Release Note**

```release-note
BREAKING: Replace in-memory key manager with filesystem key manager, dropped support for in-memory key manager.
```


sethvargo:
Oh, this fixes GH-914

pkg/keys/config.go
```diff
@@ -22,7 +22,7 @@ const (
 	KeyManagerTypeAzureKeyVault  KeyManagerType = "AZURE_KEY_VAULT"
 	KeyManagerTypeGoogleCloudKMS KeyManagerType = "GOOGLE_CLOUD_KMS"
 	KeyManagerTypeHashiCorpVault KeyManagerType = "HASHICORP_VAULT"
-	KeyManagerTypeInMemory       KeyManagerType = "IN_MEMORY"
+	KeyManagerTypeFilesystem     KeyManagerType = "FILESYSTEM"
```
icco, line 25: Since this is fully removing support for in-memory, I mentioned that in the release notes in case anyone was depending on it in some weird way.

pkg/keys/filesystem.go
```diff
@@ -0,0 +1,410 @@
...
+var _ SigningKeyManager = (*Filesystem)(nil)
+
+// Filesystem is a key manager that uses the filesystem to store and retrieve
+// keys. It should only be used for local development and testing.
+type Filesystem struct {
+	root string
+	mu   sync.RWMutex
+}
+
+// NewFilesystem creates a new, local, in memory KeyManager.
```
icco, line 51: fix comment

pkg/keys/filesystem.go
```diff
@@ -0,0 +1,410 @@
...
+func NewFilesystem(ctx context.Context, root string) (*Filesystem, error) {
+	if root != "" {
```
icco, line 53: If root is empty string, is that an error?

pkg/keys/filesystem.go
```diff
@@ -0,0 +1,410 @@
...
+		if latest == nil {
+			latest = info
+			continue
+		}
+		if info.Name() > latest.Name() {
+			latest = info
+		}
+	}
+	if latest == nil {
+		return nil, fmt.Errorf("key %s does not exist", keyID)
```
icco, line 113: nit: change %s to %q in case keyid has spaces in it for log readability.

pkg/keys/filesystem_test.go
```diff
@@ -0,0 +1,324 @@
...
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/x509"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
```
icco, line 29: plz add tests for filesystem creation

pkg/keys/config.go
sethvargo: ack

pkg/keys/filesystem.go
sethvargo: No, will document.

pkg/keys/filesystem_test.go
sethvargo: Done

pkg/keys/filesystem.go
sethvargo: Done

pkg/keys/filesystem.go
sethvargo: Done

icco (approved):
/lgtm

PR #959, opened 2020-09-08
implement random padding of publish responses

mikehelmick:
## Proposed Changes

* pad publish responses (error and success)

**Release Note**

```release-note
Start populating "padding" field on the publish responses. This field has been in the API for several releases, but wasn't populated before.
```


sethvargo (approved):
/lgtm

/hold pending sign off from the client team

Issue #884, opened 2020-08-18
generate service must account for revision tokens


mikehelmick:
### TL;DR

The generate service attempts to revise keys, but that is always failing since it doesn't handle revision tokens.

PR #963, opened 2020-09-08
Use revision tokens in generate service

sethvargo:
Fixes GH-884

This involved a bit of a refactor to support testability.

**Release Note**

```release-note
Support revision tokens in generate service
```

/assign @mikehelmick

### DIFF
--- a/internal/generate/handler.go
+++ b/internal/generate/handler.go
@@ -23,9 +23,11 @@ import (
 	"time"
 
 	"go.opencensus.io/trace"
+	"go.uber.org/zap"
 
-	"github.com/google/exposure-notifications-server/internal/publish/database"
-	"github.com/google/exposure-notifications-server/internal/publish/model"
+	"github.com/google/exposure-notifications-server/internal/pb"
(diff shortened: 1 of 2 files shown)


internal/generate/handler.go
```diff
@@ -34,166 +36,160 @@ import (
...
+	}
+
+	w.WriteHeader(http.StatusOK)
+	fmt.Fprint(w, "successfully generated exposure keys")
+}
+
+func (h *generateHandler) generate(ctx context.Context, regions []string) error {
+	logger := h.logger.Named("generate")
+
+	if h.config.KeysPerExposure < 2 {
```
mikehelmick, line 99: why 2?

sethvargo: Line 141:

```go
intervalIdx, err := util.RandomInt(len(publish.Keys) - 1)
```

That gets a random index in the list, but panics if the number of keys is 1 (and therefore tries to generate a random int between 0 and 0).

mikehelmick: worth a comment in the code then

Issue #893, opened 2020-08-19
Additional partial success support


mikehelmick:
Allow for partial success in this scenario.
Assuming 2 regions are on a shared server.

* User uploads likely diagnosis with AppA, and gets revision token indicating as such
* User changes regions and installs a different app, AppB
* 5 days later, user gets a positive test in their new region. Currently key upload would fail because AppA doesn't hold the revision token for days -6 to -14
  * In this case, accept the 5 keys that haven't been seen before. Issue a new revision token for just those

/kind enhancement

sethvargo:
This all makes sense if the third bullet meant to say "fail because **AppB** doesn't hold the revision token for days -6 to -14". Is that correct @mikehelmick?

mikehelmick:
The 3rd large bullet describes current behavior -the sub-bullet describes desired behavior

PR #964, opened 2020-09-08
Add support for partial revisions

sethvargo:
Fixes GH-893

**Release Note**

```release-note
Add configurable support for partial revisions
```

/assign @mikehelmick

### DIFF
--- a/internal/publish/config.go
+++ b/internal/publish/config.go
@@ -61,6 +61,12 @@ type Config struct {
 
 	RevisionKeyCacheDuration time.Duration `env:"REVISION_KEY_CACHE_DURATION, default=1m"`
 
+	// AllowPartialRevisions permits uploading multiple exposure keys with a
+	// revision token where only a subset of the keys are in the token. In that
+	// case, only the incoming exposure keys that match the revision token are
+	// uploaded and the remainder are discarded.
+	AllowPartialRevisions bool `env:"ALLOW_PARTIAL_REVISIONS, default=false"`
+
 	// API Versions.
 	EnableV1Alpha1API bool `env:"ENABLE_V1ALPHA1_API, default=true"`
 
--- a/internal/publish/database/exposure.go
+++ b/internal/publish/database/exposure.go
@@ -367,17 +367,38 @@ func executeReviseExposure(ctx context.Context, tx pgx.Tx, stmtName string, exp 
 	return nil
 }
 
-// InsertAndReviseExposures transactionally revises and inserts a set of keys as necessary.
-func (db *PublishDB) InsertAndReviseExposures(ctx context.Context, incoming []*model.Exposure, token *pb.RevisionTokenData, tokenRequired bool) (int, error) {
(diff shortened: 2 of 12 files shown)


internal/publish/config.go
```diff
@@ -61,6 +61,12 @@ type Config struct {
 
 	RevisionKeyCacheDuration time.Duration `env:"REVISION_KEY_CACHE_DURATION, default=1m"`
 
+	// AllowPartialRevisions permits uploading multiple exposure keys with a
+	// revision token where only a subset of the keys are in the token. In that
+	// case, only the incoming exposure keys that match the revision token are
+	// uploaded and the remainder are discarded.
+	AllowPartialRevisions bool `env:"ALLOW_PARTIAL_REVISIONS, default=false"`
```
sethvargo, line 68: I think this makes more sense as a key server config instead of a per-app config, but we could easily make it per-app too. Curious your thoughts @mikehelmick.

internal/publish/database/exposure.go
```diff
@@ -367,17 +367,38 @@ func executeReviseExposure(ctx context.Context, tx pgx.Tx, stmtName string, exp
 	return nil
 }
 
-// InsertAndReviseExposures transactionally revises and inserts a set of keys as necessary.
-func (db *PublishDB) InsertAndReviseExposures(ctx context.Context, incoming []*model.Exposure, token *pb.RevisionTokenData, tokenRequired bool) (int, error) {
-	logger := logging.FromContext(ctx)
+// InsertAndReviseExposuresRequest is used as input to InsertAndReviseExposures.
+type InsertAndReviseExposuresRequest struct {
```
sethvargo, line 371: The method signature was getting too long and difficult to understand what params meant what.

internal/publish/config.go
mikehelmick: I think server level config is fine.

internal/publish/publish.go
```diff
@@ -322,7 +322,13 @@ func (h *PublishHandler) process(ctx context.Context, data *verifyapi.Publish, b
 			metric: "publish-transform-fail", count: 1}
 	}
 
-	n, err := h.database.InsertAndReviseExposures(ctx, exposures, token, !appConfig.BypassRevisionToken)
+	n, err := h.database.InsertAndReviseExposures(ctx, &database.InsertAndReviseExposuresRequest{
+		Incoming: exposures,
```
mikehelmick, line 326: something is missing here - the new revision token will still merge the old revision token (if there is one) and all of the exposures presented.

Thus giving the user a revision token that would let them later revise the keys.

We need to only issue the revision token for keys they successfully inserted.

mikehelmick (changes requested):

internal/publish/publish.go
sethvargo: Okay, this was a bit more involved that I originally thought, because I had to change the return type, but this is done now.

mikehelmick (approved):

Issue #930, opened 2020-08-28
Check routing


mikehelmick:
We have a report that on the exposure service

`/` serves v1alpha1
`/v1/publish` serves v1
`/v1/publish/` serves v1alpha1

/kind bug

icco:
What's the proper functionality here?

mikehelmick:
I think that `/v1/publish/` should 404

sethvargo:
I've confirmed this behavior. Working on a fix.

/assign

nmiodice:
*Not asking for a change here, just a curious reader*

Is there a reason to return 404 instead of publish the `v1` API @mikehelmick?

mikehelmick:
This totally could go either way. I have a personal preference for being exact on things like this.

PR #967, opened 2020-09-09
Add x-api-version header in API responses

sethvargo:
This is just a hint to clients to know which API version they invoked.

Did this as part of investigating GH-930 and thought it might be helpful to have in prod.

**Release Note**

```release-note
Add `x-api-version` response header to exposure service responses
```


icco:
/lgtm

I was thinking of making a `/version` handler to return the build version and supported API versions. This is a nice solution as well.

PR #976, opened 2020-09-10
Run terraform smoke test with dynamic projects

chaodaiG:
Use projects pool manager boskos for managing GCP projects

Boskos not deployed yet
/cc chaodaiG
/assign chaodaiG

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@
 *.terraform/
 *.tfvars
 terraform/state.tf
+terraform-e2e/state.tf
 
 # Test stuff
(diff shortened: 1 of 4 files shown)


chaodaiG:
Ready for review

/cc @sethvargo @icco

sethvargo (approved):
/lgtm

Gonna assume it's okay for those URLs to be public?

chaodaiG:
Yeah, that URL is an in cluster endpoint, there is no way it could be reached from outside of prow build cluster

PR #978, opened 2020-09-10
Install boskosctl, the GCP projects pool manager in presubmit test image

chaodaiG:
boskosctl is used in e2e tests for dynamically allocating GCP projects, as well as managing projects life cycles


sethvargo (approved):
/lgtm

PR #981, opened 2020-09-11
Fix perf test failure

chaodaiG:
Performance test failed since https://github.com/google/exposure-notifications-server/pull/964

See performance test runs here: https://testgrid.k8s.io/googleoss-en-server#performance

/cc @sethvargo


icco (approved):
/lgtm

PR #983, opened 2020-09-11
Sort exposures in unit test to make the result deterministic

chaodaiG:
This test failed randomly with out of order problem https://oss-prow.knative.dev/view/gs/oss-prow/pr-logs/pull/google_exposure-notifications-server/981/pull-en-server-release-unit/1304300664213475328

/assign @sethvargo

### DIFF
--- a/internal/publish/database/exposure_test.go
+++ b/internal/publish/database/exposure_test.go
@@ -756,6 +756,10 @@ func TestIterateExposuresCursor(t *testing.T) {
 	if err == nil {
 		t.Fatal("expected error")
 	}
+	// IterateExposures doesn't guarantee the order, sort to make this test deterministic
+	sort.Slice(seen, func(i, j int) bool {
+		return seen[i].IntervalNumber < seen[j].IntervalNumber
+	})
 	if diff := cmp.Diff(exposures[:2], seen, ignoreUnexportedExposure); diff != "" {
 		t.Fatalf("exposures mismatch (-want, +got):\n%s", diff)
 	}
@@ -770,6 +774,9 @@ func TestIterateExposuresCursor(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	sort.Slice(seen, func(i, j int) bool {
+		return seen[i].IntervalNumber < seen[j].IntervalNumber
+	})
 	if diff := cmp.Diff(exposures[2:], seen, ignoreUnexportedExposure); diff != "" {
 		t.Fatalf("exposures mismatch (-want, +got):\n%s", diff)
 	}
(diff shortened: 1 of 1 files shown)


internal/publish/database/exposure_test.go
```diff
@@ -756,6 +756,10 @@ func TestIterateExposuresCursor(t *testing.T) {
 	if err == nil {
 		t.Fatal("expected error")
 	}
+	// IterateExposures doesn't guarantee the order, sort to make this test deterministic
+	sort.Slice(seen, func(i, j int) bool {
+		return seen[i].IntervalNumber < seen[j].IntervalNumber
```
sethvargo, line 761: Do you want to sort by b64 TEK, then IntervalNumber, then rolling period?

chaodaiG: For this case I think IntervalNumber is sufficient. Also a QQ: how do we compare b64 TEK?

sethvargo: It's just a string, so `<` or `>`.

chaodaiG: I thought they are generated from random hashes, are they guaranteed to be ordered?

sethvargo: They are random, but when you're sorting two lists to compare them, you just sort them like strings.

sethvargo: Yea

internal/publish/database/exposure_test.go
```diff
@@ -756,6 +756,10 @@ func TestIterateExposuresCursor(t *testing.T) {
 	if err == nil {
 		t.Fatal("expected error")
 	}
+	// IterateExposures doesn't guarantee the order, sort to make this test deterministic
```
mikehelmick, line 759: the preferred way to do this is to pass a sort function to cmp.Diff as a cmpopt

https://godoc.org/github.com/google/go-cmp/cmp/cmpopts#SortSlices

chaodaiG: Good to learn, done.

internal/publish/database/exposure_test.go
```diff
@@ -756,7 +756,11 @@ func TestIterateExposuresCursor(t *testing.T) {
 	if err == nil {
 		t.Fatal("expected error")
 	}
-	if diff := cmp.Diff(exposures[:2], seen, ignoreUnexportedExposure); diff != "" {
+	// IterateExposures doesn't guarantee the order, sort to make this test deterministic
+	sortExposureFunc := func(i, j int) bool {
```
mikehelmick, line 760: I think this needs to take in two `*model.Exposure` params and return which true if the first is "less than" the other

chaodaiG: That's right, fixed.

Surprisingly the previous one worked...

mikehelmick (approved):

PR #991, opened 2020-09-15
Provide a way of conditionally skipping deploy certain services during upgrade

chaodaiG:
This is good for:
- Conditionally skip deploying debugger in prod server
- Running e2e test that's based on top of terraform deployment

/assign @sethvargo


chaodaiG:
This solution is not ideal but the best I can think of. The current blocker for me is deploying services based on top of terraform deployment, which failed due to some services are not deployed by terraform, and Cloud Run doesn't allow `no-traffic` on newly created services. The alternative solution to unblock this is deploying these services with terraform, which is ok with me too

builders/deploy.yaml
```diff
@@ -20,6 +20,15 @@ substitutions:
   _REGION:
   _REPO: 'github.com/google/exposure-notifications-server'
   _TAG:
+  _SKIP_DEPLOY_CLEANUP_EXPOSURE: false
```
sethvargo, line 23: This feels... like a lot of maintenance. Can we make this a comma-separated list instead and then do a substring match?

chaodaiG: That would be a good idea, working on it

mikehelmick:
Skipping some services is potentially unsafe. Especially if there are database migrations.

I'm not sold on including this functionality.

chaodaiG:
> Skipping some services is potentially unsafe. Especially if there are database migrations.
> 
> I'm not sold on including this functionality.

If it's desired to deploy the same set of services everywhere then I can drop this PR. So for running `build -> deploy -> promote` after terraform deployment, the other solution is also deploying debugger https://github.com/google/exposure-notifications-server/blob/1c4e7c109678cc1e7467a3c5af7b984de87270e3/builders/deploy.yaml#L70 with terraform. Does this make sense to you?

chaodaiG:
> > Skipping some services is potentially unsafe. Especially if there are database migrations.
> > I'm not sold on including this functionality.
> 
> If it's desired to deploy the same set of services everywhere then I can drop this PR. So for running `build -> deploy -> promote` after terraform deployment, the other solution is also deploying debugger
> 
> https://github.com/google/exposure-notifications-server/blob/1c4e7c109678cc1e7467a3c5af7b984de87270e3/builders/deploy.yaml#L70
> 
> with terraform. Does this make sense to you?

Even easier, how about removing `--no-traffic` from `debugger` deployment https://github.com/google/exposure-notifications-server/blob/4269176955e3941d72e2413c7c35478815dd760b/builders/deploy.yaml#L84?  @mikehelmick @sethvargo , WDYT?

chaodaiG:
Actually nvm, found how to deploy debugger with terraform

/close

PR #993, opened 2020-09-15
add tool to verify ASN1 encoded signatures of digests with a provided…

mikehelmick:
… ECDSA PEM public key

## Proposed Changes

* command line tool that accepts a SHA256 digest, a signature, and a file containing a public key and verifies the signature.

**Release Note**

```release-note
Utility to verify signatures.
```


sethvargo (approved):
/lgtm

PR #997, opened 2020-09-16
Fix terraform trap destroy

chaodaiG:
The best effort terraform destroy is not executed if deploy failed, move trap above deploy should fix this problem


sethvargo (approved):
/lgtm

PR #998, opened 2020-09-16
Add key synchronization with a JWKS endpoint.

### LINKED ISSUE #911: Accept JWKS URIs for Health Authority Keys
### TL;DR

After https://github.com/google/exposure-notifications-verification-server/issues/93 is complete.

Implement the ability for health authority signing keys to be accepted via JWKS URI instead of through configuration in the admin console / database 

### Design

**Proposal**

* Add another field to `HealthAuthority` for the JWKS URI
* Add a periodic refresh for all HealthAuthorities that have this enabled (non-empty URI)
  * This should automatically managed the HealthAuthorityKeys in the database (creating and expiring as needed)

Certificate validation won't need any changes as long as the public keys and key IDs are saved to the database.

**Alternatives considered**

* hitting the JWKS uri in the request path - creates a hard dependency on the verification server actually being available

/kind enhancement
/assign @jeremyfaller

jeremyfaller:
This change introduces a new routine that runs periodically to update
the HealthAuthority keys with a JWKS endpoint. The daemon uses the key
values to determine if new keys have been introduced or retired. The URI
of the JWKS endpoint is recorded in the HealthAuthority in the database.
The periodicity is controlled with a new environment variable;
JWKS_REFRESH_RATE, which specifies (in seconds) the period the keys are
refreshed.

Fixes #911

## Proposed Changes

* Introduces a new mechanism to periodically refresh a HealthAuthority's keys from a JWKS endpoint.
* HealthAuthority now includes an optional JWKS URI that will be queried to obtain new keys.
* The key refresh rate is controlled by a new configuration parameter: JWKS_REFRESH_RATE, which specifies (in seconds) the refresh rate for all HealthAuthorities with specified JWKS endpoints.

**Release Note**

```release-note
* Introduces a new mechanism to periodically refresh a HealthAuthority's keys from a JWKS endpoint.
* HealthAuthority now includes an optional JWKS URI that will be queried to obtain new keys.
* The key refresh rate is controlled by a new configuration parameter: JWKS_REFRESH_RATE, which specifies (in seconds) the refresh rate for all HealthAuthorities with specified JWKS endpoints.
```


internal/admin/healthauthority/form.go
```diff
@@ -27,12 +27,14 @@ type formData struct {
 	Issuer   string `form:"Issuer"`
 	Audience string `form:"Audience"`
 	Name     string `form:"Name"`
+	JwksURI  string `form:"JwksUri"`
```
sethvargo, line 30: Nit, I think stylistically this would be `JWKsURI`?

internal/setup/setup.go
```diff
@@ -236,6 +243,20 @@ func SetupWith(ctx context.Context, config interface{}, l envconfig.Lookuper) (*
...
+			logger.Info("configuring jwks manager")
+
+			jwksConfig := provider.JWKSManagerConfig()
+			jwks, err := jwks.NewFromEnv(ctx, db, jwksConfig)
+			if err != nil {
+				// Ensure the db is closed on an error.
+				defer db.Close(ctx)
+				return nil, fmt.Errorf("unable to start jwks manager: %w", err)
+			}
+			jwks.Start()
```
sethvargo, line 258: Does it make sense for this to return an error? Looks like it just spawns a goroutine.

internal/verification/database/health_authority_db.go
```diff
@@ -54,11 +54,11 @@ func (db *HealthAuthorityDB) AddHealthAuthority(ctx context.Context, ha *model.H
 		row := tx.QueryRow(ctx, `
 			INSERT INTO
 				HealthAuthority
-				(iss, aud, name)
+				(iss, aud, name, jwks_uri)
```
sethvargo, line 57: This is a 1-1 mapping of a JWK URI and a PHA. Do we envision a world where one PHA could have multiple JWK URIs, maybe during a migraiton? @mikehelmick

internal/verification/model/health_authority.go
```diff
@@ -79,6 +80,15 @@ func (k *HealthAuthorityKey) IsValidAt(t time.Time) bool {
 	return t.After(k.From) && (k.Thru.IsZero() || k.Thru.After(t))
 }
 
+// Revoke revokes a key.
+func (k HealthAuthorityKey) Revoke() {
+	k.Thru = time.Now()
```
sethvargo, line 85: Do we need to UTC this?

migrations/000040_AddJwksUri.up.sql
```diff
@@ -0,0 +1,21 @@
...
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+-- Allows us to periodically update the keys.
+ALTER TABLE HealthAuthority
+	ADD COLUMN jwks_uri VARCHAR(500);
```
sethvargo, line 19: The maximum uniform resource locator (URL) length of 2,083 characters. Should this be `TEXT` instead?

sethvargo (commented):
I think this is good. I'm a little concerned about the constantly-running import of JWKs versus an architecture where we do more of a read-through cache on a per-PHA basis

internal/setup/setup.go
mikehelmick: background refresh may or may not work on serverless infrastructure due to CPU throttling when not serving requests.

Also - does this need to be started in all servers? would it make sense to have this refresh be a periodic job that that syncs keys into the database and then rely on existing public key caching infrastructure?

internal/verification/database/health_authority_db.go
mikehelmick: I'm going to say no to needing multiple.

If someone is migrating from one verification server to another, they will have to use the existing key management.

Seems like the right complexity tradeoff here.

pkg/jwks/jwks.go
```diff
@@ -0,0 +1,261 @@
...
+		return nil, nil
+	}
+
+	req, err := http.NewRequestWithContext(mgr.ctxt, "GET", ha.JwksURI, nil)
+	if err != nil {
+		return nil, fmt.Errorf("creating connection: %w", err)
+	}
+
+	var resp *http.Response
+	resp, err = http.DefaultClient.Do(req)
```
mikehelmick, line 64: set timeout on http client?

pkg/jwks/jwks.go
```diff
@@ -0,0 +1,261 @@
...
+		if keySet[stripKey(key)] == -1 {
+			newKeys = append(newKeys, key)
+		}
+	}
+	sort.Ints(deadKeys)
+	return
+}
+
+// updateHA updates HealthAuthority's keys.
+func (mgr *JWKSManager) updateHA(haDB *hadb.HealthAuthorityDB, ha *model.HealthAuthority, idx int) error {
```
mikehelmick, line 150: suggest using the lock infrastructure on this sever to ensure that this is only updated by one instance at a time and possibly to ensure they are spaced out appropriately

jeremyfaller:
/hold

Seems like polling's a mistake. I'll move this to a cached approach. As I read Mike's Issue, with my limited understanding of this server's architecture, I went with a simple periodic polling. My apologies. Most of this code stays the same -- just what drives the update changes.

mikehelmick:
> Seems like polling's a mistake. I'll move this to a cached approach. As I read Mike's Issue, with my limited understanding of this server's architecture, I went with a simple periodic polling. My apologies. Most of this code stays the same -- just what drives the update changes.

See the pattern we use for things like the cleanup and export jobs.

polling is fine, but controlled rather than distributed

jeremyfaller:
Cool, I'll dig around, find it, and move to that approach. Please hold reviews for a little bit.

internal/verification/model/health_authority.go
jeremyfaller: This was just cargo-culted from the old code. Do we need to do the same for all other times? Eg, the `From` field (which is _not_ currently like that)...

internal/setup/setup.go
jeremyfaller: obsolete.

pkg/jwks/jwks.go
jeremyfaller: obsolete.

jeremyfaller:
/unhold

I think this is ready, I still need to test all the terraform and build changes, so there might be some churn there, but otherwise, I believe this can re-enter review.

terraform/service_jwks.tf
```diff
@@ -0,0 +1,169 @@
...
+resource "google_secret_manager_secret_iam_member" "jwks-db" {
+  for_each = toset([
+    "sslcert",
+    "sslkey",
+    "sslrootcert",
+    "password",
+  ])
+
+  secret_id = google_secret_manager_secret.db-secret[each.key].id
+  role      = "roles/secretmanager.secretAccessor"
```
jeremyfaller, line 44: I'm not sure I need this. Can anyone comment?

terraform/service_jwks.tf
```diff
@@ -0,0 +1,169 @@
...
+  member    = "serviceAccount:${google_service_account.jwks.email}"
+}
+
+resource "google_storage_bucket_iam_member" "jwks-objectadmin" {
```
jeremyfaller, line 48: similarly, this one, I bet I don't, but I'm a little unclear.

builders/build.yaml
```diff
@@ -333,3 +359,4 @@ images:
 - 'gcr.io/${PROJECT_ID}/${_REPO}/generate:${_TAG}'
 - 'gcr.io/${PROJECT_ID}/${_REPO}/key-rotation:${_TAG}'
 - 'gcr.io/${PROJECT_ID}/${_REPO}/migrate:${_TAG}'
+- 'gcr.io/${PROJECT_ID}/${_REPO}/jwks:${_TAG}'
```
sethvargo, line 362: Nit: sort alphabetically

pkg/jwks/jwks.go
```diff
@@ -0,0 +1,249 @@
...
+	req, err := http.NewRequestWithContext(reqCtxt, "GET", ha.JwksURI, nil)
+	if err != nil {
+		return nil, fmt.Errorf("creating connection: %w", err)
+	}
+
+	var resp *http.Response
+	resp, err = http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("reading connection: %w", err)
+	}
```
sethvargo, line 75: Need to close the response body `defer resp.Body.Close()` or else we could leak

pkg/jwks/jwks.go
```diff
@@ -0,0 +1,249 @@
...
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("resp (%v) != %v", resp.StatusCode, http.StatusOK)
+	}
+
+	var bytes []byte
+	bytes, err = ioutil.ReadAll(resp.Body)
```
sethvargo, line 82: I think it'd be more efficient to stream unmarshall directly into the []jwt.JWT instead of having the []byte as an interim.

pkg/jwks/jwks.go
```diff
@@ -0,0 +1,249 @@
...
+
+func (c *Config) JWKSManagerConfig() *Config {
+	return c
+}
+
+// Manager handles updating all HealthAuthorities if they've specified a JWKS
+// URI.
+type Manager struct {
+	ctxt context.Context
+	db   *database.DB
```
sethvargo, line 55: I'd recommend storing the logger on the Manager instead of trying to get it from the context.

terraform/service_jwks.tf
sethvargo: You need this if you want your service to talk to the database (which I think you do 😄).

terraform/service_jwks.tf
sethvargo: This gives you permission to download objects from GCS. I don't think you need this.

tools/admin-console/templates/healthauthority.html
```diff
@@ -49,6 +49,15 @@ <h2>Edit Health Authority '{{.ha.Issuer}}'</h2>
     </div>
   </div>
 
+  <div class="form-group row">
+    <label class="control-label col-sm-3" for="JWKS URI">JWKS URI:</label>
+    <div class="col-sm-6">
+      <input type="text" id="JwksURI" name="JwksURI" size="50" value="{{.ha.JwksURI}}">
+      <small id="NameHelpBlock" class="form-text text-muted">The JWKS URI for the heath authority.
+			  If specified, the keys will be fetched periodically.
```
sethvargo, line 57: Specify how frequently?

terraform/service_jwks.tf
jeremyfaller: thanks!

pkg/jwks/jwks.go
jeremyfaller: I did it this way because it's far easier to test. As written the `getKeys` and `parseKeys` are separable. (Additionally, it's _likely_ [not guaranteed] there isn't much inefficiency...) If you feel passionately enough, I'll end up combining these two functions.

tools/admin-console/templates/healthauthority.html
jeremyfaller: I didn't because it's specified in the run scripts, and therefore is overridable when deployed. I'll update the note to reflect this reality.

pkg/jwks/jwks.go
jeremyfaller: good catch, ty.

pkg/jwks/jwks.go
jeremyfaller: done.

pkg/jwks/jwks.go
sethvargo: N is probably sufficiently small that it's not worth it. Let's do this for now.

jeremyfaller:
I don't have OWNERS here, so I need a second reviewer. @mikehelmick

sethvargo:
@jeremyfaller it actually looks like merging is failing due to some tests

internal/verification/model/health_authority.go
mikehelmick: so, technically, no. The Go comparisons will take time zone into account. But we attempt to make all of our times UTC for storage.

jeremyfaller:
Because of a spelling mistake in the html, I need a new LGTM, please.

PR #1000, opened 2020-09-16
set text/plain content type when uploading index.txt files

mikehelmick:
## Proposed Changes

* Allow for setting of content type on storage.CreateObject
* Set `text/plain` as the content type for index.txt files

**Release Note**

```release-note
Set content type for index files on blob storage.
```

### DIFF
--- a/internal/storage/google_cloud_storage.go
+++ b/internal/storage/google_cloud_storage.go
@@ -44,7 +44,7 @@ func NewGoogleCloudStorage(ctx context.Context) (Blobstore, error) {
 }
 
 // CreateObject creates a new cloud storage object or overwrites an existing one.
-func (s *GoogleCloudStorage) CreateObject(ctx context.Context, bucket, objectName string, contents []byte, cacheable bool) error {
+func (s *GoogleCloudStorage) CreateObject(ctx context.Context, bucket, objectName string, contents []byte, cacheable bool, contentType string) error {
 	cacheControl := "public, max-age=86400"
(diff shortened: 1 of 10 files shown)


gurayAlsac (approved):
/lgtm
/approve

internal/storage/google_cloud_storage.go
```diff
@@ -44,14 +44,17 @@ func NewGoogleCloudStorage(ctx context.Context) (Blobstore, error) {
...
 // CreateObject creates a new cloud storage object or overwrites an existing one.
-func (s *GoogleCloudStorage) CreateObject(ctx context.Context, bucket, objectName string, contents []byte, cacheable bool) error {
+func (s *GoogleCloudStorage) CreateObject(ctx context.Context, bucket, objectName string, contents []byte, cacheable bool, contentType string) error {
 	cacheControl := "public, max-age=86400"
 	if !cacheable {
 		cacheControl = "no-cache, max-age=0"
 	}
 
 	wc := s.client.Bucket(bucket).Object(objectName).NewWriter(ctx)
 	wc.CacheControl = cacheControl
```
icco, line 54: Should we be doing something with https://golang.org/pkg/mime/#TypeByExtension ?

sethvargo: Meh - there's only two types, so I think this is fine for now

sethvargo (approved):
/lgtm

PR #1003, opened 2020-09-16
Fix TestIterateExposuresCursor flakes

chaodaiG:
This test is still flaky https://oss-prow.knative.dev/view/gs/oss-prow/pr-logs/pull/google_exposure-notifications-server/995/pull-en-server-release-unit/1306267928399712256 after https://github.com/google/exposure-notifications-server/pull/983

The root cause of this test being flaky is that testPublishDB.IterateExposures returns non-deterministic orders of output, so it's not possible to predict which 2 exposures out of 3 are iterated first


sethvargo (approved):
/lgtm

PR #1005, opened 2020-09-17
restore previous handling of too large symptom onset

mikehelmick:
## Proposed Changes

* restore previous server behavior of zeroing out days since symptom onset when the value is outside the acceptable range
* continue to return these inside a partial success error message so that developers can catch this

**Release Note**

```release-note
If a calculated days since symptom onset is too large in magnitude (config based, default of 14 days), then the days since symptom onset value will be set to null in exports rather than having that key dropped entirely. This restores the pre v0.6.0 functionality.
```

### DIFF
--- a/internal/publish/model/exposure_model.go
+++ b/internal/publish/model/exposure_model.go
@@ -480,10 +480,15 @@ func (t *Transformer) TransformPublish(ctx context.Context, inData *verifyapi.Pu
 		// Set days since onset, either from the API or from the verified claims (see above).
 		if onsetInterval > 0 {
 			daysSince := DaysFromSymptomOnset(onsetInterval, exposure.IntervalNumber)
+			// Check if the magnitude of this value is too large. If it is too large, we won't want to set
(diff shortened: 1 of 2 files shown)


internal/publish/model/exposure_model.go
```diff
@@ -481,11 +481,11 @@ func (t *Transformer) TransformPublish(ctx context.Context, inData *verifyapi.Pu
 		if onsetInterval > 0 {
 			daysSince := DaysFromSymptomOnset(onsetInterval, exposure.IntervalNumber)
 			if abs := math.Abs(float64(daysSince)); abs > t.maxSymptomOnsetDays {
-				logger.Debugw("dropping key due to symptom onset magnitude too high", "daysSince", daysSince)
```
sethvargo, line 484: Can you comment why (specifically that some states had hard-coded this to 1)? I want to avoid accidentially removing it.

mikehelmick: added

sethvargo (approved):
/lgtm

PR #1008, opened 2020-09-17
Fix exit status for export-analyzer tool

### LINKED ISSUE #1007: Export Analyzer tool always reports error
### TL;DR
When running the export analyzer on an export file with zero errors, an error is reported and the exit status of the program is 1

```bash
$ go run ./tools/export-analyzer -sig -file 1599181440-1599181500-00001.zip -tek-age "2160h" 2>&1 | grep errors
2020/09/17 16:57:50 export file contains errors: 0 errors occurred:
```

I believe this is because a `multierror.Error` is always returned by the function.

**Expected behavior**
No errors should be reported and the exit status should be `0`

**Observed behavior**
Exit status was `1`

### Reproduction

*See above*

**Screenshots**

**Environment**

- OS:
- Browser:
- Version:

**Additional information**

nmiodice:
Fixes #1007 

## Proposed Changes

* Fix the exit status for the export-analyzer tool in the case that there are no errors found in the specified file.

**Release Note**

```release-note
NONE
```


mikehelmick (approved):
/lgtm
/ok-to-test

PR #1009, opened 2020-09-17
Fix e2e test failure when there is no authorized app

chaodaiG:
When there is no authorized app, GetAuthorizedApp returns nil, nil instead of nil, error


sethvargo (approved):
/lgtm

PR #1013, opened 2020-09-20
Return warnings instead of error when symptom onset is too long ago

sethvargo:
Follow-up from GH-1005

This fixes a bug in ENX on iOS which processes the 200 partial response success as an error.

**Release Note**

```release-note
Return warnings instead of error when symptom onset is too long ago
```

### DIFF
--- a/internal/publish/model/exposure_model.go
+++ b/internal/publish/model/exposure_model.go
@@ -412,6 +412,9 @@ func ReportTypeTransmissionRisk(reportType string, providedTR int) int {
 // * 0 exposure Keys in the requests
 // * > Transformer.maxExposureKeys in the request
 //
+// The return params are the list of exposures, a list of warnings, and any
+// errors that occur.
+//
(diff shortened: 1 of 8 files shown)


internal/publish/model/exposure_model.go
```diff
@@ -412,7 +412,7 @@ func ReportTypeTransmissionRisk(reportType string, providedTR int) int {
 // * 0 exposure Keys in the requests
 // * > Transformer.maxExposureKeys in the request
 //
```
mikehelmick, line 414: let's document return params now that there are 3

internal/publish/publish.go
```diff
@@ -332,6 +332,7 @@ func (h *PublishHandler) process(ctx context.Context, data *verifyapi.Publish, b
 			pubResponse: &verifyapi.PublishResponse{
 				ErrorMessage: message,
 				Code:         verifyapi.ErrorBadRequest,
+				Warnings:     transformWarnings,
```
mikehelmick, line 335: this only sets transformWarnings if there is also an error

mikehelmick: line 419 needs to also set warnings if it isn't empty, right?

sethvargo: Ah and further down too. Updated.

mikehelmick (approved):

PR #1019, opened 2020-09-21
update instructions for export signing and key ordering

### LINKED ISSUE #1010: signature info ordering
* update .proto spec, signature block on export files, indicating "newest" signature should appear first in the block (both)
    * add limit to number of signatures that is allowed (10)
* update server to actually implement this
* add instructions to admin console. As soon as you make this key active, notify Google/Apple immediately

/kind enhancement

mikehelmick:
Fixes #1010

## Proposed Changes

* Document ordering of signatures on export files when there is more than one
* Document limit (soft) on export signatures
* Update server for this ordering

**Release Note**

```release-note
Export file signatures will be ordered with the most recently created one first. Documented updated expectations around files with multiple signers.
```

### DIFF
--- a/internal/export/database/export.go
+++ b/internal/export/database/export.go
@@ -317,6 +317,8 @@ func (db *ExportDB) LookupSignatureInfos(ctx context.Context, ids []int64, valid
 				SignatureInfo
 			WHERE
 				id = any($1) AND (thru_timestamp is NULL OR thru_timestamp >= $2)
+			ORDER BY
+				id DESC
 		`, ids, validUntil)
 		if err != nil {
 			return fmt.Errorf("failed to list: %w", err)
--- a/internal/export/database/export_test.go
+++ b/internal/export/database/export_test.go
@@ -83,7 +83,7 @@ func TestLookupSignatureInfos(t *testing.T) {
 	ctx := context.Background()
 
 	testTime := time.Now().UTC()
-	want := []*model.SignatureInfo{
+	create := []*model.SignatureInfo{
 		{
 			SigningKey:        "/kms/project/key/version/1",
 			SigningKeyVersion: "1",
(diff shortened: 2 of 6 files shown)


internal/export/database/export.go
```diff
@@ -317,6 +317,8 @@ func (db *ExportDB) LookupSignatureInfos(ctx context.Context, ids []int64, valid
 				SignatureInfo
 			WHERE
 				id = any($1) AND (thru_timestamp is NULL OR thru_timestamp >= $2)
+			ORDER BY
+				id DESC
```
sethvargo, line 321: The code comment says time order, but this is id. They are _probably_ the same, but not guaranteed to be.

internal/pb/export/export.proto
```diff
@@ -103,6 +107,11 @@ message TemporaryExposureKey {
 }
 
 message TEKSignatureList {
+  // When their are multiple signatures, they must be sorted in time order
```
tshortli, line 110:
```suggestion
  // When there are multiple signatures, they must be sorted in time order
```

internal/export/database/export_test.go
```diff
@@ -102,20 +102,25 @@ func TestLookupSignatureInfos(t *testing.T) {
 			SigningKeyID:      "310",
 		},
 	}
-	for _, si := range want {
+	for _, si := range create {
 		if err := New(testDB).AddSignatureInfo(ctx, si); err != nil {
 			t.Fatalf("failed to add signature info %v: %v", si, err)
 		}
 	}
+	// create a 'reverse' order list.
```
sethvargo, line 110: You can do this without an extra alloc...

```go
for i, j := 0, len(want)-1; i < j; i, j = i+1, j-1 {
  want[i], want[j] = want[j], want[i]
}
```

internal/export/database/export_test.go
```diff
@@ -102,20 +102,25 @@ func TestLookupSignatureInfos(t *testing.T) {
...
 	ids := []int64{want[0].ID, want[1].ID, want[2].ID}
 	got, err := New(testDB).LookupSignatureInfos(ctx, ids, testTime)
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	// The first entry (want[0]) is expired and won't be returned.
-	want = want[1:]
+	// The last entry (want[len-1]) is expired and won't be returned.
+	want = want[:2]
```
sethvargo, line 123: Do we guarantee this won't panic? What if there is only 1 or no keys?

internal/export/database/export.go
mikehelmick: This is the best proxy we have as we don't capture create/effective times.

So relying on the serial column

internal/export/database/export_test.go
mikehelmick: changed around a bit, still creates an extra alloc... but we're talking about 2 pointers in a unit test

PR #1020, opened 2020-09-22
handle nil jwks URI in database

mikehelmick:
## Proposed Changes

* Handle nil jwks_uri columns in the healthauthority table - migrated rows have a null column value.

**Release Note**

```release-note
bugfix: handle nil JWKS URI values.
```


sethvargo (approved):
/lgtm

PR #1021, opened 2020-09-22
Add stack driver debug logging

yegle:
To better understand https://github.com/google/exposure-notifications-verification-server/issues/474

There should be no impact to services if this is not enabled.


sethvargo (approved):
/lgtm

PR #1025, opened 2020-09-23
E2E test: hook up with verification

chaodaiG:
Also refactor integration test to be able to share more code with e2e test:
- Separate server setup from data seeding
- Authorized app provider backed by db instead of memory

### DIFF
--- a/internal/integration/helpers.go
+++ b/internal/integration/helpers.go
@@ -43,6 +43,7 @@ import (
 	"github.com/sethvargo/go-envconfig"
 	"github.com/sethvargo/go-retry"
 
+	authorizedappdb "github.com/google/exposure-notifications-server/internal/authorizedapp/database"
 	authorizedappmodel "github.com/google/exposure-notifications-server/internal/authorizedapp/model"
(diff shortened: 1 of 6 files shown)


chaodaiG:
@sethvargo , PTAL

internal/integration/helpers.go
```diff
@@ -259,21 +234,57 @@ func NewTestServer(tb testing.TB, exportPeriod time.Duration) (*serverenv.Server
...
+		if err := authorizedappdb.New(db).InsertAuthorizedApp(context.Background(), &authorizedappmodel.AuthorizedApp{
+			AppPackageName: "com.example.app",
+			AllowedRegions: map[string]struct{}{
+				"TEST": {},
+			},
+			AllowedHealthAuthorityIDs: map[int64]struct{}{
+				1: {},
+			},
+
+			// TODO: hook up verification and revision
```
sethvargo, line 282: False is the default value, which means this _is_ hooked up.

sethvargo (approved):
/lgtm

One comment seems outdated.

PR #1026, opened 2020-09-23
Log HealthAuthority when publishing.

### LINKED ISSUE #1011: log "healthAuthorityId" in publish logs
this will help correlate failures back to the entity that is attempting to upload TEKs

jeremyfaller:
Increase the logging (by including the HealthAuthority's id) in the publish API.

Fixes #1011

### DIFF
--- a/internal/publish/publish.go
+++ b/internal/publish/publish.go
@@ -179,9 +179,11 @@ func (h *PublishHandler) process(ctx context.Context, data *verifyapi.Publish, b
 	ctx, span := trace.StartSpan(ctx, "(*publish.PublishHandler).process")
 	defer span.End()
 
-	logger := logging.FromContext(ctx)
+	logger := logging.FromContext(ctx).With("health_authority_id", data.HealthAuthorityID)
 	metrics := h.serverenv.MetricsExporter(ctx)
 	metricsMiddleWare := metricsware.NewMiddleWare(&metrics)
+
+	logger.Info("publish API request")
 
 	appConfig, err := h.authorizedAppProvider.AppConfig(ctx, data.HealthAuthorityID)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/publish/publish.go
```diff
@@ -183,6 +184,8 @@ func (h *PublishHandler) process(ctx context.Context, data *verifyapi.Publish, b
 	metrics := h.serverenv.MetricsExporter(ctx)
 	metricsMiddleWare := metricsware.NewMiddleWare(&metrics)
 
+	logger.Info("publish API request", zap.String("HealthAuthorityID", data.HealthAuthorityID))
```
sethvargo, line 187:
```suggestion
	logger.Infow("publish API request", "health_authority_id", data.HealthAuthorityID)
```

Note the `w` and snake case (works better for our logger)

internal/publish/publish.go
```diff
@@ -259,7 +262,7 @@ func (h *PublishHandler) process(ctx context.Context, data *verifyapi.Publish, b
 	// generous defaults. If there isn't a region set, then the TEKs
 	// won't be retrievable, so we ensure there is something set.
 	if len(regions) == 0 {
-		logger.Errorf("No regions present in request or configured for healthAuthorityID: %v", data.HealthAuthorityID)
+		logger.Errorf("No regions present in request or configured for HealthAuthorityID: %v", data.HealthAuthorityID)
```
sethvargo, line 265:
```suggestion
		logger.Errorw("no regions present in request or configured for health authority", "health_authority_id", data.HealthAuthorityID)
```

internal/publish/publish.go
sethvargo: Ah actually, after seeing the rest of the code, I think we should jump up to L183 and change it to:

```go
logger := logging.FromContext(ctx).With("health_authority_id", data.HealthAuthorityID)
```

and then leave the rest of this code as-was.

sethvargo (changes requested):

internal/publish/publish.go
jeremyfaller: Whoa, I had no idea our logger was so awesome. You got it!

sethvargo: Yea, `.Named` and `.With` create new loggers and propagate the context through. So it'll add that field to all logger.Something calls

sethvargo (approved):
/lgtm

PR #1027, opened 2020-09-23
Move key server specific view registration to a separate function.

yegle:
The exporter is shared between key server and verification server, we
should not have key server specific code in the exporter.


yegle:
Ping?

sethvargo (approved):
/lgtm

MagicTmn (commented):
![Uploading 95170701...]

PR #1037, opened 2020-09-28
Fix incremental e2e test by use new health authority

chaodaiG:
Health authority keys generated from each e2e test are different, which fails incremental e2e tests if using existing health authority. This PR updates e2e to create new health authority for every test to make sure incremental test work


chaodaiG:
/uncc icco sethvargo
Fixing the failing test

mikehelmick (approved):

PR #1039, opened 2020-09-29
Fix the metric missing issue caused by repeated metric point upload.

yegle:
See
https://github.com/google/exposure-notifications-verification-server/issues/474#issuecomment-700351001
for an explanation.


icco (approved):
/lgtm

omg. If this works, you're the truest of all heros.

sethvargo:
For the record, even if it doesn't work, I still think you're a hero 🦸‍♂️

PR #1042, opened 2020-09-30
Also migrate db in incremental e2e test

chaodaiG:
Incremental e2e tests fails when there is a DB schema change, updating the test flow to also migrate DB. This was tested locally and worked.

/cc @mikehelmick @sethvargo


sethvargo (approved):
/lgtm

PR #1049, opened 2020-10-02
E2e test: export test

chaodaiG:
As export endpoints are not publicly available, testing exporting by relying on cloud scheduler:

- Make cloud scheduler cron string configurable, and each runs every minute in e2e test
- Worst case scenario is 60 seconds for export batch, and another 60 seconds for export file, let e2e test wait 150 seconds to allow this happen


chaodaiG:
@sethvargo PTAL

sethvargo (approved):
/lgtm

PR #1051, opened 2020-10-05
Refactoring how OpenCensus views are registered.

yegle:
The commit
https://github.com/google/exposure-notifications-server/commit/fc694833c41f00f7156fdff46ff46d041867b60c,
was a good intention to not register key server views on verification
server. However, it only works on key server as there's a central place
where we have all the views needed to register. The verification server
has views scattered across different files, and cannot easily adopt the
pattern in the commit.

This PR works by 1) collect views in func init(), 2) register them in
StartExporter. Doing this will achieve the same goal of
https://github.com/google/exposure-notifications-server/commit/fc694833c41f00f7156fdff46ff46d041867b60c,
and makes it possible for verification to adopt.


internal/metrics/federationin/views.go
```diff
@@ -2,11 +2,12 @@ package federationin
 
 import (
```
sethvargo, line 3: Are we sure this file is imported to run `init`?

yegle: `./cmd/federationin/main.go` imports `github.com/google/exposure-notifications-server/internal/federationin`.

Anyway, if the code uses uses `stat.Record` on the measures (e.g. `PullInvalidRequest`), it must have imported this module.

sethvargo (approved):
/lgtm

MagicTmn (commented):
I need to listen to this device imei 86413105713044 subscriber number +77476232025

PR #1056, opened 2020-10-06
Unify custom domain logic

sethvargo:
Also adds the ability to set groups/users that can invoke the debugger and fronts the Cloud Storage bucket with the load balancer too (so users can enable the CDN and a custom domain if they want)

**Release Note**

```release-note
**Warning** - changes in the Terraform configuration will cause SSL certificates to re-provision which could cause minor downtime.
```


icco (approved):
/lgtm

PR #1061, opened 2020-10-06
Fix e2e terraform deployment by supplying hosts

chaodaiG:
Terraform smoke test failed due to index out of bound error at https://github.com/google/exposure-notifications-server/blob/5e41925d7b2c441185979827f7ae58dc29991429/terraform/lb.tf#L203

### DIFF
--- a/terraform-e2e/main.tf
+++ b/terraform-e2e/main.tf
@@ -39,6 +39,11 @@ module "en" {
 
   create_env_file = true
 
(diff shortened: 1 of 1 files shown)


chaodaiG:
@sethvargo , mind add a /lgtm label?

sethvargo:
It has it - looks like I need to 

/approve

PR #1062, opened 2020-10-06
Don't fail terraform if there is no custom domain supplied

chaodaiG:
By the way remove custom domain from e2e test, as google api requires valid domain name for creating ssl cert. 

See terraform smoke test failure: https://prow.k8s.io/view/gcs/oss-prow/logs/ci-en-server-terraform-smoke/1313591663427325952#1:build-log.txt%3A13898

### DIFF
--- a/terraform/lb.tf
+++ b/terraform/lb.tf
@@ -200,5 +200,5 @@ resource "google_compute_managed_ssl_certificate" "default" {
 }
 
 output "lb_ip" {
-  value = local.enable_lb ? google_compute_global_address.key-server[0].address : ""
+  value = google_compute_global_address.key-server.*.address
 }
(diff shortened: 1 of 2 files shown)


terraform/lb.tf
```diff
@@ -200,5 +200,5 @@ resource "google_compute_managed_ssl_certificate" "default" {
 }
 
 output "lb_ip" {
-  value = google_compute_global_address.key-server[0].address
+  value = local.enable_lb ? google_compute_global_address.key-server[0].address : ""
```
yegle, line 203: Use null instead of "".

yegle (approved):

sethvargo (approved):
/lgtm

PR #1066, opened 2020-10-08
E2E test: deflake by waiting longer for export

chaodaiG:
E2E test failed ~50% of the times, due to export not created yet. Inspected Cloud Run logs and found out that it could take > 3 minutes to create an export file in GCS, this is not avoidable as export endpoint was triggered by cloud scheduler, which could only be schduled to be on minute level.

Increase the polling time to 300 seconds should make it stable


icco (approved):
/lgtm

PR #1067, opened 2020-10-09
Exclude regions in export.

### LINKED ISSUE #1054: Export config
Add more options

* include all regions, but exclude a specific list
* only include non-travellers.

The existing setup works well for

* home country + travellers export

We need the ability to create a second export that represents

* foreign country non travelers

/kind enhancement

jeremyfaller:
Recreate the ability to exclude regions from the exports.

Updates #1054

## Proposed Changes

* Recreate ExcludeRegions in exports. This will allow situations where traveling users can be excluded in an export.

**Release Note**

```release-note
Recreate ExcludeRegions in exports. This will allow situations where traveling users can be excluded in an export.

```

### DIFF
--- a/internal/admin/exports/form.go
+++ b/internal/admin/exports/form.go
@@ -17,6 +17,7 @@ package exports
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 
@@ -28,6 +29,7 @@ type formData struct {
 	OutputRegion     string        `form:"OutputRegion"`
 	InputRegions     string        `form:"InputRegions"`
 	IncludeTravelers bool          `form:"IncludeTravelers"`
+	ExcludeRegions   string        `form:"ExcludeRegions"`
 	BucketName       string        `form:"BucketName"`
 	FilenameRoot     string        `form:"FilenameRoot"`
 	Period           time.Duration `form:"Period"`
@@ -36,6 +38,20 @@ type formData struct {
 	ThruDate         string        `form:"thrudate"`
 	ThruTime         string        `form:"thrutime"`
 	SigInfoIDs       []int64       `form:"siginfo"`
+}
+
+// splitRegions turns a string of regions (generally separated by newlines), and
+// breaks them up into an alphabetically sorted slice of strings.
+func splitRegions(regions string) []string {
+	ret := make([]string, 0, 20)
+	for _, s := range strings.Split(regions, "\n") {
+		s := strings.TrimSpace(s)
+		if s != "" {
+			ret = append(ret, s)
+		}
+	}
+	sort.Strings(ret)
+	return ret
 }
 
 func (f *formData) PopulateExportConfig(ec *model.ExportConfig) error {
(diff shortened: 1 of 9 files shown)


jeremyfaller:
I think this now actually ready for review.

/unhold

internal/admin/exports/form.go
```diff
@@ -60,6 +61,13 @@ func (f *formData) PopulateExportConfig(ec *model.ExportConfig) error {
 		}
 	}
 	ec.IncludeTravelers = f.IncludeTravelers
+	ec.ExcludeRegions = make([]string, 0)
+	for _, s := range strings.Split(f.ExcludeRegions, "\n") {
```
mikehelmick, line 65: can we extract this loop out to a function? duplicate code.

internal/export/database/export_test.go
```diff
@@ -541,6 +542,109 @@ func TestTravelerKeys(t *testing.T) {
...
+		OutputRegion:   ec.OutputRegion,
+		Status:         model.ExportBatchOpen,
+		ExcludeRegions: ec.ExcludeRegions,
+	}
+	if err := New(testDB).AddExportBatches(ctx, []*model.ExportBatch{eb}); err != nil {
+		t.Fatal(err)
+	}
+
+	// Create person in main region.
+	trav := &publishmodel.Exposure{
```
mikehelmick, line 585: this says `trav` but the exposure isn't marked as a traveler

mikehelmick (commented):
This is only part of what #1054 needs - fine for the rest to be a followup.

The missing part is the "only include non-travellers" option

There is the "include travelers" option today, which lets one set up
 -home region + travelers
but we can't set up yet
- foreign regions, non-travelers

internal/admin/exports/form.go
jeremyfaller: done, and added test.

internal/export/database/export_test.go
jeremyfaller: my mistake

internal/admin/exports/form.go
```diff
@@ -38,6 +39,19 @@ type formData struct {
 	SigInfoIDs       []int64       `form:"siginfo"`
 }
 
+// splitRegions turns a string of regions (generally separated by newlines), and
+// breaks them up into a slice of strings.
+func splitRegions(regions string) []string {
+	ret := make([]string, 0)
```
sethvargo, line 45: This should probably have some sane starting capacity.

sethvargo: And should we do any kind of sorting for consistency?

internal/admin/exports/form_test.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func TestSplitRegions(t *testing.T) {
+	tests := []struct {
+		r string
+		e []string
+	}{
+		{"", []string{}},
+		{"  test  ", []string{"test"}},
+		{"test\n\rfoo", []string{"test", "foo"}},
```
sethvargo, line 15: Can we add a test for _just_ `\n` separated?

sethvargo (commented):
Two Qs

internal/admin/exports/form_test.go
jeremyfaller: Done.

internal/admin/exports/form.go
jeremyfaller: Sure, NBD.

mikehelmick:
/lgtm

Followup still needed: configuration for foreign regions, non-travelers

Issue #928, opened 2020-08-28
Proposal: Key Sever Migration Support


mikehelmick:
### TL;DR 

For various reasons, some jurisdictions that have launched exposure notifications systems would like to change servers (usually to merge with neighboring jurisdictions).

We make a couple of assumptions / state some facts

1. The severs are not able to share export signing keys

1. It is not possible for there to be 2 active export signing keys for an application and that active signing key version only rolls forward from a -> b

1. There is some transition period while the end user applications switch over where they read and write from. Because of the way that signing keys are published, an application can only read from one server.

1. Bidirectional federation is not an option either due to size of keys that would have to be replicated to the smaller jurisdiction or legal agreements that might need to be put into place to allow for this.

In order to move a jurisdiction from one key sever to another we need to transition where their client applications read exports from and publish TEKs to and we need to ensure that clients that haven't yet migrated are able to get the unified key set.

#### Quick path ⏱️ 

Ideally, applications will have a built in remote configuration capability that allows the jurisdiction to dynamically switch from one server to another. This allows for a rapid cutover and a one time ETL to move recently upload keys from server A to server B.

This is dependent on public key rotation speed as well. Even in the quick path, we may have to execute the not so quick path for a period.

#### Not so quick path ⌛ 

If, the only way to make this switch is through pushing an application update, we have to plan for a prolonged period where older versions only know how to talk to the old server.

We use this high level process, assuming a move from server `A` to server `B`

1. Set up outbound federation from server `A` to server `B`

1. Start the proxy / resigner server at server `A` (see design below). At this point in time, operators of server `A` stop generating their own export files and all of their keys transit through server `B` and the proxied exports.

1. Change the public key for the application to be the public key for server `B`

1. Push the application update. New versions of the app will only communicate with server `B`. Clients that haven't updated will only communicate with server `A`. Once download traffic to server `A` subsides, that sever can be decommissioned.

### Design

**Proposal**

Create an export proxy job that knows how to mirror exports from one server into the CDN storage of the other.

During the proxy, the zip file is unpacked, a second signature is added, and the zip file is repackaged.

Once the key has been rotated for the app to just the sever the exports originate from, the resigner functionality can be turned off. At that point, it's just mirroring the export files from one CDN to another.

**Alternatives considered**

Dual read based solutions are not viable because of the key rotation limitations.

/kind enhancement

gurayAlsac:
[...]

PR #1070, opened 2020-10-10
add mirror service

mikehelmick:
Towards #928

## Proposed Changes

* Add mirror service that can mirror an index.txt / export archive structure from a CDN to the configured cloud storage. For use in key server migrations.

**Release Note**

```release-note
New: Mirror service `cmd/mirror` for mirroring an export directory from one key server to another.
```


internal/mirror/database/mirror.go
```diff
@@ -0,0 +1,223 @@
...
+
+func (db *MirrorDB) Mirrors(ctx context.Context) ([]*model.Mirror, error) {
+	var mirrors []*model.Mirror
+
+	if err := db.db.InTx(ctx, pgx.ReadCommitted, func(tx pgx.Tx) error {
+		rows, err := tx.Query(ctx, `
+			SELECT
+				id, index_file, export_root, cloud_storage_bucket, filename_root
+			FROM
+				mirror
```
sethvargo, line 88: `ORDER BY id` or something?

internal/mirror/database/mirror.go
```diff
@@ -0,0 +1,223 @@
...
+					mirror_id = $1 AND filename = $2
+				`); err != nil {
+				return fmt.Errorf("failed to prepare delete DB statement: %w", err)
+			}
+
+			for _, mf := range toDelete {
+				if _, err := tx.Exec(ctx, deleteName, mf.MirrorID, mf.Filename); err != nil {
+					return fmt.Errorf("failed to delete mirrofile: %w", err)
+				}
+			}
```
sethvargo, line 153: Do we need to check if the number of rows modified == len(toDelete)?

internal/mirror/database/mirror.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+
+		// Create files if they still need to be created.
+		if len(wantFiles) > 0 {
+			if _, err := tx.Prepare(ctx, insertName, `
```
sethvargo, line 158: This is the same transaction as the delete, right? If wantFiles and toDelete include the same (mirror_id, filename), what happens?

internal/mirror/mirror.go
```diff
@@ -0,0 +1,276 @@
...
+func (s *Server) downloadIndex(mirror *model.Mirror) ([]string, error) {
+	client := &http.Client{
+		Timeout: s.config.IndexFileDownloadTimeout,
+	}
+	resp, err := client.Get(mirror.IndexFile)
+	if err != nil {
+		return nil, fmt.Errorf("error downloading index file: %w", err)
+	}
+	defer resp.Body.Close()
+	bytes, err := ioutil.ReadAll(resp.Body)
```
sethvargo, line 62: ReadAll has no limit and will consume all available memory given. We should use a limitReader or something with a sane max here.

internal/mirror/mirror.go
```diff
@@ -0,0 +1,276 @@
...
+			if err != nil {
+				if errors.Is(err, database.ErrAlreadyLocked) {
+					logger.Infow("mirror already locked, skipping", "mirrorID", mirror.ID)
+					continue
+				}
+				logger.Warn(err)
+				w.WriteHeader(http.StatusInternalServerError)
+				return
+			}
+			defer func() {
```
sethvargo, line 122: This defer won't run until _all_ things in the loop have finished and the rest of the function finishes executing. I don't think that's what you want?

internal/mirror/mirror.go
```diff
@@ -0,0 +1,276 @@
...
+			if err != nil {
+				logger.Errorw("unable to list mirror files", "error", err)
+				continue
+			}
+
+			// download the index
+			// curFiles contains the fullly qualified download link for the files.
+			curFiles, err := s.downloadIndex(mirror)
+			if err != nil {
+				logger.Errorw("unable to download index.txt", "error", err)
```
sethvargo, line 141: continue or break?

internal/mirror/mirror.go
```diff
@@ -0,0 +1,276 @@
...
+					}
+				}
+			}
+
+			// delete any files we no longer have
+			for filename, status := range actions {
+				if status.needsDelete() {
+					logger.Infow("deleting stale export file", "file", filename)
+
+					bsCtx, cancel := context.WithTimeout(ctx, time.Second*30)
```
sethvargo, line 172: Nit: 30*time.Second

internal/mirror/mirror.go
```diff
@@ -0,0 +1,276 @@
...
+			indexObjects := make([]*FileStatus, 0, len(actions))
+			// quickly bypass anything that doesn't need to be downloaded.
+			// This ensures that if we timeout, files we previously knew about remain saved.
+			for _, status := range actions {
+				if !status.needsDownload() {
+					status.Saved = true
+					indexObjects = append(indexObjects, status)
+				}
+			}
+			logger.Infow("mirror state", "mirrorID", mirror.ID, "total", len(actions), "existing", len(indexObjects), "todo", len(actions)-len(indexObjects))
```
sethvargo, line 194: TODO

internal/mirror/mirror.go
```diff
@@ -0,0 +1,276 @@
...
+					defer resp.Body.Close()
+					bytes, err := ioutil.ReadAll(resp.Body)
+					if err != nil {
+						logger.Errorw("unable to read file", "filename", filename, "error", err)
+						status.Failed = true
+						continue
+					}
+
+					// Write this file to the blob store.
+					bsCtx, cancel := context.WithTimeout(ctx, time.Minute)
```
sethvargo, line 225: Why is the other 30*time.Second and this is one minute?

internal/mirror/mirror.go
```diff
@@ -0,0 +1,276 @@
...
+
+		status := http.StatusOK
+		w.WriteHeader(status)
+		w.Write([]byte(http.StatusText(status)))
+	}
+}
+
+func deadlinePassed(deadline time.Time) bool {
+	return time.Now().After(deadline)
+}
```
sethvargo, line 276: I'd really like to see this broken down into smaller functions that are easily tested. The current HTTP handler is very long and _very_ nested.

sethvargo (changes requested):
I think the mirror _algorithm_ is fine, but the implementation has a few flaws. Since everything is a giant loop, none of the defers are running until the very end. Furthermore, for a very large N, we'll hit unnecessary timeouts due to lack of parallelization.

I'd like to suggest we:

1. Break apart the main HTTP handler into (at least) one testable function.

1. Leverage some parallelization (either primitively with goroutines or using a library like https://github.com/gammazero/workerpool) with a configurable max.

1. Add tests for `internal/mirror/mirror.go`. The blobstore should be stubbable to the in-memory implementation.

internal/mirror/database/mirror.go
mikehelmick: added, but not necessary for proper function.

internal/mirror/database/mirror.go
mikehelmick: added - it's individual row deletes on the prepared statement though

internal/mirror/database/mirror.go
mikehelmick: wantFiles only contains files that weren't read back into knownFiles. 

added comments to try and make this clearer

internal/mirror/mirror.go
mikehelmick: added limits to both downloads

internal/mirror/mirror.go
mikehelmick: extracted processing to a function so all defers will run.

internal/mirror/mirror.go
mikehelmick: continue.

internal/mirror/mirror.go
mikehelmick: toto - as in, what's left todo 🤣

internal/mirror/mirror.go
mikehelmick: one is delete and one is write

internal/mirror/mirror.go
mikehelmick: can do as a follow up if that's OK

mikehelmick:
> I think the mirror _algorithm_ is fine, but the implementation has a few flaws. Since everything is a giant loop, none of the defers are running until the very end. Furthermore, for a very large N, we'll hit unnecessary timeouts due to lack of parallelization.

Parallelization is achieved at the HTTP layer - will have a cron to active this service every 60 seconds. if one mirror is locked and taking a while, the next invocation will move onto the next one.

State is saved on an http request timeout and processing will continue on next invoke.


> 
> I'd like to suggest we:
> 
> 1. Break apart the main HTTP handler into (at least) one testable function.

will do in follow up.

> 2. Leverage some parallelization (either primitively with goroutines or using a library like https://github.com/gammazero/workerpool) with a configurable max.

we can add additional parallelism, but doesn't seem to be necessary in the testing i've done.

> 3. Add tests for `internal/mirror/mirror.go`. The blobstore should be stubbable to the in-memory implementation.

would like to do in follow up

internal/mirror/mirror.go
sethvargo: yup

sethvargo (approved):
/lgtm
/hold

in case someone else wants to give it a once-over. Feel free to unhold.

PR #1072, opened 2020-10-13
Add OnlyNonTravelers to exports.

### LINKED ISSUE #1054: Export config

jeremyfaller:
Fixes #1054

## Proposed Changes

* Add OnlyNonTravelers to Exports -- exports only non-traveling external users.

**Release Note**

```release-note
Adds ability to only export non-travelers.
```

### DIFF
--- a/internal/publish/database/exposure.go
+++ b/internal/publish/database/exposure.go
@@ -73,6 +73,7 @@ func New(db *database.DB) *PublishDB {
 type IterateExposuresCriteria struct {
 	IncludeRegions   []string
 	IncludeTravelers bool // Include records in the IncludeRegions OR travalers
+	OnlyNonTravelers bool
 	OnlyTravelers    bool // Only includes records marked as travelers.
 	ExcludeRegions   []string
 	SinceTimestamp   time.Time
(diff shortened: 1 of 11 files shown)


icco (approved):
/lgtm

internal/publish/database/exposure.go
```diff
@@ -236,6 +237,11 @@ func generateExposureQuery(criteria IterateExposuresCriteria) (string, []interfa
 		q += fmt.Sprintf(" AND traveler = $%d", len(args))
 	}
 
+	if criteria.OnlyNonTravelers {
```
mikehelmick, line 240: we need validation here so as not to have a paradox.

i.e. if you select OnlyTravlers and OnlyNonTravelers

jeremyfaller: Add a check, and added some text to the HTML. PTAL.

mikehelmick (approved):

PR #1074, opened 2020-10-13
add mirror service to terraform

### LINKED ISSUE #928: Proposal: Key Sever Migration Support

mikehelmick:
Fixes #928

## Proposed Changes

* Add terraform to support the mirror job

**Release Note**

```release-note
Mirror service terraform config for key server migration.
```

### DIFF
--- a/terraform/service_mirror.tf
+++ b/terraform/service_mirror.tf
@@ -140,35 +140,8 @@ resource "google_cloud_run_service_iam_member" "mirror-invoker" {
   member   = "serviceAccount:${google_service_account.mirror-invoker.email}"
 }
 
-resource "google_cloud_scheduler_job" "mirror-worker" {
-  name             = "mirror-worker"
-  region           = var.cloudscheduler_location
-  schedule         = "* * * * *"
-  time_zone        = "America/Los_Angeles"
-  attempt_deadline = "600s"
-
-  retry_config {
-    retry_count = 1
-  }
-
-  http_target {
-    http_method = "POST"
-    uri         = "${google_cloud_run_service.mirror.status.0.url}/"
-    oidc_token {
-      audience              = google_cloud_run_service.mirror.status.0.url
-      service_account_email = google_service_account.mirror-invoker.email
-    }
-  }
-
-  depends_on = [
-    google_app_engine_application.app,
-    google_cloud_run_service_iam_member.mirror-invoker,
-    google_project_service.services["cloudscheduler.googleapis.com"],
-  ]
-}
-
-resource "google_cloud_scheduler_job" "mirror-schedule" {
-  name             = "mirror-cschedule"
+resource "google_cloud_scheduler_job" "mirror-invoke" {
+  name             = "mirror-invoke"
   region           = var.cloudscheduler_location
   schedule         = "*/5 * * * *"
   time_zone        = "America/Los_Angeles"
@@ -180,7 +153,7 @@ resource "google_cloud_scheduler_job" "mirror-schedule" {
 
   http_target {
     http_method = "GET"
-    uri         = "${google_cloud_run_service.mirror.status.0.url}/schedule"
+    uri         = "${google_cloud_run_service.mirror.status.0.url}/"
     oidc_token {
       audience              = google_cloud_run_service.mirror.status.0.url
       service_account_email = google_service_account.mirror-invoker.email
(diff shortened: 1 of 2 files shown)


terraform/service_mirror.tf
```diff
@@ -0,0 +1,195 @@
...
+  role     = "roles/run.invoker"
+  member   = "serviceAccount:${google_service_account.mirror-invoker.email}"
+}
+
+resource "google_cloud_scheduler_job" "mirror-worker" {
+  name             = "mirror-worker"
+  region           = var.cloudscheduler_location
+  schedule         = "* * * * *"
+  time_zone        = "America/Los_Angeles"
+  attempt_deadline = "600s"
```
icco, line 148: With a deadline of 600s, and a schedule of every minute, this could have up to ten running at the same time, is that ok?

terraform/service_mirror.tf
```diff
@@ -0,0 +1,195 @@
...
+  http_target {
+    http_method = "POST"
+    uri         = "${google_cloud_run_service.mirror.status.0.url}/"
+    oidc_token {
+      audience              = google_cloud_run_service.mirror.status.0.url
+      service_account_email = google_service_account.mirror-invoker.email
+    }
+  }
+
+  depends_on = [
```
icco, line 163: Should also depend on google_cloud_run_service.mirror

terraform/service_mirror.tf
```diff
@@ -0,0 +1,195 @@
...
+    google_project_service.services["cloudscheduler.googleapis.com"],
+  ]
+}
+
+resource "google_cloud_scheduler_job" "mirror-schedule" {
+  name             = "mirror-cschedule"
+  region           = var.cloudscheduler_location
+  schedule         = "*/5 * * * *"
+  time_zone        = "America/Los_Angeles"
+  attempt_deadline = "600s"
```
icco, line 175: with a deadline of 600s, and a schedule of every five minutes, this could have two running at the same time, is that ok?

terraform/service_mirror.tf
mikehelmick: yes - that's intentional.

Any given mirror instance holds a lock while it's running.

terraform/service_mirror.tf
mikehelmick: yes - intentional.

DB level locks are used to ensure one worker at a time on any given mirror.

icco (approved):
/lgtm

PR #1078, opened 2020-10-14
Make sure map is initialized

sethvargo:
**Release Note**

```release-note
Ensure map is initialized preventing runtime panic
```

### DIFF
--- a/internal/export/database/export.go
+++ b/internal/export/database/export.go
@@ -425,7 +425,7 @@ func (db *ExportDB) LatestExportBatchEnd(ctx context.Context, ec *model.ExportCo
 // ListLatestExportBatchEnds returns a map of export config IDs to their latest
 // batch end times.
 func (db *ExportDB) ListLatestExportBatchEnds(ctx context.Context) (map[int64]*time.Time, error) {
-	var ts map[int64]*time.Time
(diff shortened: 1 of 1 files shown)


jeremyfaller (approved):

mikehelmick (approved):

whaught (approved):
/lgtm

internal/export/database/export.go
```diff
@@ -425,7 +425,7 @@ func (db *ExportDB) LatestExportBatchEnd(ctx context.Context, ec *model.ExportCo
 // ListLatestExportBatchEnds returns a map of export config IDs to their latest
 // batch end times.
 func (db *ExportDB) ListLatestExportBatchEnds(ctx context.Context) (map[int64]*time.Time, error) {
-	var ts map[int64]*time.Time
+	ts := make(map[int64]*time.Time, 8)
```
chaodaiG, line 428: Where is the number 8 from? Should there be a comment explaining it?

sethvargo: It's completely arbitrary. It pre-allocates the map of that size. We have no way of knowing how many records there will be, but 8 is pretty safe"

chaodaiG: ack

PR #1080, opened 2020-10-14
add some export-importer tests

mikehelmick:
## Proposed Changes

* adds tests. Tests are good.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/internal/exportimport/scheduler.go
+++ b/internal/exportimport/scheduler.go
@@ -20,10 +20,14 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
+	"path"
 	"strings"
 	"time"
 
 	"github.com/google/exposure-notifications-server/internal/database"
+	exportimportdb "github.com/google/exposure-notifications-server/internal/exportimport/database"
+	"github.com/google/exposure-notifications-server/internal/exportimport/model"
 	"github.com/google/exposure-notifications-server/pkg/logging"
 	"go.opencensus.io/trace"
 )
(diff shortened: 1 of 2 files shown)


internal/exportimport/scheduler.go
```diff
@@ -116,3 +106,27 @@ func (s *Server) handleSchedule(ctx context.Context) http.HandlerFunc {
...
+
+func syncFilesFromIndex(ctx context.Context, db *exportimportdb.ExportImportDB, config *model.ExportImport, index string) (int, error) {
+	zipNames := strings.Split(index, "\n")
+	currentFiles := make([]string, 0, len(zipNames))
+	for _, zipFile := range zipNames {
+		if len(strings.TrimSpace(zipFile)) == 0 {
+			// drop blank lines.
+			continue
+		}
+		proposedPath := fmt.Sprintf("%s%s", config.ExportRoot, strings.TrimSpace(zipFile))
```
jeremyfaller, line 118: path.Join?

sethvargo: `filepath.Join` in case someone runs this on Windows ever

mikehelmick: these aren't file paths, they're URL paths

path.Join doesn't do well when the protocol/hostname is in play

You end up with `https:///myserver/a.zip` if Join is used.

jeremyfaller: perhaps rename paths to urls?

jeremyfaller: perhaps rename paths to urls?

mikehelmick: done

sethvargo (approved):
/lgtm

PR #1086, opened 2020-10-15
Use different envvars to trigger integration

sethvargo:
I lost a few hours trying to debug why my tests were failing locally. DB_NAME is the envvar automatically set by the dev script, so that triggered the "integration test" path, except my local database already had data, leading to bizarre and very difficult to debug failures. This could have been worse - it's possible someone runs tests when connected to the production database.

This changes the environment variables to be prefixed with INTEGRATION_ to reduce the chances of such collisions.

/cc @chaodaiG 

**Release Note**

```release-note
Use different envvars to trigger integration tests
```


sethvargo:
I think we also need to change something in the integration environment for this to work.

mikehelmick (approved):

mikehelmick:
waiting for word on integration env changes

internal/integration/helpers.go
```diff
@@ -55,23 +55,32 @@ import (
...
+		if err != nil {
+			tb.Fatal(err)
+		}
 	}
 
-	db := database.NewTestDatabase(tb)
-	if v := os.Getenv("DB_NAME"); v != "" && !testing.Short() {
-		dbConfig := &database.Config{}
+	var db *database.DB
+	if v := os.Getenv("INTEGRATION_DB_NAME"); v != "" {
```
chaodaiG, line 81: The logic inside this if clause uses `DB_NAME` env var for establishing database connection from e2e test, I would suggest change this if condition to be
```suggestion
	if v := os.Getenv("RUN_E2E_TEST"); v != "" {
```

internal/integration/helpers.go
```diff
@@ -55,23 +55,32 @@ import (
...
+	if testing.Short() {
+		tb.SkipNow()
 	}
 
-	if v := os.Getenv("GOOGLE_CLOUD_BUCKET"); v != "" && !testing.Short() {
+	ctx := context.Background()
+
+	var bs storage.Blobstore
+	var err error
+	if v := os.Getenv("INTEGRATION_GOOGLE_CLOUD_BUCKET"); v != "" {
```
chaodaiG, line 67: Please also update this line: https://github.com/google/exposure-notifications-server/blob/5ddd4aa0799a029927eb9e5f74fa245d69e4edad/scripts/e2e-test.sh#L51

internal/integration/helpers.go
chaodaiG: Based on the discussion at https://github.com/google/exposure-notifications-verification-server/pull/841#discussion_r506455010, `INTEGRATION_DB_NAME` would work. Needs several additional changes at:

- Prefix `INTEGRATION_` the `DB_` vars:  https://github.com/google/exposure-notifications-server/blob/5ddd4aa0799a029927eb9e5f74fa245d69e4edad/scripts/e2e-test.sh#L47
- Add  `env:",prefix:INTEGRATION_"` at https://github.com/google/exposure-notifications-server/blob/5ddd4aa0799a029927eb9e5f74fa245d69e4edad/internal/e2e/e2e_test.go#L53

sethvargo:
@chaodaiG updated to prefix with `E2E_`- PTAnotherL

PR #1094, opened 2020-10-16
Incremental e2e test uses terraform instead of scripts

chaodaiG:
- Avoid export DB_* variables that are required by ./scripts/migrate
- Avoid deployment failure caused by newly added service: when a new service is added, just call ./scripts/deploy would fail due to newly created service in Cloud Run cannot be zero traffic


sethvargo (approved):
/lgtm

PR #1097, opened 2020-10-17
Terraform db instance explicitly depends_on db, cert and user

chaodaiG:
When terraform destroys database instance, it could happen that database instance was deleted ahead of db, cert and user, which were left orphaned and terraform didn't know how to handle them, and this prevents terraform smoke test from succeeding. Example error:

```
Error: Error when reading or editing SQL User "notification" in instance "en-server-kbcnp": googleapi: Error 400: Invalid request: Invalid request since instance is not running., invalid
```

https://prow.k8s.io/view/gcs/oss-prow/logs/ci-en-server-terraform-smoke/1317495728049754112#1:build-log.txt%3A227

### DIFF
--- a/terraform/database.tf
+++ b/terraform/database.tf
@@ -64,6 +64,11 @@ resource "google_sql_database_instance" "db-inst" {
 
   depends_on = [
     google_project_service.services["sql-component.googleapis.com"],
+    # Explicitly label dependencies of db instance, so that destroy won't
+    # delete db instance first, and leave the following resources orphaned.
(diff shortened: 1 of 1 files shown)


chaodaiG:
/cc @sethvargo 
Found the solution from https://github.com/hashicorp/terraform/issues/20196#issuecomment-502648235, and don't have a good way to test this out. Seth please advise with your terraform expertise.

sethvargo:
This looks good 

/lgtm

PR #1098, opened 2020-10-19
Terraform e2e: add back custom db deletion step

chaodaiG:
Unfortunately terraform doesn't like #1097 , saying cycling dependencies:

```
Error: Cycle: module.en.google_sql_user.user, module.en.google_sql_database.db, module.en.google_sql_database_instance.db-inst, module.en.google_sql_ssl_cert.db-cert
```

https://prow.k8s.io/view/gcs/oss-prow/logs/ci-en-server-terraform-smoke/1318009362538565632#1:build-log.txt%3A88

Restore custom db deletion step to make terraform smoke test work

### DIFF
--- a/scripts/ci-terraform.sh
+++ b/scripts/ci-terraform.sh
@@ -131,6 +131,8 @@ function destroy() {
   if [[ -n "${db_inst_name}" ]]; then
     echo "Deleting db ${db_inst_name}"
     gcloud sql instances delete ${db_inst_name} -q --project=${PROJECT_ID}
+    # There still might be open connection between vpc and db, wait to improve destroy
+    sleep 120
   fi
   # Clean up states after manual DB delete
   terraform state rm module.en.google_sql_user.user || best_effort
--- a/scripts/ci-terraform.sh
+++ b/scripts/ci-terraform.sh
@@ -121,6 +121,21 @@ function deploy() {
 function destroy() {
   pushd "${ROOT}/terraform-e2e-ci" > /dev/null
   init
+
+  # DB always failed to be destroyed by terraform as it's set to not to destroy,
+  # so delete it manually
+  local db_inst_name
+  # Fetching databases from previous terraform deployment output is not always reliable,
+  # especially when previous terraform deployment failed. So grepping from terraform state instead.
+  db_inst_name="$(terraform state show module.en.google_sql_database_instance.db-inst | grep -Eo 'en-server-[a-zA-Z0-9]+' | uniq)"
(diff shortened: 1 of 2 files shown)


sethvargo:
Hmmm - what happens without the depends_on? Terraform walks the graph _backwards_ when deleting, so it should delete all deps, then it should delete the database.

chaodaiG:
While testing locally, without `depends_on` I have got another error quite frequently. The detail error message was not captured but it's something like `database can not be deleted as it's used by another user, please delete with cli if this is not running as a pg super user`.

sethvargo:
Ah - that's not from Terraform, that's from Postgres. It basically means that there are open connections to the database.

chaodaiG:
Yeah that's what I thought too. I have tried to make db user depends on vpc connector expecting vpc connector being deleted prior to db user, but didn't get much luck

sethvargo:
Do we close the connections on the e2e runner? And I think Cloud Run already depends on the db, so those should be torn down first.

chaodaiG:
This is key server, I don't think there is an e2e test runner. I assume that's the case too, not sure if it's some gcloud stuff that utilizes the other user `postgres` doing something

chaodaiG:
@sethvargo , any other thoughts?

sethvargo:
Sorry - I don't think the screenshot came through.

chaodaiG:
Trying again. Just want to show the default user created for db:

Screen Shot 2020-10-21 at 2 49 09 PM

sethvargo:
Ironically, this is the thing that saved us during the last outage...

sethvargo:
Can we wait a bit longer after deleting the database? I wonder if the VPC has a connection open still.

chaodaiG:
> Can we wait a bit longer after deleting the database? I wonder if the VPC has a connection open still.

Not super clear, did you mean add `sleep X` after `gcloud` clean command?

sethvargo:
Yea. My untested suspicion is that Cloud Run and/or the VPC still have open connections to postgres, which is why the database is failing to delete. Waiting 2min should solve it.

chaodaiG:
> Yea. My untested suspicion is that Cloud Run and/or the VPC still have open connections to postgres, which is why the database is failing to delete. Waiting 2min should solve it.

Inserted a sleep, I haven't seen db instance deletion fail with `gcloud` yet, but I think this might help with flakiness I have seen that `terraform destroy` failed to delete vpc and got it into a bad state

chaodaiG:
@sethvargo PTAL

sethvargo (approved):
/lgtm

PR #1106, opened 2020-10-22
Export analyzer nil error

whaught:
## Proposed Changes

* func checkKeys to properly return nil on empty
* spelling


mikehelmick (approved):
must be a new test for the presbumit...

PR #1108, opened 2020-10-23
Allow server to bind to addresses and ports

sethvargo:
This is a **breaking change** to the server to allow specifying the bind address in address and port. 

It also fixes an issue when binding to the IPv4 or IPv6 loopback addresses, where the returned IP would be the anycast IP (0.0.0.0) instead of loopback (127.0.0.1).

**Release Note**

```release-note
Accept address _and_ port in `server.New()`
```


mikehelmick (approved):

sethvargo:
Seems like prow either doesnt like ipv6 or we're running out of descriptors when binding to a single IP. I'm going to work around this in the browser test harness instead.

PR #1109, opened 2020-10-23
factor out generate regions, add traveler generate

mikehelmick:
## Proposed Changes

* Generate service will only put keys in 1 region, each region passed in will have distinct keys
* introduce random change of a key being assigned traveler status.

**Release Note**

```release-note
Generate service mirrors v1 API more closely now.
```


sethvargo (approved):
/lgtm

PR #1117, opened 2020-10-28
Terrafrom smoke test: don't fail if failed to taint build/migrate

chaodaiG:
If previous run failed, it's possible that module.en.null_resource.build or module.en.null_resource.migrate was destroyed but not restalled, which could fail all subsequent runs


sethvargo (approved):
/lgtm

PR #1119, opened 2020-10-29
write generated export padding to the database, ensure overlapping ex…

mikehelmick:
…ports contain same data

## Proposed Changes

* Ensures that exports with overlapping time periods contain the same generated data by persisting the data to the database
* Obtain a lock over export regions when creating exports, all region locks are required, and they are sorted and obtained in order
* Batches are attempted to be processed in an order such that padding for earlier batches should be generated and thus included in later batches

**Release Note**

```release-note
Export padding is persisted to the database to ensure stability over time.
```

### DIFF
--- a/internal/database/lock.go
+++ b/internal/database/lock.go
@@ -52,9 +52,7 @@ func (db *DB) MultiLock(ctx context.Context, lockIDs []string, ttl time.Duration
 	var expires time.Time
 	err := db.InTx(ctx, pgx.Serializable, func(tx pgx.Tx) error {
 		for _, lockID := range lockOrder {
-			row := tx.QueryRow(ctx, `
-			SELECT AcquireLock($1, $2)
-		`, lockID, int(ttl.Seconds()))
+			row := tx.QueryRow(ctx, `SELECT AcquireLock($1, $2)`, lockID, int(ttl.Seconds()))
 			if err := row.Scan(&expires); err != nil {
 				return err
 			}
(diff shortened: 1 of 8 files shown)


mikehelmick:
ready to review but putting a 
/hold
while I check on some things

internal/export/worker.go
```diff
@@ -185,20 +201,70 @@ func (s *Server) exportBatch(ctx context.Context, eb *model.ExportBatch, emitInd
...
+			return nil, fmt.Errorf("ensureMinNumExposures: %w", err)
 		}
-		lastGroup.revised, err = ensureMinNumExposures(lastGroup.revised, eb.OutputRegion, s.config.MinRecords, s.config.PaddingRange)
-		if err != nil {
-			return fmt.Errorf("ensureMinNumExposures: %w", err)
+		// padding revised keys doesn't provide any useful protection as one can work backwords and figure out which
+		// keys appeared as primary keys in a previous export.
+
+		if length := len(generated); length > 0 {
+			log.Printf("generated %+v", generated)
```
sethvargo, line 217: Debug

internal/database/lock.go
```diff
@@ -35,6 +36,59 @@ var (
 // UnlockFn can be deferred to release a lock.
 type UnlockFn func() error
 
+// MultiLock obtains multiple locks in a single transaction. Either all locks are obtained, or
+// the transaction is rolled back.
+// The lockIDs are sorted by normal ascending string sort order before obtaining the locks.
+func (db *DB) MultiLock(ctx context.Context, lockIDs []string, ttl time.Duration) (UnlockFn, error) {
```
sethvargo, line 42: Is this so we can lock across multiple tables? It's unclear to me why we don't just use a transaction instead.

internal/export/worker.go
mikehelmick: removed

internal/database/lock.go
mikehelmick: added comments to worker.go where we take the locks

sethvargo (approved):
/lgtm
/hold

in case you have some other tweaks

mikehelmick:
/hold cancel

nope - this is good to go

PR #1120, opened 2020-10-29
update release cadence

mikehelmick:
## Proposed Changes

* change release cadence

**Release Note**

```release-note
Documented release cadence changed from weekly to bi-weekly.
```

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,8 +1,13 @@
 # Exposure Notifications Servers Release Process
 
-  * Releases are cut every Tuesday (target time 9am, US Pacific Time)
+  * Releases are cut every other 
+    * Monday for this server (target time 9am, US Pacific Time)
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -1,9 +1,14 @@
 # Exposure Notifications Servers Release Process
 
-  * Releases are cut every Tuesday (target time 9am, US Pacific Time)
+  * Releases are cut every other 
+    * Monday for this server (target time 9am, US Pacific Time)
```
sethvargo, line 4: Is there a reason we don't do them at the same time? I'd rather lump that toil into a single task:

- Relnotes key server, tag release
- Update deps in verification service
- Relnotes verification server, tag release

sethvargo (approved):
/lhtm

PR #1122, opened 2020-10-29
certificate verification improvements

### LINKED ISSUE #1121: Verification certificate aud not checked
### TL;DR

The issuer is currently checked, but the audience is not.
Given the tight configurations, this is unlikely to cause an issue, but should be fixed for correctness.

/assign

mikehelmick:
Fixes #1121

## Proposed Changes

* Cache not found issuers the same as found issuers
* Give an advisory warning for mismatched audience fields
* Prep for next release where audience fields are missing
* also, considerably speeds up the pha verify tests by reusing the database for all test cases

**Release Note**

```release-note
ATTENTION SERVER OPERATORS:
This release will print ERROR level log messages if there are verification certificates where the issuer matches, but the audience field does not match. In the next release these mismatches will be hard failures. If you find a mismatch, work with your verification server operators to resolve any issues before deploying the release after this one.
```

### DIFF
--- a/internal/verification/phaverify.go
+++ b/internal/verification/phaverify.go
@@ -61,7 +61,7 @@ type VerifiedClaims struct {
 // VerifyDiagnosisCertificate accepts a publish request (from which is extracts the JWT),
 // fully verifies the JWT and signture against what the passed in authorrized app is allowed
 // to use. Returns any transmission risk overrides if they are present.
-func (v *Verifier) VerifyDiagnosisCertificate(ctx context.Context, authApp *aamodel.AuthorizedApp, publish *verifyapi.Publish) (*VerifiedClaims, error) {
+func (v *Verifier) VerifyDiagnosisCertificate(ctx context.Context, authApp *aamodel.AuthorizedApp, publish *verifyapi.Publish, enforceAudienceMatch bool) (*VerifiedClaims, error) {
 	logger := logging.FromContext(ctx)
 	// These get assigned during the ParseWithClaims closure.
 	var healthAuthorityID int64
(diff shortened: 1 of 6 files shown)


internal/verification/phaverify.go
```diff
@@ -94,8 +102,19 @@ func (v *Verifier) VerifyDiagnosisCertificate(ctx context.Context, authApp *aamo
 			return nil, err
 		}
 
+		if cacheVal == nil {
```
sethvargo, line 105: This should never happen? I don't think we cache nil values.

internal/verification/phaverify.go
```diff
@@ -94,8 +102,19 @@ func (v *Verifier) VerifyDiagnosisCertificate(ctx context.Context, authApp *aamo
...
+			return nil, fmt.Errorf("issuer not found: %v", claims.Issuer)
+		}
+
 		ha := cacheVal.(*model.HealthAuthority)
 
+		// Advisory check the aud.
+		if claims.Audience != ha.Audience {
+			logger.Errorw("certifice audience mismatch - will be a failure in the next release", "claims.Aud", claims.Audience, "allowed", ha.Audience, "iss", ha.Issuer)
+			// Hard error will start in next release.
+			// return nil, fmt.Errorf("audience mismatch for issuer: %v", ha.Issuer)
```
sethvargo, line 115: Should we guard this with a feature flag or at least a greppable `TODO`?

internal/verification/phaverify.go
mikehelmick: See line 93 - the primary lookup function can return `nil` as a valid cached value now
the in memory cache implementation allows it.

internal/verification/phaverify.go
mikehelmick: added feature flag - will file issue for removal

sethvargo (approved):
/lgtm

@mikehelmick can you create an issue to remove this in the next release and block it on that milestone so we don't forget?

PR #1124, opened 2020-10-29
fix nit with multierror

whaught:
## Proposed Changes

* Follow-up from: https://github.com/google/exposure-notifications-server/pull/1123
* Sorry was too late to comment,  but I think this is the more common pattern.


sethvargo (approved):
/lgtm

PR #1131, opened 2020-10-30
adjust symptom onset handling

mikehelmick:
## Proposed Changes

* "unreasonable" symptom onset intervals < 28d ago by default will be treated as if 0 is passed in. 
* keys with a valid symptom onset interval but fall outside the range of -14d..14d will be dropped (restores old / correct behavior.) THIS IS A CHANGE - currently those keys would get adjusted to nil, which would cause default
* if NO symptom onset date is passed in, we tread the symptom onset day as 4 days ago and adjust the data on the TEKs accordingly

This modeling is being adjusted in consultation with the Apple+Google clinical teams working on EN.

**Release Note**

```release-note
BEHAVIOR CHANGE: The handling of symptom onset intervals is changing. Any symptom onsets older than 28 days ago (rounded down to UTC midnight) will be treated as invalid and as if 0 was passed in. Keys with a valid symptom onset days outside of -14..14 will be dropped. If there is not symptom onset passed, the symptom onset interval is set to 4 days ago.
Timing flags can be adjusted via environment variables. Default values are recommended.
```

### DIFF
--- a/internal/integration/helpers.go
+++ b/internal/integration/helpers.go
@@ -215,7 +215,7 @@ func NewTestServer(tb testing.TB) (*serverenv.ServerEnv, *Client) {
 	publishConfig.MaxKeysOnPublish = 15
 	publishConfig.MaxSameStartIntervalKeys = 2
 	publishConfig.MaxIntervalAge = 360 * time.Hour
-	publishConfig.CreatedAtTruncateWindow = 1 * time.Second
+	publishConfig.CreatedAtTruncateWindow = time.Second
 	publishConfig.ReleaseSameDayKeys = true
 	publishConfig.RevisionKeyCacheDuration = time.Second
 
--- a/internal/integration/helpers.go
+++ b/internal/integration/helpers.go
@@ -218,7 +218,6 @@ func NewTestServer(tb testing.TB) (*serverenv.ServerEnv, *Client) {
 	publishConfig.CreatedAtTruncateWindow = 1 * time.Second
 	publishConfig.ReleaseSameDayKeys = true
 	publishConfig.RevisionKeyCacheDuration = time.Second
-	publishConfig.UseDefaultSymptomOnset = false
 
 	publishHandler, err := publish.NewHandler(ctx, &publishConfig, env)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


mikehelmick:
/hold 
until monday - want to give people a chance to comment

internal/integration/helpers.go
```diff
@@ -207,18 +207,19 @@ func NewTestServer(tb testing.TB) (*serverenv.ServerEnv, *Client) {
...
+	publishConfig := publish.Config{}
+	envconfig.ProcessWith(ctx, &publishConfig, envconfig.OsLookuper())
+	// Make overrides.
+	publishConfig.RevisionToken = revConfig
+	publishConfig.MaxKeysOnPublish = 15
+	publishConfig.MaxSameStartIntervalKeys = 2
+	publishConfig.MaxIntervalAge = 360 * time.Hour
+	publishConfig.CreatedAtTruncateWindow = 1 * time.Second
+	publishConfig.ReleaseSameDayKeys = true
+	publishConfig.RevisionKeyCacheDuration = time.Second
```
whaught, line 220:
```suggestion
	publishConfig.RevisionKeyCacheDuration = time.Second
```

above you did `1 * time.Second` - let's be consistent with one or the other?

internal/publish/model/exposure_model.go
```diff
@@ -559,11 +560,32 @@ func (t *Transformer) TransformPublish(ctx context.Context, inData *verifyapi.Pu
 		BatchWindow:           t.truncateWindow,
 	}
 
-	onsetInterval := inData.SymptomOnsetInterval
-	// If the symtom onset interval provided on publish is too old to be relevant
-	// and one was provided in the verification certificate, take that one.
-	if onsetInterval < settings.MinStartInterval && claims != nil && claims.SymptomOnsetInterval > 0 {
-		onsetInterval = int32(claims.SymptomOnsetInterval)
+	// For validating the
```
whaught, line 563: is this comment truncated?

internal/publish/publish_test.go
```diff
@@ -1091,33 +1107,39 @@ func TestKeyRevision(t *testing.T) {
 				want := make(map[string]*model.Exposure)
 				revisedReportType := verifyapi.ReportTypeConfirmed
 				revisedTransmissionRisk := verifyapi.TransmissionRiskConfirmedStandard
+				sort.Slice(tc.Publish.Keys, func(i, j int) bool {
+					return tc.Publish.Keys[i].IntervalNumber >= tc.Publish.Keys[j].IntervalNumber
+				})
+				log.Printf("KEYS %+v", tc.Publish.Keys)
```
whaught, line 1113: Do we want to log this whole thing?

whaught (approved):

PR #1136, opened 2020-11-03
better handling of export-import URL config

### LINKED ISSUE #1135: Trailing slash on export_root
### TL;DR
Exportimporter does not add a "/" in between export_root and export filepath if one is missing when importing the exportfiles.

**Expected behavior**
Exportimporter should check if a "/" is included in the paths listed in an index.txt file, and if one is not there, add it on to the export_root before importing exports.

**Observed behavior**
Exportimporter will append whatever is in the index.txt file onto the xport_root, and will miss a "/" if its is not present either at the beginning of rows in index.txt or at the end of export_root.

### Reproduction
Create an exportimport config with no trailing "/" and import exports that do not begin with a "/"

mikehelmick:
Fixes #1135 

## Proposed Changes

* Better handling of export-importer config URLs

**Release Note**

```release-note
NONE
```

### DIFF
--- a/internal/exportimport/scheduler.go
+++ b/internal/exportimport/scheduler.go
@@ -115,19 +115,13 @@ func syncFilesFromIndex(ctx context.Context, db *exportimportdb.ExportImportDB, 
 			// drop blank lines.
 			continue
 		}
+
 		// Parse the export root to see if there is a defined path element.
 		base, err := url.Parse(config.ExportRoot)
 		if err != nil {
 			return 0, fmt.Errorf("config.ExportRoot is invalid: %s: %w", config.ExportRoot, err)
 		}
-		// If there is not a defined path element, we may need to add a "/" if
-		// the relative paths don't have a leading "/".
-		maybeAddSlash := base.Path == ""
-		relativePath := strings.TrimSpace(zipFile)
-		if maybeAddSlash && !strings.HasPrefix(relativePath, "/") {
-			relativePath = "/" + relativePath
-		}
-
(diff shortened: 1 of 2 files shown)


internal/exportimport/scheduler.go
```diff
@@ -115,7 +115,21 @@ func syncFilesFromIndex(ctx context.Context, db *exportimportdb.ExportImportDB,
 			// drop blank lines.
 			continue
 		}
-		proposedURL := fmt.Sprintf("%s%s", config.ExportRoot, strings.TrimSpace(zipFile))
+		// Parse the export root to see if there is a defined path element.
+		base, err := url.Parse(config.ExportRoot)
```
sethvargo, line 119: This is fine, but isn't this the logic that `path.Join` does? I'm confused why we need to implement this ourselves.

sethvargo (commented):
/lgtm
/hold
for a question

internal/exportimport/scheduler.go
jeremyfaller: *cough* unit test *cough*

(I hate being that guy...)

sethvargo: Please continue to be that person 😄

mikehelmick: still need to parse, because we only want to path.Join on the path elements, and not protocol.

added unit test to cover error conditions - happy path was well covered.

sethvargo (approved):
/lgtm

Issue #1036, opened 2020-09-26
Bump latency alert for key-rotation and cleanup jobs


sethvargo:
For background jobs like key-rotation and cleanup, we need to increase the allowed latency before paging. Right now everything is at 2s, but it's fine that these background jobs take 15s+ in some cases. I think we only want the 2s response time on in-path request services.

/cc @yegle @icco

sethvargo:
/close 

done in #1139

PR #1139, opened 2020-11-04
Add basic latency alerts as Terraform config

yegle:
#1036


terraform/alerting/alerts.tf
```diff
@@ -0,0 +1,58 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+locals {
+  slow_services = format("(%s)", join("|", [
+    "export",
```
mikehelmick, line 17: also needs cleanup services

yegle: export service actually rarely exceeds 10s at 99 percentile.

It looks like based on the last 1 month of data, only cleanup-exposure and cleanup-export should have 20s of alerting threshold.

image

terraform/alerting/alerts.tf
```diff
@@ -0,0 +1,59 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+locals {
+  slow_services = format("(%s)", join("|", sort([
+    "cleanup-export",
```
mikehelmick, line 17: export not being slow is an artifact of our setup (we may need to add it back)

yegle: I see. I added it to the SLOWEST tier instead.

mikehelmick (approved):

PR #1143, opened 2020-11-05
change default max records on export file to 500,000

mikehelmick:
## Proposed Changes

* The 30k number was based on daily infection rate in the US back in April. The theoretical max is more like 750,000 keys per export file, moving to 500k

**Release Note**

```release-note
DEFAULT CHANGE: Max records per export file default changed from 30,000 to 500,000
```


gurayAlsac (approved):
/lgtm
/approve

whaught (approved):

Issue #965, opened 2020-09-09
EFGS Federation Gateway support


marekaf:
Hello 👋 
Do you plan on supporting EFGS project somehow? (open sourced [with docs here](https://github.com/eu-federation-gateway-service/efgs-federation-gateway/blob/master/docs/software-design-federation-gateway-service.md), [architecture spec here](https://ec.europa.eu/health/sites/health/files/ehealth/docs/mobileapps_interop_architecture_en.pdf))

Briefly, the project is described as
```
The least complex and most robust way to connect the backends behind all the different
national proximity tracing apps is a Federation Gateway Service, which accepts diagnosis
keys from all countries, buffers them temporarily, and provides them for all countries to be
downloaded. Additionally, all backends can be informed immediately if new data is available,
so that transmission lags are kept minimal.
```

We are currently evaluating how can we extend key server's functionality to support EFGS and it would be great to know if there's any support / feature (in this repo or another) from Google already.

Thanks!

mikehelmick:
Hello -

We haven't had any requests yet to support this. If it is needed in your jurisdiction, it is something we can evaluate.

Let us know if this is a blocker for you and what kind of timeline would be needed.

marekaf:
hi @mikehelmick 
thanks for the answer,

There is a lot of European Union countries trying to implement this as soon as possible. We have integration/e2e testing scheduled in about 7-10 days, prod rollout before the end of month AFAIK.

Some countries are building their own key servers, we are trying to use upstream for as long as possible but this might force us to fork it. We are currently evaluating if the fork is needed or not while architecting the new solution.

Any help from you guys would be highly appreciated, it is quite a blocker for us. 

Thanks a lot!

mikehelmick:
I'm don't think we can meet that timeline of having this production ready by the end of the month. The EFGS federation protocol is quite different from the one that we already implement for federation.

If you wanted to contribute code back to this repo to enable that, we would be available to review (it requires assigning copyright to Google and singing our Common Licensing Agreement).

I think that we will need to support this eventually, but have a pretty full schedule in front of us right now.

marekaf:
I understand that. Honestly I don't understand the current federation implementation in key server - can't find proper docs for that. I'll try to dig into the code to see if we can reuse some of it or, as you say, reimplement it for EFGS. 

 At least I know now for sure that it is not compatible with EFGS, not sure why they decided to go 'their way'. 

[...]

PR #1145, opened 2020-11-06
remove use of EFGS DB columns

mikehelmick:
## Proposed Changes

ref: #965 

* Rollback initial EFGS work since we are not proceeding with integration right now. Will revisit in the future.

**Release Note**

```release-note
NONE
```


whaught (commented):
Is there a moment during release when we are still doing 
`VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17)`
but have dropped those columns?

Does this need to be split across releases?

mikehelmick:
> Is there a moment during release when we are still doing
> `VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17)`
> but have dropped those columns?
> 
> Does this need to be split across releases?

good point - removed the DB migrations from this.

mikehelmick:
filed #1146 to remove the columns in v0.17

whaught (approved):

PR #1149, opened 2020-11-07
Remove unused ctx from HandlHealthz function.

yegle:


sethvargo (changes requested):
I actually need that context as part of https://github.com/google/exposure-notifications-verification-server/pull/1012

yegle:
> I actually need that context as part of [google/exposure-notifications-verification-server#1012](https://github.com/google/exposure-notifications-verification-server/pull/1012)

See my comment in that PR. I suspect the `ctx` there is also not actually used and can be omitted. We should use the `ctx` from `http.Request` not the `ctx` from when the server is initialized.

PR #1150, opened 2020-11-09
Retry failed imports.

### LINKED ISSUE #1134: Exportimporter leaves importfiles OPEN or PENDING
### TL;DR
If an importfile is unable to be ingested by exportimporter (due to incorrect formatting of export url, etc), it sometimes remains in "PENDING" status instead of failing.

**Expected behavior**
Import should fail after X number of tries.

**Observed behavior**
Importfile left in PENDING status.

@mikehelmick

jeremyfaller:
Fixes #1134

## Proposed Changes

* Retry failed imports up to a configurable amount.

**Release Note**

```release-note
Failed imports will be retried up to a configurable amount (default = 5).
```

### DIFF
--- a/internal/exportimport/config.go
+++ b/internal/exportimport/config.go
@@ -51,6 +51,9 @@ type Config struct {
 	// Each exposure is inserted with the app_package_name / healthAuthorityID that it was published with
 	// Use this string to signal that a key came from the export-importer job.
 	ImportAPKName string `env:"IMPORT_APP_PACKAGE_NAME, default=exportimport"`
+	// ImportRetryRate is the rate at which files that encounter an error while
+	// importing are retried.
+	ImportRetryRate time.Duration `env:"IMPORT_RETRY_RATE, default=6h"`
 }
 
 func (c *Config) DatabaseConfig() *database.Config {
--- a/internal/exportimport/database/exportimport.go
+++ b/internal/exportimport/database/exportimport.go
@@ -317,9 +317,11 @@ func prepareInsertImportFile(ctx context.Context, tx pgx.Tx) (string, error) {
 	return stmtName, err
 }
 
-func (db *ExportImportDB) CreateFiles(ctx context.Context, ei *model.ExportImport, filenames []string) (int, error) {
+// CreateNewFilesAndFailOld creates all the specified files named, returning
+// the number of created files, and the number moved to an failed state.
+func (db *ExportImportDB) CreateNewFilesAndFailOld(ctx context.Context, ei *model.ExportImport, filenames []string) (int, int, error) {
 	logger := logging.FromContext(ctx)
-	insertedFiles := 0
+	insertedFiles, failedFiles := 0, 0
 
 	now := time.Now().UTC()
 	if err := db.db.InTx(ctx, pgx.ReadCommitted, func(tx pgx.Tx) error {
@@ -369,12 +371,26 @@ func (db *ExportImportDB) CreateFiles(ctx context.Context, ei *model.ExportImpor
 			logger.Debugw("scheduled new export file for importing", "exportImportID", ei.ID, "filename", fname)
 			insertedFiles++
 		}
+
+		// Close any files that aren't in the list that are retrying.
+		failed, err := tx.Exec(ctx, `
+				UPDATE
+					ImportFile
+				SET
+					status = $1
+				WHERE export_import_id=$2 AND NOT zip_filename = ANY($3) AND status != $4
+			`, model.ImportFileFailed, ei.ID, filenames, model.ImportFileComplete)
+		if err != nil {
+			return fmt.Errorf("failed to update retries: %w", err)
+		}
+		failedFiles = int(failed.RowsAffected())
+
 		return nil
 	}); err != nil {
-		return 0, fmt.Errorf("creating import files: %w", err)
+		return 0, 0, fmt.Errorf("creating import files: %w", err)
 	}
 
-	return insertedFiles, nil
+	return insertedFiles, failedFiles, nil
 }
 
 func (db *ExportImportDB) CompleteImportFile(ctx context.Context, ef *model.ImportFile, status string) error {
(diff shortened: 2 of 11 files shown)


migrations/000048_AddRetries.up.sql
```diff
@@ -0,0 +1,25 @@
...
+
+BEGIN;
+
+ALTER TYPE ImportFileStatus ADD VALUE 'RETRY';
+
+ALTER TABLE importfile
+    ADD COLUMN retries int DEFAULT 0;
+
+ALTER TABLE exportfile
+    ADD COLUMN retries int DEFAULT 0;
```
mikehelmick, line 23: why is this being added to exportfile it isn't being used?

internal/exportimport/config.go
```diff
@@ -51,6 +51,9 @@ type Config struct {
 	// Each exposure is inserted with the app_package_name / healthAuthorityID that it was published with
 	// Use this string to signal that a key came from the export-importer job.
 	ImportAPKName string `env:"IMPORT_APP_PACKAGE_NAME, default=exportimport"`
+	// If importing a file encounters an error, we will retry a number of times
+	// before stopping.
+	MaxImportRetries uint `env:"MAX_IMPORT_RETRIES, default=5"`
```
mikehelmick, line 56: this is different than what I had suggested in the bug.

I think a low # of retries without a backoff strategy will cause a pretty fast fail. (25 minutes w/ default terraform config - job runs every 5 minutes).

I don't think we should stop retrying that aggressively as then it falls to a human to notice and move the file back to pending.

internal/exportimport/importer.go
```diff
@@ -137,10 +146,15 @@ func (s *Server) runImport(ctx context.Context, config *model.ExportImport) erro
 		if err := s.exportImportDB.CompleteImportFile(ctx, file, status); err != nil {
 			logger.Errorf("failed to mark file completed", "file", file, "error", err)
 		}
+
+		// TODO: Do we want to keep trying files rather than returning?
+		//       This seems like we quit awfully early.
+		if err != nil {
+			return fmt.Errorf("error processing import file: %w", err)
```
mikehelmick, line 153: I think we want to keep processing.

There is a race condition where the oldest files might be cleaned up before the index is rewritten and the oldest files may be attempted to be imported first.

migrations/000048_AddRetries.up.sql
jeremyfaller: my mistake. fixed.

internal/exportimport/importer.go
jeremyfaller: fixed.

internal/exportimport/config.go
jeremyfaller: Ok, so if I had some backoff, would that be adequate?

jeremyfaller: Ok, so we now do some backoff, and if the file is no longer present in the index, we mark it as failed.

migrations/000048_AddRetries.up.sql
```diff
@@ -0,0 +1,22 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+ALTER TYPE ImportFileStatus ADD VALUE 'RETRY';
```
mikehelmick, line 17: this shouldn't be needed... will look at rest of PR first

internal/exportimport/database/exportimport.go
```diff
@@ -369,6 +369,19 @@ func (db *ExportImportDB) CreateFiles(ctx context.Context, ei *model.ExportImpor
...
 			insertedFiles++
 		}
+
+		// Close any files that aren't in the list that are retrying.
+		_, err = tx.Exec(ctx, `
+				UPDATE
+					ImportFile
+				SET
+					status = $1
+				WHERE export_import_id=$2 AND NOT zip_filename = ANY($3) AND status = $4
```
mikehelmick, line 379: I think this should just be AND status != complete

if the file was never tried and no longer appears in the index (however unlikely that is) it should still move to a failed state

internal/exportimport/database/exportimport.go
```diff
@@ -511,9 +525,56 @@ func (db *ExportImportDB) GetOpenImportFiles(ctx context.Context, lockDuration t
...
+	}); err != nil {
+		return nil, fmt.Errorf("unable to read open import files: %w", err)
+	}
+
+	return importFiles, nil
+}
+
+// GetAllInputFiles returns all input files for a config, regardless of their state.
+// This function is used for testing.
+func (db *ExportImportDB) GetAllInputFiles(ctx context.Context, lockDuration time.Duration, ei *model.ExportImport) ([]*model.ImportFile, error) {
```
mikehelmick, line 547: s/Input/Import/ ?

internal/exportimport/model/file.go
```diff
@@ -30,4 +31,24 @@ type ImportFile struct {
...
 	Status         string
+	Retries        uint
+}
+
+// ShouldTry performs some introspection on an import file from the DB, and
+// returns true if that file should be tried for download.
+func (f *ImportFile) ShouldTry() bool {
+	if f.Status == ImportFileRetry {
+		// ProcessedAt should be non-null for all codepaths that set status to retry.
+		d := *f.ProcessedAt
```
mikehelmick, line 42: i would like to see the retry parameters configurable and seems like all retry units should be just a time function

migrations/000048_AddRetries.up.sql
mikehelmick: I'm fine w/ this staying, but I think the extra state made the pr slightly more complex than it needed to be.
having the state be pending w/ retries>0 indicates the same thing.

no need to change

internal/exportimport/model/file.go
jeremyfaller: This should do it fer ya!

migrations/000048_AddRetries.up.sql
jeremyfaller: removed.

internal/exportimport/database/exportimport.go
```diff
@@ -369,6 +369,19 @@ func (db *ExportImportDB) CreateFiles(ctx context.Context, ei *model.ExportImpor
 			logger.Debugw("scheduled new export file for importing", "exportImportID", ei.ID, "filename", fname)
 			insertedFiles++
 		}
+
+		// Close any files that aren't in the list that are retrying.
+		_, err = tx.Exec(ctx, `
```
mikehelmick, line 374: it would be good to return the number of files that moved into a failed state and log this information

mikehelmick (approved):

PR #1151, opened 2020-11-09
Do not call CreateMetricDescriptor

yegle:
Rely on Terraform to manage metric descriptors.


sethvargo (commented):
What's the behavior if this is true, but no metric is defined. Is it an error?

yegle:
> What's the behavior if this is true, but no metric is defined. Is it an error?

Hmm I just realized this won't work, as Cloud Monitoring will create the metric automatically (http://cloud/monitoring/custom-metrics/creating-metrics#auto-creation).

It looks like we will be in a tough spot if we **really** want to recreate a metric.

PR #1060, opened 2020-10-06
allow for default days since symptom onset

mikehelmick:
Fixes #1055

## Proposed Changes

* By default, if no value is known for days since symptom onset is known, set it to the default value in the system.
* 10 is the agreed upon default by the Google and Apple clinical teams

**Release Note**

```release-note
BEHAVIOR CHANGE: 
Two new environment variables, USE_DEFAULT_SYMPTOM_ONSET_DAYS and DEFAULT_SYMPTOM_ONSET_DAYS. 
If enabled, TEKs that previously would have a null value for days since symptom onset will receive the default value.

This new functionality is enabled by default in this release.
```

PR #1158, opened 2020-11-11
Register ocsql views in observability package.

yegle:
Conditionally register those views only when ocsql driver is registered.

Part of #1060


mikehelmick:
/retest

after integration test fix

mikehelmick (approved):

PR #1159, opened 2020-11-11
Make api requests return chaffed json data.

### LINKED ISSUE #1155: Return JSON response bodies on Chaff requests
### TL;DR

The go-chaff library already supports this
  https://github.com/mikehelmick/go-chaff/blob/main/json.go

Use this to respond w/ properly formed JSON based on the request being chaffed - for our APIs, this really just means having a `padding` key in the JSON object.

REMEMBER - the size of the random data will need to be reduced as base 64 encoding will cause ~33% size increase.

/priority important-soon
/kind enhancement

jeremyfaller:
Fixes #1155

```release-note
Chaff v1 api requests now return chaffed JSON data.
```

### DIFF
--- a/internal/publish/publish.go
+++ b/internal/publish/publish.go
@@ -98,12 +98,17 @@ func NewHandler(ctx context.Context, config *Config, env *serverenv.ServerEnv) (
 		return nil, fmt.Errorf("config validation: %w", err)
 	}
 
+	chaffer, err := chaff.NewTracker(chaff.NewJSONResponder(chaffPublishResponse), chaff.DefaultCapacity)
+	if err != nil {
+		return nil, fmt.Errorf("error making chaffer: %v", err)
+	}
+
 	return &PublishHandler{
 		serverenv:             env,
 		transformer:           transformer,
 		config:                config,
 		database:              database.New(env.Database()),
-		tracker:               chaff.New(),
+		tracker:               chaffer,
 		tokenManager:          tm,
 		tokenAAD:              aadBytes,
 		authorizedAppProvider: env.AuthorizedAppProvider(),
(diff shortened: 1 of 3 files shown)


internal/publish/publish.go
```diff
@@ -444,3 +449,9 @@ func (h *PublishHandler) process(ctx context.Context, data *verifyapi.Publish, b
 		},
 	}
 }
+
+// chaffPushResponse takes a chaffing string, and builds a chaff response.
+func chaffPublishResponse(s string) interface{} {
+	return verifyapi.PublishResponse{Padding: s[:(len(s)*7)/10]}
```
mikehelmick, line 455: the chaff library already accounts for base64 encoding

mikehelmick: and doing a sub-slice here might make this invalid base64

mikehelmick (changes requested):

internal/publish/publish.go
jeremyfaller: I didn't realize anyone was ever going to decode the chaff. In any case, removed.

internal/publish/publish.go
```diff
@@ -444,3 +449,9 @@ func (h *PublishHandler) process(ctx context.Context, data *verifyapi.Publish, b
 		},
 	}
 }
+
+// chaffPushResponse takes a chaffing string, and builds a chaff response.
+func chaffPublishResponse(s string) interface{} {
+	return verifyapi.PublishResponse{Padding: s}
+
```
mikehelmick, line 456: nit: extra blank line

jeremyfaller: done.

mikehelmick (approved):

PR #1161, opened 2020-11-12
allow a revised key to come in again, as long is it no changey

mikehelmick:
## Proposed Changes

* If a revised key is uploaded again and the reportType hasn't changed, don't error.

**Release Note**

```release-note
 If a revised key is uploaded again and the reportType hasn't changed, don't error.
```

### DIFF
--- a/internal/publish/database/exposure_test.go
+++ b/internal/publish/database/exposure_test.go
@@ -577,6 +577,9 @@ func TestReviseExposures(t *testing.T) {
 			t.Errorf("expected %d to be %d: %#v", got, want, reviseResp)
 		}
 
+		// Change report, type, attempt to revise again.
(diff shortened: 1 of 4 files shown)


sethvargo:
Is this to work around a client issue? It feels like a weird behavior to enable...

mikehelmick:
> Is this to work around a client issue? It feels like a weird behavior to enable...

Currently - if a client attempts to upload unchanged keys multiple times that works.
And if you revise, and then try to keep uploading the revised state, it doesn't work.

This is necessary to enable additional key uploads after you have revised to a confirmed diagnosis from likely.

Not a CUJ any clients are covering, but it is one the server supports.

whaught (approved):

PR #1162, opened 2020-11-12
Add a grace period between removal from the export indices, and deletion

### LINKED ISSUE #1148: export cleanup should regenerate index.txt
### TL;DR

When the cleanup job runs, it deletes from the database, and then removes from storage old files.

Those old files continue to appear in the index.txt until the next export runs.

**Expected behavior**

All files that appear in index.txt are not deleted by our application.

**Suggested fixes**

2 possible fixes.

1) determine which files are to be deleted. regnerate the index.txt file without those files. then delete the files.

2) move cleanup to more of a mark and sweep. after X days, mark the file as stale in the database. modify export index generation to skip stale files. after X+Y days do the normal cleanup.

The first path is more correct, the second path is possibly easier to implement.

/priority important-soon

jeremyfaller:
Fixes #1148

**Release Note**

```release-note
Add a 1 hour grace time between when a file is no longer included in the export file list, and when it's actually deleted.
```

### DIFF
--- a/internal/cleanup/cleanup.go
+++ b/internal/cleanup/cleanup.go
@@ -34,6 +34,10 @@ import (
 
 const (
 	minTTL = 10 * 24 * time.Hour
+
+	// exportTTLGrace is the window of time between when a file will be listed in
+	// the export batch, and when it will actually be deleted.
+	exportTTLGrace = 1 * time.Hour
 )
 
 // NewExposureHandler creates a http.Handler for deleting exposure keys
@@ -167,5 +171,5 @@ func cutoffDate(ctx context.Context, d time.Duration, overrideMinTTL bool) (time
 			return time.Time{}, fmt.Errorf("cleanup ttl is less than configured minimum ttl")
 		}
 	}
-	return time.Now().Add(-d), nil
+	return time.Now().Add(-(d + exportTTLGrace)), nil
 }
--- a/internal/cleanup/cleanup_test.go
+++ b/internal/cleanup/cleanup_test.go
@@ -147,7 +147,7 @@ func TestCutoffDate(t *testing.T) {
 			} else if err != nil {
 				t.Errorf("%q: got error %v", test.d, err)
 			} else {
-				want := now.Add(-test.wantDur)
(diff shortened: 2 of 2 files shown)


mikehelmick (changes requested):
the grace period method doesn't work with long batch periods (i.e. 24 hours)
the index file is at best regenerated ever time a batch runs (i.e. every 4 hours in a typical production instance)

we need to guarantee that a file isn't deleted while it is still in the index and this shortens, but doesn't eliminate the race

jeremyfaller:
> the grace period method doesn't work with long batch periods (i.e. 24 hours)
> the index file is at best regenerated ever time a batch runs (i.e. every 4 hours in a typical production instance)
> 
> we need to guarantee that a file isn't deleted while it is still in the index and this shortens, but doesn't eliminate the race

Sorry. When I read the TF, it looked to me like the indices were being generated every minute. Can you quickly drop a pointer to the section of code I'm apparently misunderstanding.

Thanks, and sorry for the confusion.

mikehelmick:
> > the grace period method doesn't work with long batch periods (i.e. 24 hours)
> > the index file is at best regenerated ever time a batch runs (i.e. every 4 hours in a typical production instance)
> > we need to guarantee that a file isn't deleted while it is still in the index and this shortens, but doesn't eliminate the race
> 
> Sorry. When I read the TF, it looked to me like the indices were being generated every minute. Can you quickly drop a pointer to the section of code I'm apparently misunderstanding.
> 
> Thanks, and sorry for the confusion.

the job gets a chance to run once a minute, yes, but most of the time has no work to do.

if there are no open batches - then the index file isn't re-written

https://github.com/google/exposure-notifications-server/blob/main/internal/export/worker.go#L79

jeremyfaller:
Ah, ok, so as I see it, we could take an "easy" way out, and just bump this gracePeriod to be larger than maxexportbatchtime, or actually do a mark and sweep. (I was hoping to avoid that as I'm a little unclear on all the different pieces here, and what the cleanup job can trigger in the main server.) Unless just bumping (and making configurable) this gracePeriod is acceptable, I'll think a bit more about how to mark and sweep.

/hold

PR #1163, opened 2020-11-12
Incremental e2e tests: use latest service for running tests

chaodaiG:
Just discovered that incremental e2e test had been creating images and migrating db, but didn't really deploy new services, so the test had been testing the old services from last month. This PR should fix it

Ported from  https://github.com/google/exposure-notifications-verification-server/pull/1067


mikehelmick (approved):

sethvargo (approved):
/lgtm

PR #1167, opened 2020-11-14
allow export filename rewriting in mirror job

### LINKED ISSUE #1164: mirror filename rewrite
Allow for the mirror service to rewrite filenames into different patterns.

/assign

mikehelmick:
Fixes #1164 

## Proposed Changes

* Allow rewriting of filenames during mirror
* process mirrors in remote index order

**Release Note**

```release-note
cmd/mirror only : allow for mirrors to generate different filenames than what is upstream. Two patterns are accepted [timestamp] and [uuid] - uses of timestamp will always have increasing values within an individual mirror config.
```

### DIFF
--- a/cmd/mirror/main.go
+++ b/cmd/mirror/main.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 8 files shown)


sethvargo (approved):
/lgtm

PR #1169, opened 2020-11-16
Delete serverless vpc connector ahead of time

chaodaiG:
serverless-vpc-connector had been into bad state quite a few times after deleted by terraform, terraform remove it from state with the existence of a broken vpc connector, which prevents future terraform apply due to `serverless vpc connector already exist`. Forcing delete to avoid manual intervention


sethvargo (approved):
/lgtm

PR #1171, opened 2020-11-18
Add a mark/sweep to ExportFile deletion.

### LINKED ISSUE #1148: export cleanup should regenerate index.txt

jeremyfaller:
Fixes #1148

This PR replaces PR #1162

**Release Note**

```release-note
ExportFiles won't be deleted until they're removed from the index.html
```

### DIFF
--- a/internal/export/database/export.go
+++ b/internal/export/database/export.go
@@ -683,6 +683,38 @@ func (db *ExportDB) FinalizeBatch(ctx context.Context, eb *model.ExportBatch, fi
 	})
 }
 
+// MarkExpiredFiles marks files for deletion.
+func (db *ExportDB) MarkExpiredFiles(ctx context.Context, configID int64, ttl time.Duration) (int, error) {
+	var filesToDelete int
+	return filesToDelete, db.db.InTx(ctx, pgx.ReadCommitted, func(tx pgx.Tx) error {
+		minTime := time.Now().Add(-1 * ttl)
+		res, err := tx.Exec(ctx, `
+		UPDATE
+			ExportFile AS ef
+		SET
+			status = $5
+		FROM
+			ExportBatch AS eb
+		WHERE
+			eb.config_id = $1
+		AND
+			ef.batch_id = eb.batch_id
+		AND
+			eb.end_timestamp < $2
+		AND
+			eb.status = $3
+		AND
+			ef.status = $4
+		`,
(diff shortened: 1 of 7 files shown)


internal/export/database/export.go
```diff
@@ -683,6 +683,38 @@ func (db *ExportDB) FinalizeBatch(ctx context.Context, eb *model.ExportBatch, fi
...
+		SET
+			status = $5
+		FROM
+			ExportBatch AS eb
+		WHERE
+			eb.config_id = $1
+		AND
+			ef.batch_id = eb.batch_id
+		AND
+			eb.start_timestamp < $2
```
mikehelmick, line 703: this is inconsistent w/ current cleanup, see DeleteFilesBefore

mikehelmick: current cleanup based on batch end_timestamp, not start

internal/export/worker.go
```diff
@@ -391,6 +391,12 @@ func (s *Server) retryingCreateIndex(ctx context.Context, eb *model.ExportBatch,
 			}
 			return fmt.Errorf("creating index file for batch %d: %w", eb.BatchID, err)
 		}
+
+		// Mark files that we've previously cared about as expired.
+		if err := s.markExpiredFiles(ctx, eb); err != nil {
```
mikehelmick, line 396: I feel like this call should come before the call to createIndex so that we can mark the files expired, and then create the index.

internal/export/database/export.go
jeremyfaller: my mistake. fixed.

internal/export/worker.go
jeremyfaller: The two sets *shouldn't* overlap, but I think that's kosher. moved.

mikehelmick (approved):

jeremyfaller:
/retest

one more time.

PR #1176, opened 2020-11-19
Fix for batching when there are more TEKs than the max records settin…

mikehelmick:
…g per batch

Fixes #

## Proposed Changes

* Ensure that TEK groups for a large export batch are distinct TEKs and not duplicates. 
* Increase test coverage

**Release Note**

```release-note
BUG FIX: When using a low number for max records per export file and there are a high number of records per export file, ensure that the same exposures aren't duplicated across all files in the batch.
```


sethvargo (approved):
/lgtm

PR #1181, opened 2020-11-20
change filename pattern, always use batch size of 1

mikehelmick:
## Proposed Changes

* Change filename to include first 6 HEX digits of SHA256 of the TEK protobuff data.
* Load all keys, sort and then group to ensure stability over time

**Release Note**

```release-note
IMPORTANT: Export file naming and batching has changed. All batches going forward will always show all files as 1 of 1 even if there are multiple files in the batch. The sorting and grouping has changes to ensure that keys are sorted over an entire batch window and not just over the group. This bring stability over time assuming the same export parameters.
```


internal/export/exportfile.go
```diff
@@ -51,43 +52,46 @@ type Signer struct {
...
+func MarshalExportFile(eb *model.ExportBatch, exposures, revisedExposures []*publishmodel.Exposure, batchNum, batchSize int, signers []*Signer) ([]byte, string, error) {
 	// create main exposure key export binary
 	expContents, err := marshalContents(eb, exposures, revisedExposures, int32(batchNum), int32(batchSize), signers)
 	if err != nil {
-		return nil, fmt.Errorf("unable to marshal exposure keys: %w", err)
+		return nil, "", fmt.Errorf("unable to marshal exposure keys: %w", err)
 	}
 
+	digest := sha256.Sum256(expContents)
+	pbSHA := hex.EncodeToString(digest[:])
```
sethvargo, line 64: Somewhat nit, but it doesn't make sense to calculate this until later when it's needed.

internal/export/worker.go
```diff
@@ -433,8 +451,9 @@ func (s *Server) createIndex(ctx context.Context, eb *model.ExportBatch, newObje
 	return indexObjectName, len(objects), nil
 }
 
-func exportFilename(eb *model.ExportBatch, batchNum int) string {
-	return fmt.Sprintf("%s/%d-%d-%05d%s", eb.FilenameRoot, eb.StartTimestamp.Unix(), eb.EndTimestamp.Unix(), batchNum, filenameSuffix)
+func exportFilename(eb *model.ExportBatch, groupNum int, pbHexSHA string) string {
+	first6 := pbHexSHA[0:6]
```
sethvargo, line 455: Let's sanity check this with bounds

sethvargo (approved):
/lgtm

PR #1183, opened 2020-11-21
make signature block match

mikehelmick:
## Proposed Changes

* signature block of bathNum/batchSize 1/1. needs to match file contents

**Release Note**

```release-note
fix batch num/size in signature block
```


sethvargo (approved):
/lgtm

PR #1190, opened 2020-11-24
restore the original timeline

mikehelmick:
## Proposed Changes

* Restore default export filenames to be what they were last week. Too many deployed haps have taken dependencies on them.
* Introduce `REPROCESS_COUNT` to compensate - will allow an operator to cause filenames to be different (off by 1)

**Release Note**

```release-note
Restores filename patterns to their original format and provides instructions for regenerating export batches.
App developers should not depend on the filename pattern.
```

### DIFF
--- a/internal/export/config.go
+++ b/internal/export/config.go
@@ -52,7 +52,7 @@ type Config struct {
 	TruncateWindow     time.Duration `env:"TRUNCATE_WINDOW, default=1h"`
 	MinWindowAge       time.Duration `env:"MIN_WINDOW_AGE, default=2h"`
 	TTL                time.Duration `env:"CLEANUP_TTL, default=336h"`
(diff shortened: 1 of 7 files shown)


internal/export/config.go
```diff
@@ -52,6 +52,13 @@ type Config struct {
 	TruncateWindow     time.Duration `env:"TRUNCATE_WINDOW, default=1h"`
 	MinWindowAge       time.Duration `env:"MIN_WINDOW_AGE, default=2h"`
 	TTL                time.Duration `env:"CLEANUP_TTL, default=336h"`
+	// ReprocessCount needs to be incremented by one ever time you go back and
```
sethvargo, line 55:
```suggestion
	// ReprocessCount needs to be incremented by one every time you go back and
```

sethvargo (approved):
/lgtm

PR #1191, opened 2020-11-24
Adjust the alerting threshold to 10min.

yegle:
The export service sometimes takes more than 8min to finish. We can use
10min for now and see if it will lower the noise. If it's still noisy we
can just exclude the service from latency alert and add a request
error rate alert for it.


mikehelmick:
It's noisy because I'm regenerating 5 days of exports

mikehelmick (approved):
/retest

PR #1197, opened 2020-11-29
Use a single container for database tests

sethvargo:
Well, it's not exactly a single container, but more a single container per test package. Go tests are run in different processes, so it's not possible to share a single global container, but we share it with all tests in a single package, about 12 containers in total.

Previously we were spinning up a new container, starting the database, running migrations, seeding data, running tests, and stopping the container. This led to resource exhaustion and flakey tests. This new design involves slightly more manual work: once for each package that wants a database, you have to manually add 5 LOC to start/stop the container for that package. The tests run in 1/3 of the time though, and I've had to turn my furnace back on in the house since my CPU no longer provides enough enthalpy to heat the entire house when running the tests (joking, this is a good thing).

As an even further optimization, migrations are only run once per package. When the container first starts, we connect, run migrations, and then save that database as a "template" database. When an actual test needs a database, we clone that database (which has the complete schema and any test data) which is a near-instant operation. Each test is still fully isolated, but we only eat the cost of the database migration once (per package). The overhead of cloning a database amortizes over just 2 tests.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,15 @@
 
 VETTERS = "asmdecl,assign,atomic,bools,buildtag,cgocall,composites,copylocks,errorsas,httpresponse,loopclosure,lostcancel,nilfunc,printf,shift,stdmethods,structtag,tests,unmarshal,unreachable,unsafeptr,unusedresult"
 GOFMT_FILES = $(shell go list -f '{{.Dir}}' ./... | grep -v '/pb')
+GO_FILES = $(shell find . -name \*.go)
+
+copyrightcheck:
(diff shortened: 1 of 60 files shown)


sethvargo:
🙏 Prow

sethvargo:
Okay, I've run these tests 10 times now and got no failures, so I'm 10/10 confident this will fix our flakes.

/cc @mikehelmick 
/cc @chaodaiG 
/cc @whaught

chaodaiG:
This is awesome!
/lgtm

sethvargo:
@chaodaiG can you turn down the periodic runner since we fixed the flakiness now?

PR #1198, opened 2020-11-30
Performance test script also pulls images from GCP artifact registry

chaodaiG:


sethvargo (approved):
/lgtm

sethvargo:
@chaodaiG does the performance test run with a different service account?

chaodaiG:
Yeah it uses a different service account, but that shouldn't be the reason for presubmit test failure as the changes in this PR is not exercised in presubmit test at all

chaodaiG:
It's very likely because the prow job uses `GOOGLE_APPLICATION_CREDENTIALS` when it's set, as https://github.com/GoogleCloudPlatform/oss-test-infra/blob/538b6d9ae82637e9e6c474937f0eebe02d338b96/prow/prowjobs/google/exposure-notifications-server/en-server.yaml#L19. Let me see if `GOOGLE_APPLICATION_CREDENTIALS` is needed or not

chaodaiG:
I think `GOOGLE_APPLICATION_CREDENTIALS` is required by https://github.com/google/exposure-notifications-server/blob/7765d03485d0d0c484f970cbaa172c26a72190ce/internal/storage/google_cloud_storage_test.go#L111, can we give the service account associated with this secret permission on artifact registry? @sethvargo

PR #1205, opened 2020-12-02
Enable pgaudit

sethvargo:
Gracefully fail if the extension is not available

**Release Note**

```release-note
Enable pgaudit
```

/assign @mikehelmick


mikehelmick (approved):

mikehelmick (approved):

sethvargo:
/retest

looks like we're stuck?

PR #1208, opened 2020-12-03
Unit test testing large batch exports

### LINKED ISSUE #1202: Integration tests - test large / split batches
The integration tests currently don't have coverage over large batches / split batches.

chaodaiG:
Fixes #1202

### DIFF
--- a/internal/export/worker_test.go
+++ b/internal/export/worker_test.go
@@ -153,11 +153,11 @@ func TestBatchExposures(t *testing.T) {
 		count int
 	}{
 		{
-			"base case",
+			"base_case",
 			4,
 		},
 		{
-			"a lot",
+			"a_lot",
 			1000,
 		},
 	}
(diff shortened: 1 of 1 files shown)


internal/export/worker_test.go
```diff
@@ -148,220 +148,239 @@ func getKey(t *testing.T) []byte {
...
-	}
-	server := Server{
-		config: &config,
-		env:    serverenv.New(ctx, serverenv.WithDatabase(testDB)),
+	tests := []struct {
+		name  string
+		count int
+	}{
+		{
+			"base case",
```
sethvargo, line 156: Nit: don't use spaces in test case names

sethvargo (approved):

internal/export/worker_test.go
chaodaiG: Done. Please add another lgtm

PR #1212, opened 2020-12-06
Increase p50 on cleanup-export to 60s

sethvargo:
**Release Note**

```release-note
Increase p50 on cleanup-export to 60s
```


whaught (approved):

mikehelmick:
the latency here isn't super important - and will get worse on a large multi-tenant installation.

Instead we should measure if we're getting everything cleaned up or not.

/cc @yegle

PR #1215, opened 2020-12-07
allow for max records to be overridden on a per-export-config basis

### LINKED ISSUE #1214: Allow max records on export to be configurable per-exportconfig
### TL;DR

Allow max records to be overridden in the exportconfig

/assign

mikehelmick:
Fixes #1214 

## Proposed Changes

* Allow exportconfig entries to override max records

**Release Note**

```release-note
Allow for individual exportconfigs to have variable sizes for the max records setting.
```

### DIFF
--- a/internal/admin/exports/form.go
+++ b/internal/admin/exports/form.go
@@ -29,19 +29,20 @@ import (
 var ErrCannotSetBothTravelers = errors.New("cannot have both 'include travelers', and 'only non-travelers' set")
 
 type formData struct {
-	OutputRegion     string        `form:"OutputRegion"`
-	InputRegions     string        `form:"InputRegions"`
-	IncludeTravelers bool          `form:"IncludeTravelers"`
-	OnlyNonTravelers bool          `form:"OnlyNonTravelers"`
-	ExcludeRegions   string        `form:"ExcludeRegions"`
-	BucketName       string        `form:"BucketName"`
-	FilenameRoot     string        `form:"FilenameRoot"`
-	Period           time.Duration `form:"Period"`
-	FromDate         string        `form:"fromdate"`
-	FromTime         string        `form:"fromtime"`
-	ThruDate         string        `form:"thrudate"`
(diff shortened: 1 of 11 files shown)


internal/admin/exports/form.go
```diff
@@ -29,19 +29,20 @@ import (
...
+	BucketName         string        `form:"BucketName"`
+	FilenameRoot       string        `form:"FilenameRoot"`
+	Period             time.Duration `form:"Period"`
+	FromDate           string        `form:"fromdate"`
+	FromTime           string        `form:"fromtime"`
+	ThruDate           string        `form:"thrudate"`
+	ThruTime           string        `form:"thrutime"`
+	SigInfoIDs         []int64       `form:"siginfo"`
+	MaxRecordsOverride int           `form:"MaxRecordsOverride"`
 }
```
whaught, line 46: it's too late for this, but weird how we've mixed lowercase and CamelCase. I tend to prefer case-insensitive if it's coming from web.

internal/export/worker.go
```diff
@@ -267,7 +267,12 @@ func (s *Server) exportBatch(ctx context.Context, eb *model.ExportBatch, emitInd
 	logger := logging.FromContext(ctx)
 	db := s.env.Database()
 
-	logger.Infof("Processing export batch %d (root: %q, region: %s), max records per file %d", eb.BatchID, eb.FilenameRoot, eb.OutputRegion, s.config.MaxRecords)
+	maxRecords := eb.EffectiveMaxRecords(s.config.MaxRecords)
+	if maxRecords != s.config.MaxRecords {
+		logger.Debugw("max records is being overridden", "batchID", eb.BatchID)
```
sethvargo, line 272: Maybe log exp vs configured?

sethvargo (approved):
/lgtm

internal/admin/exports/form.go
sethvargo: We can change this pretty easily - it's an internal API

PR #1217, opened 2020-12-08
make data padding more better

mikehelmick:
## Proposed Changes

* Take randomness out of export padding and just clone every key at least once
* Makes some paths less complicated, but makes many paths more complicated (see before and after)

Using the same source data:

Before:
before

After
after

**Release Note**

```release-note
Export padding will clone every key at least once to better obscure TEK links.
```


whaught (approved):
"complify" is the word of the day

mikehelmick:
> "complify" is the word of the day

I try to hide Easter eggs in my branch names

PR #1218, opened 2020-12-08
don't stop retrying azure credential refresh on single failure

### LINKED ISSUE #1213: Kubernetes pods only get auth token on startup
### TL;DR
For the azure driver, Pods only get authentication token at startup.  This means that if the pod is not restarted for a long stretch of time, the token will expire after 60 days, causing the pod to fail requests to storage, etc.

**Expected behavior**
Get authentication token for every call, or at some more frequent interval.

**Observed behavior**
Auth token only refreshed only upon startup.

### Reproduction

Keep pod running for 60 days

mikehelmick:
Fixes #1213

## Proposed Changes

* Keep retrying azure blobstore credential refresh 
* If the token expires, cause the server to gracefully shutdown

**Release Note**

```release-note
Retry Azure credential refresh in the event of a refresh failure. Shutdown if token expires.
```

/assign @nmiodice

### DIFF
--- a/internal/storage/azure_blobstore.go
+++ b/internal/storage/azure_blobstore.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/Azure/azure-storage-blob-go/azblob"
 	"github.com/Azure/go-autorest/autorest/adal"
+	"github.com/google/exposure-notifications-server/pkg/signal"
 	"github.com/prometheus/common/log"
 )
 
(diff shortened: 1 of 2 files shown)


mikehelmick:
/hold
for @nmiodice to review

sethvargo (approved):
/lgtm

sethvargo (approved):
/lgtm

nmiodice:
This LGTM as well. Thanks @mikehelmick and @sethvargo

PR #1224, opened 2020-12-10
Use proper unit and avoid converting between units.

yegle:
It looks like the 'm' in 'ms' is a PREFIX_SYMBOL of 's' and we can use
's' as unit directly.

Reference: https://cloud.google.com/monitoring/mql/reference#units-of-measure


sethvargo:
CI is unhappy

mikehelmick:
/retest

that was my flaky test

mikehelmick (approved):

PR #1226, opened 2020-12-10
Remove the reference to the metrics middleware.

yegle:
Per the internal discussion, we don't actually use the log based metric
exporter. It would be simpler to just call stats.Record directly without
an extra layer of indirection.


sethvargo (approved):
/lgtm

mikehelmick (approved):

yegle:
Friendly ping?

PR #1227, opened 2020-12-10
Terraform: VPC connector depends on service networking connection

chaodaiG:
Adapted from verification server terrafrom config, terraform could be flaky without this


sethvargo (approved):
/lgtm

PR #1232, opened 2020-12-14
Add comment on how to use units/prefixes in threshold

yegle:


yegle:
This won't be needed as we are switching to use SLO based alert.

PR #1239, opened 2020-12-16
Adjust Stackdriver export timeout to 1 minute.

yegle:


yegle:
I think it would be easier to just change the default value (single PR instead of two).

yegle:
Oops closed the wrong PR.

sethvargo (approved):
/lgtm

PR #1240, opened 2020-12-16
Do not ignore annotation changes

yegle:
The annotation will be used to do the ingress control (currently EAP).

Part of https://github.com/google/exposure-notifications-verification-server/issues/895


sethvargo:
Hmm - there's a bunch of annotations that Cloud Run applies that were causing a bunch of temporal diffs. My preference would be to add this to the Terraform config (so new services get it), and then manually apply it to existing envs.

yegle:
I've double checked and the change here doesn't cause any diff when I run `terraform plan` on my personal project.

sethvargo:
IIRC, Cloud Runs adds annotations over time. If Cloud Run adds an annotation, the next time Terraform runs, it will try to remove that annotation.

PR #1241, opened 2020-12-16
Update notification channel setting in key server.

yegle:
This is to make it consistent between key server and verification server.


sethvargo (approved):
/lgtm

PR #1245, opened 2020-12-17
Terraform destroy first remove database from state

chaodaiG:
Prior to terraform0.14 this was fine, but now terraform destroy isn't happy that the database instance exists in terraform state but not physically exist in GCP


sethvargo (approved):
/lgtm

Issue #1250, opened 2020-12-20
Key server metrics for health authorities


mikehelmick:
### TL;DR

Track metrics at the key server for later retrieval by the verification server.

/assign

PR #1251, opened 2020-12-20
Metrics

mikehelmick:
First part of #1250

## Proposed Changes

* Create healthauthoritystats
* On publish, increment hourly counters for publish metrics

TODO: Add an API to retrieve these metrics by the health authority (signed w/ same key as verification certificates)

**Release Note**

```release-note
Statistics are captured on publish for health authorities, including: number of publishes by platform, number of TEKs uploaded, the days from onset to upload, and the age of the oldest TEK in the upload.
```


internal/publish/model/stats.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Package model is a model abstraction for health authority telemetry.
+package model
+
+import (
+	"time"
+)
+
+const (
+	MaxOldestTEK = 14
```
whaught, line 23: Should these max ages be config?

whaught (approved):
/lgtm
/hold feel free to unhold if you're satisfied

internal/publish/model/stats.go
mikehelmick: I don’t think so in this case. oldtest TEK age is bounded by what devices will report - although, I think we should probably make that 15

and the max onset days is what we now to be the largest epi relivent date (And that’s a bit of a stretch.

this is also the collection - display will be in verification server.

whaught: Yea verification takes 14 but 15 if tzOffset is negative (just to be careful about time zones)

internal/publish/model/exposure_model.go
```diff
@@ -558,7 +558,7 @@ func ReportTypeTransmissionRisk(reportType string, providedTR int) int {
 // The return params are the list of exposures, a list of warnings, and any
 // errors that occur.
 //
-func (t *Transformer) TransformPublish(ctx context.Context, inData *verifyapi.Publish, regions []string, claims *verification.VerifiedClaims, batchTime time.Time) ([]*Exposure, []string, error) {
+func (t *Transformer) TransformPublish(ctx context.Context, inData *verifyapi.Publish, regions []string, claims *verification.VerifiedClaims, batchTime time.Time) ([]*Exposure, *PublishInfo, []string, error) {
```
sethvargo, line 561: This is probably getting to the point where it makes sense to return a struct instead of 4 values.

internal/publish/model/exposure_model_test.go
```diff
@@ -445,6 +445,14 @@ func TestStillValidKey(t *testing.T) {
 			if !tf[0].CreatedAt.Equal(tc.createdAt) {
 				t.Errorf("wrong createdAt time, want: %v got: %v", tc.createdAt, tf[0].CreatedAt)
 			}
+
+			wantStats := &PublishInfo{
```
sethvargo, line 449: I've been trying to move us away from go-cmp for simple struct comparisons, since it's not exactly clear what you're trying to test.

internal/publish/model/stats.go
sethvargo, line 23: Can you document these two constants?

internal/publish/model/stats.go
```diff
@@ -0,0 +1,103 @@
...
+	MaxOnsetDays = 28
+
+	PlatformAndroid = "android"
+	PlatformIOS     = "ios"
+	PlatformUnknown = "unknown"
+)
+
+var (
+	platforms = map[string]int{
```
sethvargo, line 32: This feels like a potentially unnecessary optimization. I think we might be better off with a:

```go
func PlatformToInt(s string) int64 {
  switch s {
  case "android":
    return 1
  case "ios":
    return 2
  }

  return 0
}
```

I'm also very hesitant about having a "real" platform being mapped to the zero value.

internal/publish/model/stats.go
```diff
@@ -0,0 +1,103 @@
...
+		PlatformUnknown: 2,
+	}
+)
+
+// HealthAuthorityStats represents the raw metrics for an individual
+// health authority for a given hour.
+type HealthAuthorityStats struct {
+	HealthAuthorityID int64
+	Hour              time.Time
+	PublishCount      []int32
```
sethvargo, line 44: Using int32 here feels like a premature optimization that's going to come back and bite us. I made some premature optimizations with the modeler and I would prefer if we didn't make the same mistake again.

internal/publish/model/stats.go
```diff
@@ -0,0 +1,103 @@
...
+	CreatedAt    time.Time
+	Platform     string
+	NumTEKs      int32
+	Revision     bool
+	OldestDays   int
+	OnsetDaysAgo int
+	MissingOnset bool
+}
+
+// AddPublish increments the stats for a given hour. This should be called
```
sethvargo, line 77: Can we enforce "this should be called inside a RWM?" Require a tx or something?

sethvargo: It might also be helpful to explain what it does (or more specifically what it _doesn't_ do). It doesn't touch the database and is only mutating the struct.

internal/publish/platform.go
```diff
@@ -0,0 +1,47 @@
...
+package publish
+
+import (
+	"net/http"
+	"strings"
+
+	"github.com/google/exposure-notifications-server/internal/publish/model"
+)
+
+func platform(r *http.Request) string {
```
sethvargo, line 24: Nit: this should just take a string. There's no need to couple this to `http.Request`.

internal/publish/platform_test.go
```diff
@@ -0,0 +1,71 @@
...
+			wantPlatform: model.PlatformUnknown,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			r := httptest.NewRequest("POST", "/", strings.NewReader(""))
+			r.Header.Set("User-Agent", tc.userAgent)
+
+			if got := platform(r); got != tc.wantPlatform {
```
sethvargo, line 65: This is much easier to test if you decouple that function from `*http.Request`

internal/publish/publish_test.go
```diff
@@ -700,6 +739,26 @@ func TestPublishWithBypass(t *testing.T) {
 						if diff := cmp.Diff(tokenWant.RevisableKeys, revToken.RevisableKeys, revisableSorter, cmpopts.IgnoreUnexported(pb.RevisableKey{})); diff != "" {
 							t.Errorf("mismatch (-want, +got):\n%s", diff)
 						}
+
+						if tc.HealthAuthority != nil {
+							// There was a valid certificate present. there should be statistics.
+							stats, err := pubDB.ReadStats(ctx, tc.HealthAuthority.ID)
+							if err != nil {
+								t.Errorf("error reading back stats after publish: %v", err)
```
sethvargo, line 747: This should be Fatal, otherwise the test will NPE on `stats.`

migrations/000054_HealthAuthorityStats.up.sql
```diff
@@ -0,0 +1,38 @@
...
+-- limitations under the License.
+
+BEGIN;
+
+CREATE TABLE HealthAuthorityStats(
+    health_authority_id INT NOT NULL REFERENCES HealthAuthority(id) ON DELETE CASCADE,
+    -- UTC hour
+    hour TIMESTAMPTZ NOT NULL,
+    -- 3 elements, android/iphone/unknown
+    publish INT [],
```
sethvargo, line 22: Let's just go `bigint` on all of these.

sethvargo (changes requested):

internal/publish/database/exposure.go
```diff
@@ -643,7 +652,21 @@ func (db *PublishDB) InsertAndReviseExposures(ctx context.Context, req *InsertAn
...
 			}
+
+			if healthAuthorityID == 0 && exp.HealthAuthorityID != nil {
+				healthAuthorityID = *exp.HealthAuthorityID
+			}
 		}
+
+		// If requested, update the publish stats for the associated health authority.
+		if stats != nil && healthAuthorityID != 0 {
+			stats.NumTEKs = int32(resp.Inserted) + int32(resp.Revised)
```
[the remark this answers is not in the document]
sethvargo: Following up after reading more of the code: we should use `bigint` and `int64` everywhere instead.

internal/publish/database/exposure.go
```diff
@@ -643,7 +652,21 @@ func (db *PublishDB) InsertAndReviseExposures(ctx context.Context, req *InsertAn
 				}
 				resp.Revised++
 			}
+
+			if healthAuthorityID == 0 && exp.HealthAuthorityID != nil {
```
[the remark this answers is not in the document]
mikehelmick: more than one would be an error - changed to enforce that if.present, the are the same

internal/publish/database/exposure.go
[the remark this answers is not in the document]
mikehelmick: that are now uint32s - as unit64 was never going to make sense for number of keys inserted or revised in a single transaction.

for all practical purposes, these numbers will be order 100s at most.

internal/publish/database/stats.go
```diff
@@ -0,0 +1,124 @@
...
+
+// ReadStats will return all stats before the current hour, ordered in ascenting time.
+func (db *PublishDB) ReadStats(ctx context.Context, healthAuthorityID int64) ([]*model.HealthAuthorityStats, error) {
+	if healthAuthorityID <= 0 {
+		return nil, fmt.Errorf("missing healthAuthorityID")
+	}
+
+	// Slightly larger than normal upper bound.
+	results := make([]*model.HealthAuthorityStats, 0, 15*24)
+	now := time.Now()
```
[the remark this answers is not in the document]
mikehelmick: done + added comment

internal/publish/database/stats.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func scanOneHealthAuthorityStats(rows pgx.Rows, stats *model.HealthAuthorityStats) error {
+	return rows.Scan(
+		&stats.HealthAuthorityID, &stats.Hour, &stats.PublishCount, &stats.TEKCount,
+		&stats.RevisionCount, &stats.OldestTekDays, &stats.OnsetAgeDays, &stats.MissingOnset)
+}
+
+// UpdateStats records information for the health authority during a publish request.
+func (db *PublishDB) UpdateStats(ctx context.Context, tx pgx.Tx, hour time.Time, healthAuthorityID int64, info *model.PublishInfo) error {
```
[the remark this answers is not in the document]
mikehelmick: changed - 
accepting the Tx is important to keep this in the publish flow

internal/publish/model/stats.go
mikehelmick: ok - I changed this so that it is 16
0-15 and > 15

and 0-18 and > 28

internal/publish/model/stats.go
mikehelmick: I widened these to int64/ bigint

internal/publish/model/stats.go
mikehelmick: updated comment

internal/publish/model/stats.go
mikehelmick: made it so unknown is 0

mikehelmick:
all comments addressed

sethvargo (approved):
/lgtm

PR #1253, opened 2020-12-22
Refactor Terraform to allow easy override of Cloud Run service annotations

yegle:
Part of https://github.com/google/exposure-notifications-verification-server/issues/895

```release-note
Breaking: To continue using the Terraform module, the following input variable is needed to avoid introducing a diff:

service_annotations = {
    debugger        = { "autoscaling.knative.dev/maxScale" : "10" }
    export          = { "autoscaling.knative.dev/maxScale" : "10" }
    export-importer = { "autoscaling.knative.dev/maxScale" : "10" }
    exposure        = { "autoscaling.knative.dev/maxScale" : "500" }
    federationin    = { "autoscaling.knative.dev/maxScale" : "3" }
    federationout   = { "autoscaling.knative.dev/maxScale" : "5" }
    mirror          = { "autoscaling.knative.dev/maxScale" : "10" }
}
```

### DIFF
--- a/terraform/service_cleanup_export.tf
+++ b/terraform/service_cleanup_export.tf
@@ -101,11 +101,14 @@ resource "google_cloud_run_service" "cleanup-export" {
     }
 
     metadata {
-      annotations = {
+      annotations = merge({
         "autoscaling.knative.dev/maxScale" : "1",
         "run.googleapis.com/vpc-access-connector" : google_vpc_access_connector.connector.id
         "run.googleapis.com/vpc-access-egress" : "private-ranges-only"
-      }
+        }, local.enable_lb ? {
+        "run.googleapis.com/ingress" : "internal-and-cloud-load-balancing"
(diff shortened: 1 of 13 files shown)


mikehelmick:
Currently this will break e2e tests since they use the CR name by default. We will need to fix that.

sethvargo (commented):
This is probably fine as-is, so feel free to push back if you don't like my suggestions.

1. I think we should make this optional, even if the lb is enabled
2. I wonder if we should generalize this and create a variable like `additional_annotations` and merge with that, then we can apply this annotation and any future annotations at the module instantiation.

yegle:
@sethvargo I've repurposed this PR to be a refactor only (note the breaking change in the release note section). PTAL and let me know if this LGTY before I do the same thing to the verification server.

sethvargo (approved):
/lgtm

PR #1259, opened 2021-01-05
add stats retrieval API

### LINKED ISSUE #1250: Key server metrics for health authorities

mikehelmick:
Fixes #1250

## Proposed Changes

* adds API to retrieve PHA metrics - for use by verification server
* adds cleanup for stats older than 14/15 days ago (only deletes whole days at a time)

**Release Note**

```release-note
Adds a new API to retrieve publish related metrics for a PHA. Total # of publish requests by platform, total number of TEKs, distribution of oldest TEK in an upload, distribution of onset to upload time.
```


pkg/api/v1/metrics_types.go
```diff
@@ -0,0 +1,75 @@
...
+type MetricsResponse struct {
+	// Individual days. There may be gaps if a day does not have enough data.
+	Days []MetricsDay `json:"days,omitempty"`
+
+	ErrorMessage string `json:"error,omitempty"`
+	ErrorCode    string `json:"code,omitempty"`
+}
+
+// MetricsDay represents stats from an individual day.
+type MetricsDay struct {
```
mikehelmick, line 48: @sethvargo - pay special attention to this, because we'll need to be able to pull this into the verification sever and display something useful.

cmd/exposure/main.go
```diff
@@ -70,6 +70,9 @@ func realMain(ctx context.Context) error {
 	mux.Handle("/v1/publish", handler.Handle())
 	mux.Handle("/v1/publish/", http.NotFoundHandler())
 
+	// Handle metrics retrieval API
+	mux.Handle("/v1/metrics", handler.HandleMetrics())
```
sethvargo, line 74: Should this be /stats instead? I think this is the only time it's called "metrics" and that word tends to mean something else in the context of this system.

internal/cleanup/cleanup.go
```diff
@@ -91,6 +91,20 @@ func (h *exposureCleanupHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
 
 	stats.Record(ctx, cleanup.ExposuresDeleted.M(count))
 	logger.Infof("cleanup run complete, deleted %v records.", count)
+
+	// Clean up associated stats.
+	statsCount, err := h.database.DeleteStatsBefore(timeoutCtx, cutoff)
```
sethvargo, line 96: We may need to increase the allowed latency on this job now.

internal/publish/config.go
```diff
@@ -98,6 +98,9 @@ type Config struct {
 	// Normally "still valid" keys can be accepted, but are embargoed.
 	ReleaseSameDayKeys      bool `env:"DEBUG_RELEASE_SAME_DAY_KEYS"`
 	DebugLogBadCertificates bool `env:"DEBUG_LOG_BAD_CERTIFICATES"`
+
+	// Publish stats API config
+	StatsUploadMinimum int64 `env:"STATS_UPLOAD_MINIMUM, default=10"`
```
sethvargo, line 103: StatusUploadMinimum is the minimum number of statistics that must be present to return a response? It'd be good to document the behavior of what happens if the number is less than this.

internal/publish/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/google/exposure-notifications-server/pkg/logging"
+	"go.opencensus.io/trace"
+
+	"github.com/mikehelmick/go-chaff"
+)
+
+func (h *PublishHandler) HandleMetrics() http.Handler {
+	mResponder := maintenance.New(h.config)
+	return h.tracker.HandleTrack(chaff.HeaderDetector("X-Chaff"),
+		mResponder.Handle(
```
sethvargo, line 38: Can we move this line up to the previous? That will align with the closing at the bottom and also save us a full tab shift

internal/publish/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+				if err != nil {
+					message := fmt.Sprintf("error unmarshalling API call, code: %v: %v", code, err)
+					span.SetStatus(trace.Status{Code: trace.StatusCodeInternal, Message: message})
+					errorCode := verifyapi.ErrorBadRequest
+					if code == http.StatusInternalServerError {
+						errorCode = verifyapi.ErrorInternalError
+					}
+					w.Header().Set("Content-Type", "application/json")
+					w.WriteHeader(http.StatusBadRequest)
+					fmt.Fprintf(w, "{\"error\": \"%v\", \"code\": \"%v\"}", err.Error(), errorCode)
```
sethvargo, line 54: This scares me. I've been in verification server land for awhile, but I thought we had a JSON renderer in this project? As written, if the error contains a quote character, this is no longer valid JSON.

If you're opposed to marshalling as JSON, slightly safer would be:

```go
fmt.Fprintf(w, `{"error": %q, "code": %q}`, err.Error(), errCode)
```

internal/publish/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+					return
+				}
+
+				response, status := h.handleMetricsRequest(ctx, r.Header.Get("Authorization"), &request)
+
+				data, err := json.Marshal(response)
+				if err != nil {
+					w.Header().Set("Content-Type", "application/json")
+					w.WriteHeader(http.StatusInternalServerError)
+					fmt.Fprintf(w, "{\"error\": \"%v\"}", err.Error())
```
sethvargo, line 64: Same note about marshalling

internal/publish/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+					return
+				}
+
+				w.Header().Set("Content-Type", "application/json")
```
sethvargo, line 68: The content-type is always appjson - you could probably set that once higher up in the body

internal/publish/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+				w.WriteHeader(status)
+				fmt.Fprintf(w, "%s", data)
```
sethvargo, line 70:
```go
fmt.Fprint(w, data)
```

internal/publish/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+			})))
+}
+
+func (h *PublishHandler) handleMetricsRequest(ctx context.Context, bearerToken string, request *verifyapi.MetricsRequest) (*verifyapi.MetricsResponse, int) {
+	logger := logging.FromContext(ctx)
+	response := &verifyapi.MetricsResponse{}
+
+	if !strings.HasPrefix(bearerToken, "Bearer ") {
```
sethvargo, line 78: Could we pull the auth header parsing into a helper for easier testing? Can be an issue + follow up.

internal/publish/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+		response.ErrorMessage = "Authorization header is not in `Bearer <token>` format"
+		return response, http.StatusBadRequest
+	}
+	// Remove 'Bearer ' from the token.
+	bearerToken = bearerToken[7:]
```
sethvargo, line 83: Probably want to trimspace and sanity check it's not empty

internal/publish/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+		return response, http.StatusInternalServerError
+	}
+
+	// Nothing from the current hour can be shown.
+	onlyBefore := time.Now().UTC().Truncate(time.Hour)
+
+	// Combine days - this also filters things that are "too new" and days that don't meet the threshold.
+	metricsDays := model.ReduceStats(stats, onlyBefore, h.config.StatsUploadMinimum)
+	// Transform array of points to array of structs before return.
+	response.Days = make([]verifyapi.MetricsDay, len(metricsDays))
```
sethvargo, line 108: It'd be better to pass around `[]*verifyapi.MetricsDay` (pointers)

internal/publish/model/stats.go
```diff
@@ -57,6 +61,60 @@ type HealthAuthorityStats struct {
...
+		if !hour.Hour.Before(onlyBefore) {
+			continue
+		}
+
+		day := timeutils.UTCMidnight(hour.Hour)
+		if _, ok := days[day]; !ok {
+			// Initialize this day
+			days[day] = &verifyapi.MetricsDay{
+				Day:                   day,
+				OldestTEKDistribution: make([]int64, StatsMaxOldestTEK+1),
```
sethvargo, line 81: These +1 feel arbitrary?

internal/utils/utils.go
```diff
@@ -49,6 +49,34 @@ type SigningKey struct {
 	PublicKey string
 }
 
+type StatsJWTConfig struct {
```
sethvargo, line 52: Doc

internal/utils/utils.go
```diff
@@ -49,6 +49,34 @@ type SigningKey struct {
...
+	HealthAuthority    *vm.HealthAuthority
+	HealthAuthorityKey *vm.HealthAuthorityKey
+	Key                *ecdsa.PrivateKey
+	Audience           string
+	JWTWarp            time.Duration
+}
+
+func (c *StatsJWTConfig) IssueStatsJWT(t *testing.T) string {
```
sethvargo, line 60: Doc

sethvargo, line 60: Take testing.TB instead

internal/verification/authenticate_stats.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+// AuthenticateStatsToken parse the provided JWT and determines if it is an authorized stats request.
+func (v *Verifier) AuthenticateStatsToken(ctx context.Context, rawToken string) (int64, error) {
+	logger := logging.FromContext(ctx)
+
+	var healthAuthorityID int64
+	var claims *jwt.StandardClaims
+
+	token, err := jwt.ParseWithClaims(rawToken, &jwt.StandardClaims{}, func(token *jwt.Token) (interface{}, error) {
```
sethvargo, line 35: I think @whaught found an edge case using `ParseWithClaims` in the verification server, Do you remember what it was @whaught?

/hold

internal/verification/authenticate_stats.go
```diff
@@ -0,0 +1,105 @@
...
+		if method, ok := token.Method.(*jwt.SigningMethodECDSA); !ok || method.Name != jwt.SigningMethodES256.Name {
+			return nil, fmt.Errorf("unsupported signing method, must be %v", jwt.SigningMethodES256.Name)
+		}
+
+		kidHeader, ok := token.Header["kid"]
+		if !ok {
+			err := errors.New("missing 'kid' header in token")
```
sethvargo, line 42: These should all probably just return an error instead of logging. Move the logger higher

internal/verification/authenticate_stats.go
```diff
@@ -0,0 +1,105 @@
...
+		cacheVal, err := v.haCache.WriteThruLookup(claims.Issuer, lookup)
+		if err != nil {
+			return nil, err
+		}
+
+		if cacheVal == nil {
+			return nil, fmt.Errorf("issuer not found: %v", claims.Issuer)
+		}
+
+		healthAuthority := cacheVal.(*model.HealthAuthority)
```
sethvargo, line 81: Can you check this just so it doesn't panic?

internal/verification/authenticate_stats.go
```diff
@@ -0,0 +1,105 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/dgrijalva/jwt-go"
+	"github.com/google/exposure-notifications-server/internal/verification/database"
+	"github.com/google/exposure-notifications-server/internal/verification/model"
+	"github.com/google/exposure-notifications-server/pkg/logging"
+)
+
+// AuthenticateStatsToken parse the provided JWT and determines if it is an authorized stats request.
```
sethvargo, line 28: "It returns the health authority id"

internal/verification/config.go
```diff
@@ -20,4 +20,6 @@ import "time"
 // verification piece.
 type Config struct {
 	CacheDuration time.Duration `env:"VERIFICATION_CACHE_DURATION, default=5m"`
+
+	StatsAudience string `env:"STATS_AUDIENCE, default=keyserver"`
```
sethvargo, line 24: Doc

pkg/api/v1/metrics_types.go
```diff
@@ -0,0 +1,75 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1
+
+import "time"
+
+const (
+	ErrorUnauthorized = "unauthorized"
```
sethvargo, line 20: Doc

pkg/api/v1/metrics_types.go
```diff
@@ -0,0 +1,75 @@
...
+type MetricsRequest struct {
+	// currently no fields in the request.
+}
+
+// MetricsResponse returns all currently known metrics for the authenticated health authority.
+//
+// There may be gaps in the Days if a day has insufficient data.
+type MetricsResponse struct {
+	// Individual days. There may be gaps if a day does not have enough data.
+	Days []MetricsDay `json:"days,omitempty"`
```
sethvargo, line 41:
```suggestion
	Days []*MetricsDay `json:"days,omitempty"`
```

sethvargo: (otherwise omitempty is useless too)

pkg/api/v1/metrics_types.go
```diff
@@ -0,0 +1,75 @@
...
+
+package v1
+
+import "time"
+
+const (
+	ErrorUnauthorized = "unauthorized"
+)
+
+// MetricsRequest represents the request to retrieve publish metrics for a specific
```
sethvargo, line 23: I'd prefer if we called this "stats" to match the other nomenclature.

pkg/api/v1/metrics_types.go
```diff
@@ -0,0 +1,75 @@
...
+	// OldestTEKDistribution shows a distribution of the oldest tek in an upload.
+	// The count at index 0-15 represent the number of uploads there the oldest TEK is that value.
+	// Index 16 represents > 15.
+	OldestTEKDistribution []int64 `json:"oldestTEKDistribution"`
+	// OnsetToUpload shows a distribution of onset to upload, the index is in days.
+	// The count at index 0-29 represents the number of uploads with that symptom onset age.
+	// Index 30 represents > 29 days.
+	OnsetToUpload []int64 `json:"onsetToUpload"`
+
+	MissingOnset int64 `json:"missingOnset"`
```
sethvargo, line 64: What is this field?

pkg/api/v1/metrics_types.go
```diff
@@ -0,0 +1,75 @@
...
+	ErrorMessage string `json:"error,omitempty"`
+	ErrorCode    string `json:"code,omitempty"`
+}
+
+// MetricsDay represents stats from an individual day.
+type MetricsDay struct {
+	// Day will be set to midnight UTC of the day represented. An individual day
+	// isn't released until there is a minimum threshold for updates has been met.
+	Day          time.Time    `json:"day"`
+	PublishCount PublishCount `json:"publishCount"`
```
sethvargo, line 52: IIUC, the data we're trying to show here is:

- Total publish **requests**
- Publish **requests** by platform

- Can we make it clearer that this is the number of _requests_?
- Is that _successful_ publish count or just the total attempted count?
- Is there a reason to have this nested instead of flat?

sethvargo, line 52:
```suggestion
	PublishCount PublishCount `json:"publish_count"`
```

pkg/api/v1/metrics_types.go
```diff
@@ -0,0 +1,75 @@
...
+	TEKs         int64        `json:"teks"`
```
sethvargo, line 53: This is the total number of TEKs?

pkg/api/v1/metrics_types.go
```diff
@@ -0,0 +1,75 @@
...
+	Revisions    int64        `json:"revisions"`
```
sethvargo, line 54: Is this the number of revised TEKs or the number of requests that had any revisions?

pkg/api/v1/metrics_types.go
```diff
@@ -0,0 +1,75 @@
...
+	// OldestTEKDistribution shows a distribution of the oldest tek in an upload.
+	// The count at index 0-15 represent the number of uploads there the oldest TEK is that value.
+	// Index 16 represents > 15.
+	OldestTEKDistribution []int64 `json:"oldestTEKDistribution"`
```
sethvargo, line 58:
```suggestion
	TEKAgeDistribution []int64 `json:"tek_age_distribution"`
```

pkg/api/v1/metrics_types.go
```diff
@@ -0,0 +1,75 @@
...
+	// OnsetToUpload shows a distribution of onset to upload, the index is in days.
+	// The count at index 0-29 represents the number of uploads with that symptom onset age.
+	// Index 30 represents > 29 days.
+	OnsetToUpload []int64 `json:"onsetToUpload"`
```
sethvargo, line 62: I'm having trouble wrapping my head around this and the previous value. Are these relative to the `Day` field?

sethvargo (changes requested):
I think this looks good. I'm confused about some of the field's values. I think we could do a better job naming and describing some of these fields, specifically with respect to what _exactly_ is being counted by a particular entry.

cmd/exposure/main.go
mikehelmick: renamed - and all associated.

internal/cleanup/cleanup.go
mikehelmick: right now, the job timeout is 5 minutes - I'm not worried this will increase that.
current runtimes are about ~10s. We could increase the frequency of the exposure cleanup as well.

internal/publish/metrics.go
mikehelmick: we didn't have one - fixed here and others.

internal/publish/metrics.go
mikehelmick: it actually doesn't matter if it's empty - it will fail on the JWT parsing (There are test cases to cover that already)

internal/publish/metrics.go
mikehelmick: this is intentionally - the response struct isn't pointers (perhaps it should be?)

the return is a pointer to the struct that contains this slice.

internal/publish/model/stats.go
mikehelmick: it's not - the +1 is the > max bucket throughout

internal/verification/authenticate_stats.go
whaught: I do! ParseWithClaims automatically check claims.Valid(), but if you wish to validate yourself you can set Skip on the parser. That said I don't think we are manually checking claim validity here

whaught: See: https://github.com/google/exposure-notifications-verification-server/pull/1465/files

pkg/api/v1/metrics_types.go
mikehelmick: yes, expanded name.

pkg/api/v1/metrics_types.go
mikehelmick: added comment

pkg/api/v1/metrics_types.go
mikehelmick: they are absolute number of days for uploads occurring on the day field

internal/verification/authenticate_stats.go
mikehelmick: ok, there's no issue here then.

sethvargo (approved):
/lgtm

The descriptive field names and accompanying doc make it clearer, thanks!

PR #1261, opened 2021-01-06
Remove dupe ctx logger in mains

jklaw90:
## Proposed Changes

* Nothing important, just noticed some dupe contexts in the mains


whaught (approved):

whaught:
Thanks for helping us tidy up!

PR #1263, opened 2021-01-06
write publish stats in the background

mikehelmick:
## Proposed Changes

* be consistent with verification server stats - i.e. best effort.

**Release Note**

```release-note
Publish statistics are written in the background instead of inline with the publish request.
```

### DIFF
--- a/internal/publish/publish_test.go
+++ b/internal/publish/publish_test.go
@@ -736,6 +736,8 @@ func TestPublishWithBypass(t *testing.T) {
 						}
 
 						if tc.HealthAuthority != nil {
+							// statistics are written in the background.
+							time.Sleep(2 * time.Second)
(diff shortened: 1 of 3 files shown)


internal/publish/publish_test.go
```diff
@@ -736,6 +736,8 @@ func TestPublishWithBypass(t *testing.T) {
 						}
 
 						if tc.HealthAuthority != nil {
+							// statistics are written in the background.
+							time.Sleep(2 * time.Second)
```
sethvargo, line 740: Can we make this 5s just to avoid flakes? I'd rather have a slightly longer test than have to re-run it 4x.

sethvargo (approved):
/hold

whaught (approved):

whaught (approved):

PR #1271, opened 2021-01-07
readcommitted is right protection level, since it's a simple for update

mikehelmick:
## Proposed Changes

* choose correct isolation level

**Release Note**

```release-note
NONE
```


sethvargo (approved):
/lgtm

PR #1272, opened 2021-01-08
allow stats API to be disabled on a per HA basis. Improve test coverage

### LINKED ISSUE #1267: allow for stats API disabled on a per HA ID
/assign
/kind enhancement

mikehelmick:
Fixes #1267

## Proposed Changes

* Allow stats API to be disabled for a specific HA (enabled by default)
* Improve test coverage
* Fix make file, prow doesn't have awk apparently

**Release Note**

```release-note
Allow stats API to be enabled/disabled on a per HA basis.
```

### DIFF
--- a/internal/verification/authenticate_stats.go
+++ b/internal/verification/authenticate_stats.go
@@ -77,6 +77,11 @@ func (v *Verifier) AuthenticateStatsToken(ctx context.Context, rawToken string) 
 		if !ok {
 			return nil, fmt.Errorf("incorrect type in cache: %w", err)
 		}
+		// check that the API is enabled for this HA.
+		if !healthAuthority.EnableStatsAPI {
+			return nil, fmt.Errorf("API access forbidden")
+		}
+
 		// Look for the matching 'kid'
 		for _, key := range healthAuthority.Keys {
 			if key.Version == kid && key.IsValid() {
--- a/migrations/000055_AddDisableHAStats.down.sql
+++ b/migrations/000055_AddDisableHAStats.down.sql
@@ -1,0 +1,20 @@
+-- Copyright 2020 Google LLC
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--      http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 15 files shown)


internal/verification/authenticate_stats.go
```diff
@@ -77,6 +77,11 @@ func (v *Verifier) AuthenticateStatsToken(ctx context.Context, rawToken string)
 		if !ok {
 			return nil, fmt.Errorf("incorrect type in cache: %w", err)
 		}
+		// check that the API is enabled for this HA.
+		if !healthAuthority.EnableStatsAPI {
```
sethvargo, line 81: We'll need to make a note to come back to this during productionization. This is a user error and should be 4xx not 5xx.

migrations/000055_AddDisableHAStats.down.sql
```diff
@@ -0,0 +1,20 @@
+-- Copyright 2020 Google LLC
```
sethvargo, line 1: 2021

migrations/000055_AddDisableHAStats.up.sql
```diff
@@ -0,0 +1,23 @@
+-- Copyright 2020 Google LLC
```
sethvargo, line 1: 2021

tools/admin-console/templates/healthauthority.html
```diff
@@ -55,6 +55,17 @@
         </small>
       </div>
 
+      <div class="form-group">
+        <label for="enable-stats-api">Enable Stats API Access</label>
+        <select name="enable-stats-api" id="enable-stats-api" class="form-control custom-select">
```
sethvargo, line 60: This is probably better served as a checkbox, but meh.

sethvargo (commented):
/lgtm
/hold

internal/verification/authenticate_stats.go
mikehelmick: this results in http.StatusUnauthorized

migrations/000055_AddDisableHAStats.down.sql
mikehelmick: also updated template...

tools/admin-console/templates/healthauthority.html
mikehelmick: we use select a lot for this...

PR #1274, opened 2021-01-08
ensure imported keys are future dated

mikehelmick:
## Proposed Changes

* for imported keys - future date them if not expired. Works around potential bugs in upstream systems
* put created at date for published future dated keys into the NEXT interval. 

**Release Note**

```release-note
Imported keys will be future dated if they haven't expired yet.
Published keys will be moved forward one extra interval when they are adjusted. **There is no actual issue here with the default configuration because of the additional 2 hour embargo, but this is more technically correct.**
```


sethvargo (approved):
/lgtm

Issue #1273, opened 2021-01-08
Make admin console deployable


sethvargo:
### TL;DR
Run admin console as a Cloud Run service, only select users have invoker permissions.

PR #1275, opened 2021-01-08
Make admin-console deployable

sethvargo:
/assign @mikehelmick 

Part of https://github.com/google/exposure-notifications-server/issues/1273

**Release Note**

```release-note
Make admin-console deployable
```


sethvargo (commented):
The diff is large, but it's mostly copy-paste. I had to rename some things to prevent a circular dependency though.

builders/server.dockerfile
```diff
@@ -0,0 +1,31 @@
+# Copyright 2021 Google LLC
```
mikehelmick, line 1: why doesn't he existing service.dockerfile work?

mikehelmick (approved):
/hold
for the one question

builders/server.dockerfile
sethvargo: Have to copy the assets over

sethvargo: (templates)

cmd/admin-console/main.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/google/exposure-notifications-server/internal/buildinfo"
+	"github.com/google/exposure-notifications-server/internal/setup"
+	"github.com/google/exposure-notifications-server/pkg/logging"
+	"github.com/google/exposure-notifications-server/pkg/server"
+	"github.com/sethvargo/go-signalcontext"
+)
+
+func main() {
+	ctx, done := signalcontext.OnInterrupt()
+
```
odeke-em, line 32: If invoking done again is harmless, I'd encourage adding `defer done()` here so that regardless of what panics happen down below, it'll always be run, despite `done()` being explicitly invoked after `realMain(ctx)`

internal/admin/handle_authorized_apps.go
```diff
@@ -0,0 +1,215 @@
...
+		if err != nil {
+			ErrorPage(c, err.Error())
+			return
+		}
+
+		ctx := c.Request.Context()
+		m := TemplateMap{}
+
+		aadb := database.New(s.env.Database())
+		if form.Action == "save" {
```
odeke-em, line 46: Minor nit, it took reading through a bunch of code blocks to see that the only permitted actions were "save" and "delete". Perhaps we could restructure this into a switch so
```go
switch form.Action {
default:
     ErrorPage(c, fmt.Sprintf("Invalid request, app to edit not found, with action: %q", form.Action))

case "delete":
      ...

case "save":
      ...
}
```

internal/admin/handle_authorized_apps.go
```diff
@@ -0,0 +1,215 @@
...
+
+func (f *authorizedAppFormData) PriorKey() string {
+	if f.FormKey != "" {
+		bytes, err := base64.StdEncoding.DecodeString(f.FormKey)
+		if err != nil {
+			return ""
+		}
+		return string(bytes)
+	}
+	return ""
```
odeke-em, line 196: Minor nit on inversion order to reduce nesting in the more critical case
```go
if f.FormKey == "" {
     return ""
}

bs, err := base64...
...
return string(bytes)
```

internal/admin/handle_export_importers.go
```diff
@@ -0,0 +1,157 @@
...
+				ErrorPage(c, fmt.Sprintf("Failed to parse `id` param: %s", err))
+				return
+			}
+
+			model, err = db.GetConfig(ctx, id)
+			if err != nil {
+				ErrorPage(c, fmt.Sprintf("Failed to load export importer config: %s", err))
+				return
+			}
+		}
```
odeke-em, line 101: This code seems duplicated also in `(*Server).HandleExportImportersSave` perhaps this helper
```go
func modelFromContext(c *gin.Context, db *model.ExportImportDB) (*model.ExportImport, erorr) {
    idRaw := strings.TrimSpace(c.Param("id"))
    if idRaw == "" || idRaw == "0" {
         // Just return a blank model.
         return new(model.ExportImport), nil
    }
   
    id, err := strconv.ParseInt(idRaw, 10, 64)
    if err != nil {
         return nil, fmt.Errorf("Failed to parse `id` param: %w", err)
    }

    model, err := db.GetConfig(ctx, id)
    if err != nil {
        err = fmt.Errorf("Failed to load export importer config: %w", err)
    }
    return model, err
}
```

internal/admin/handle_export_importers.go
```diff
@@ -0,0 +1,157 @@
...
+		c.Region = val
+	}
+
+	if !from.IsZero() {
+		c.From = from
+	}
+
+	if !thru.IsZero() {
+		c.Thru = &thru
+	}
```
odeke-em, line 154: Please beware that if at any point a time.Time value was created by unmarshaling Seconds and Nanosecond respectively and it got the zero value, these checks will fail as per https://play.golang.org/p/ma-1o9J4ynh or inlined  below
```go
package main

import (
	"encoding/json"
	"time"
)

func main() {
	type recv struct {
		Sec int64 `json:"sec"`
		Ns  int64 `json:"ns"`
	}
	rc := new(recv)
	if err := json.Unmarshal([]byte(`{}`), rc); err != nil {
		panic(err)
	}
	t := time.Unix(rc.Sec, rc.Ns)
	println("isZero", t.IsZero(), "unix <= 0", t.UnixNano() <= 0)
}
```
which prints `isZero false unix <= 0 true`

For that I'd highly recommend using `(time.Time).UnixNano() > 0`

internal/admin/handle_exports.go
```diff
@@ -0,0 +1,200 @@
...
+			return
+		}
+
+		ctx := c.Request.Context()
+		m := TemplateMap{}
+
+		exportDB := database.New(s.env.Database())
+		exportConfig := &model.ExportConfig{}
+		idParam := c.Param("id")
+		if idParam != "0" {
```
odeke-em, line 47: If it is `""` this code will trip, but in prior checks you explicitly checked for `idParam != "" && idParam != "0"`. This is more impetus to use the refactored helper that I suggested above :-)

internal/admin/handle_exports.go
```diff
@@ -0,0 +1,200 @@
...
+
+// HandleExportsShow handles the show action for exports.
+func (s *Server) HandleExportsShow() func(c *gin.Context) {
+	return func(c *gin.Context) {
+		ctx := c.Request.Context()
+		m := TemplateMap{}
+
+		exportDB := database.New(s.env.Database())
+		exportConfig := &model.ExportConfig{}
+		if idParam := c.Param("id"); idParam == "0" {
```
odeke-em, line 100: Ditto about `idParam == ""` too?

internal/admin/handle_health_authority.go
```diff
@@ -0,0 +1,240 @@
...
+// HandleHealthAuthorityShow handles the show action for health authorities.
+func (s *Server) HandleHealthAuthorityShow() func(c *gin.Context) {
+	return func(c *gin.Context) {
+		ctx := c.Request.Context()
+		m := TemplateMap{}
+
+		healthAuthority := &model.HealthAuthority{
+			EnableStatsAPI: true, // default enabled.
+		}
+		if IDParam := c.Param("id"); IDParam == "0" {
```
odeke-em, line 86: Does `idParam == ""` matter here?

internal/admin/handle_health_authority.go
```diff
@@ -0,0 +1,240 @@
...
+	ha.EnableStatsAPI = f.EnableStatsAPI
+	ha.SetJWKS(f.JwksURI)
+}
+
+type keyhealthAuthorityFormData struct {
+	Version  string `form:"version"`
+	PEMBlock string `form:"public-key-pem"`
+	FromDate string `form:"from-date"`
+	FromTime string `form:"from-time"`
+	ThruDate string `form:"Thru-date"`
```
odeke-em, line 208: Did you intend to capitalize it to "Thru-date", or did you mean "thru-date"?

internal/admin/handle_health_authority.go
```diff
@@ -0,0 +1,240 @@
...
+	hak.From = fTime
+	hak.Thru = tTime
+	hak.PublicKeyPEM = strings.TrimSpace(f.PEMBlock)
+
+	_, err = hak.PublicKey()
+	if err != nil {
+		return err
+	}
+
+	return nil
```
odeke-em, line 239: Please simplify this check to
```go
return err
```

odeke-em (commented):
Thank you for this change @sethvargo, and happy New Year to you and to the team! Just some comments, please take a look.

internal/admin/handle_authorized_apps.go
sethvargo: This PR is just moving existing code around. To reduce the size of a diff and to reduce the chance of introducing a bug, I'm only moving code, no refactors or other changes in this PR.

internal/admin/handle_authorized_apps.go
sethvargo: This PR is just moving existing code around. To reduce the size of a diff and to reduce the chance of introducing a bug, I'm only moving code, no refactors or other changes in this PR.

internal/admin/handle_export_importers.go
sethvargo: This PR is just moving existing code around. To reduce the size of a diff and to reduce the chance of introducing a bug, I'm only moving code, no refactors or other changes in this PR.

internal/admin/handle_export_importers.go
sethvargo: This PR is just moving existing code around. To reduce the size of a diff and to reduce the chance of introducing a bug, I'm only moving code, no refactors or other changes in this PR.

internal/admin/handle_exports.go
sethvargo: It's required to be a value by the router to reach this function. The router enforces a digit in the path.

This PR is just moving existing code around. To reduce the size of a diff and to reduce the chance of introducing a bug, I'm only moving code, no refactors or other changes in this PR.

internal/admin/handle_exports.go
sethvargo: It's required to be a value by the router to reach this function. The router enforces a digit in the path.

This PR is just moving existing code around. To reduce the size of a diff and to reduce the chance of introducing a bug, I'm only moving code, no refactors or other changes in this PR.

internal/admin/handle_health_authority.go
sethvargo: It's required to be a value by the router to reach this function. The router enforces a digit in the path.

This PR is just moving existing code around. To reduce the size of a diff and to reduce the chance of introducing a bug, I'm only moving code, no refactors or other changes in this PR.

internal/admin/handle_health_authority.go
sethvargo: This PR is just moving existing code around. To reduce the size of a diff and to reduce the chance of introducing a bug, I'm only moving code, no refactors or other changes in this PR.

internal/admin/handle_health_authority.go
sethvargo: This is most likely a bug, but again, this PR is just moving existing code around. To reduce the size of a diff and to reduce the chance of introducing a bug, I'm only moving code, no refactors or other changes in this PR.

sethvargo:
Hi @odeke-em 

Thank you for the review. The items you identified are important and we'll be sure to address them in a follow-up PR. However, all these bugs already exist (this is a copy-paste and package rename of existing code), and I'd prefer to keep this PR isolated to _just_ the move. I can create another PR after this is merged fixing up the items you mentioned. I just don't want to couple a rename with changes in case we need to do a rollback.

PR #1277, opened 2021-01-10
publish model to 100% coverage - add tests / remove dead code

mikehelmick:
## Proposed Changes

* 100% test coverage on publish/model package

**Release Note**

```release-note
increased test coverage on internal/publish/model
```

### DIFF
--- a/internal/publish/model/exposure_model.go
+++ b/internal/publish/model/exposure_model.go
@@ -39,7 +39,7 @@ var (
 	ErrorExposureKeyMismatch = fmt.Errorf("attempted to revise a key with a different key")
 	// ErrorNonLocalProvenance - key revision attempted on federated key, which is not allowed
 	ErrorNonLocalProvenance = fmt.Errorf("key not origionally uploaded to this server, cannot revise")
-	// ErrorNotSameFederationSource - if a key arrived by federation, it can onlybe be revised by the same query (same source)
+	// ErrorNotSameFederationSource - if a key arrived by federation, it can only be be revised by the same query (same source)
 	ErrorNotSameFederationSource = fmt.Errorf("key cannot be revised by a different federation query")
 	// ErrorKeyAlreadyRevised - attempt to revise a key that has already been revised.
 	ErrorKeyAlreadyRevised = fmt.Errorf("key has already been revised and cannot be revised again")
@@ -93,6 +93,9 @@ type Exposure struct {
 	base64Key string
 }
 
+// ExportImportConfig represents the configuration for
+// processing of export files from other systems and how they are imported
+// into this server.
 type ExportImportConfig struct {
 	DefaultReportType         string
 	BackfillSymptomOnset      bool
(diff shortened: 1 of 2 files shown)


internal/publish/model/exposure_model_test.go
```diff
@@ -1360,11 +1420,100 @@ func TestDaysFromSymptomOnset(t *testing.T) {
...
+			incoming: &Exposure{
+				ExposureKey:       []byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15},
+				ReportType:        "confirmed",
+				FederationQueryID: "foo",
+			},
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
```
odeke-em, line 1490: Minor nit: given that you wanted to parallel the main test above, in here you could do this:
```go
tc := tc
t.Run(tc.name, func(t *testing.T) {
    t.Parallel()
```

internal/publish/model/exposure_model_test.go
```diff
@@ -1360,11 +1420,100 @@ func TestDaysFromSymptomOnset(t *testing.T) {
...
+			incoming[0] = tc.incoming
+
+			_, err := ReviseKeys(ctx, existing, incoming)
+			if err == nil && tc.err != nil {
+				t.Errorf("missing expected error: %v", tc.err)
+			} else if err != nil && tc.err == nil {
+				t.Errorf("unexpected error: %v", err)
+			} else if !errors.Is(err, tc.err) {
+				t.Errorf("wrong error, want: %v got: %v", tc.err, err)
+			}
```
odeke-em, line 1505: Might be clearer to avoid the 3 sets of checks at the same level and split it instead into `tc.err != nil` and `tc.err == nil` respectively, so perhaps to do this:
```go
if tc.err != nil {
     if !errors.Is(err, tc.err) || err == nil {
          t.Fatalf("wrong error, got: %v want: %v", err, tc.err)
     }
} else if err != nil {
     t.Fatalf("unexpectedly returned an error: %v", err)
}
```

internal/publish/model/exposure_model.go
```diff
@@ -390,15 +395,8 @@ func DaysBetweenIntervals(a int32, b int32) int32 {
 	a = IntervalNumber(timeutils.UTCMidnight(TimeForIntervalNumber(a)))
 	b = IntervalNumber(timeutils.UTCMidnight(TimeForIntervalNumber(b)))
 	distance := b - a
+	// This will always divide evenly since a and b are aligned on UTC midnight boundaries.
```
odeke-em, line 398: Given this new condition, I'd just add here an assertion and panic if it fails.

odeke-em (commented):
Happy New Year @mikehelmick, and thank you for this change. Just some comments, please take a look.

internal/publish/model/exposure_model_test.go
```diff
@@ -1290,6 +1291,65 @@ func TestTransformOverlapping(t *testing.T) {
...
+	if !e.HasBeenRevised() {
+		t.Error("reports not revised after revisedAt set")
+	}
+
+	if revTime != *e.RevisedAt {
+		t.Errorf("revisedAt mismatch, want: %v got: %v", revTime, *e.RevisedAt)
+	}
+	if revType != *e.RevisedReportType {
+		t.Errorf("revised report type mismatch, want: %v got: %v", revTime, *e.RevisedAt)
+	}
```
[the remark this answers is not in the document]
sethvargo: All times and database configurations are UTC for this reason.

sethvargo (approved):
/lgtm

PR #1278, opened 2021-01-11
Fix typo: ImportResposne -> ImportResponse

ithinkihaveacat:
Fix spelling. The type `ImportResponse` is exported, but doesn't seem to be used.


sethvargo (approved):
/lgtm

PR #1288, opened 2021-01-12
Create a struct to provide build info

yegle:
The intention here is to be able to have an interface that can be passed to WithObservability(), so that WithObservability can be shared between key server and verification server.


sethvargo:
Hey @yegle what's the goal in sharing with the verification server? The verification server has it's own buildinfo package.

sethvargo:
Also, the values are set from -X flags which require the full package path. It doesn't make sense for this to exist external to the project.

yegle:
> Hey @yegle what's the goal in sharing with the verification server? The verification server has it's own buildinfo package.

The end goal is to share `WithObservability` between the two projects (https://github.com/google/exposure-notifications-verification-server/blob/5b8bc5da0de6cc4c99e4b27b90cff9b0e1d1206a/pkg/controller/middleware/observability.go#L26).

Note I also have a PR pending to use gorilla mux in key server code so the middleware can be shared.

sethvargo:
Hmm, I don't think this will actually work though. The buildinfo is [set via LDFLAGS](https://github.com/google/exposure-notifications-server/blob/7d5556e685cd996fb3080dfee8e591e6e1aebb3b/builders/build.yaml#L73) during buildtime. At the very least, the package import path needs to be updated there.

However, if the `WithObservabilty` helper is going to be shared, then the verification server would have to set the build tags for the key server repo, which feels wrong.

yegle:
> Hmm, I don't think this will actually work though. The buildinfo is [set via LDFLAGS](https://github.com/google/exposure-notifications-server/blob/7d5556e685cd996fb3080dfee8e591e6e1aebb3b/builders/build.yaml#L73) during buildtime. At the very least, the package import path needs to be updated there.
> 
> However, if the `WithObservabilty` helper is going to be shared, then the verification server would have to set the build tags for the key server repo, which feels wrong.

Ah I see. Updated build.yaml. PTAL and see if the change there makes the intention clear.

builders/build.yaml
```diff
@@ -70,7 +71,8 @@ steps:
   - 'go'
   - 'build'
   - '-trimpath'
-  - '-ldflags=-s -w -X=${_REPO}/internal/buildinfo.BuildID=${BUILD_ID} -X=${_REPO}/internal/buildinfo.BuildTag=${_TAG} -extldflags=-static'
+  - &add-buildinfo
```
sethvargo, line 74: AFAIK, Cloud Build does not support the full YAML spec: b/152180846. That's largely why there's so much copy-paste in these files.

sethvargo (commented):
Right, but if your plan is to share `WithObservability` and `WithObservability` is going to depend on the key server's `buildinfo` package. That means the _verification_ server is going to be setting LDFLAGS on a _key_ server package.

A better approach might be to make `buildinfo` an interface you can pass into `WithObservability`.

internal/buildinfo/buildinfo.go
```diff
@@ -17,9 +17,30 @@
 package buildinfo
 
 var (
-	// BuildID is the unique build identifier.
-	BuildID string = "unknown"
+	// id is the unique build identifier.
+	id string = "unknown"
```
sethvargo, line 21: I'm 85% certain that variables that are set from LDFLAGS have to be public. I think you can leave this as-is (and revert the build.yaml changes) and still have the same desired effect.

internal/buildinfo/buildinfo.go
```diff
@@ -17,9 +17,30 @@
...
 )
+
+// Build is an interface to get the build information.
+type Build interface {
+	ID() string
+	Tag() string
+}
+
+// KeyServer provides the build information about the key server.
+type KeyServer struct{}
```
sethvargo, line 28: Make this private

internal/buildinfo/buildinfo.go
```diff
@@ -17,9 +17,30 @@
...
+func (KeyServer) ID() string {
+	return id
+}
+
+// Tag returns the build tag.
+func (KeyServer) Tag() string {
+	return tag
+}
+
+var _ Build = KeyServer{}
```
sethvargo: Nit: doc
Nit: this is typically named something like "DefaultBuildInfo"

builders/build.yaml
yegle: TIL :-(

internal/buildinfo/buildinfo.go
yegle: Was this comment on the KeyServer struct, or a previous interface (that I've deleted)?

sethvargo: This one - I don't think this should be public. Also see my comment about LDFLAGS and private variables. I don't think you can set `tag` or `id` via an LDFLAG because they are private.

yegle: Are you saying we should define the struct as private, but have a public variable to use outside of this package?

I did test with setting the private variable using `ldflags` and it does seem to work.

mikehelmick:
My suggestion would be to revert most of this and just add the ability to create a struct from the buildinfo package w/ old flags.

yegle:
OK, the PR is simplified to just add a new public variable.

mikehelmick (approved):

PR #1296, opened 2021-01-15
Use gorilla mux.

yegle:
The intention is to be able to share middlewares between key server and
verification server.

NOTE: some of the changed lines are not strictly required (e.g. the
Prometheus /metrics handler can use the Go http mux perfectly fine), but
are changed to be consistent.

### DIFF
--- a/cmd/cleanup-export/main.go
+++ b/cmd/cleanup-export/main.go
@@ -18,7 +18,6 @@ package main
 import (
 	"context"
 	"fmt"
-	"net/http"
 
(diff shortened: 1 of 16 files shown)


mikehelmick:
see test failures

Also - any objection to holding this until after v0.20 is cut? I would like it to sit in our dev environment longer

yegle:
> see test failures
> 
> Also - any objection to holding this until after v0.20 is cut? I would like it to sit in our dev environment longer

SG

mikehelmick (commented):
/lgtm
/hold 
for after v0.20 is cut

PR #1297, opened 2021-01-19
don't save newline when importing public keys via JWKS

mikehelmick:
## Proposed Changes

* Don't save newlines at the end of public keys when importing

**Release Note**

```release-note
Don't save newlines at the end of public keys when importing via JWKS
```

### DIFF
--- a/pkg/jwks/jwks.go
+++ b/pkg/jwks/jwks.go
@@ -30,6 +30,7 @@ import (
 	"time"
 
 	"github.com/google/exposure-notifications-server/internal/database"
+	"github.com/google/exposure-notifications-server/internal/project"
 	"github.com/google/exposure-notifications-server/internal/setup"
 	hadb "github.com/google/exposure-notifications-server/internal/verification/database"
 	"github.com/google/exposure-notifications-server/internal/verification/model"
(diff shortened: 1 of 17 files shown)


pkg/jwks/jwks.go
```diff
@@ -136,7 +136,7 @@ func parseKeys(data []byte) ([]string, map[string]string, error) {
 		if err != nil {
 			return nil, nil, fmt.Errorf("marshal error: %w", err)
 		}
-		keys[i] = string(pem.EncodeToMemory(&pem.Block{Type: "PUBLIC KEY", Bytes: encoded}))
+		keys[i] = strings.TrimSpace(string(pem.EncodeToMemory(&pem.Block{Type: "PUBLIC KEY", Bytes: encoded})))
```
sethvargo, line 139: Should we copy over the project-level helper from the verification server instead? I'm fine with a copy-paste implementation.

sethvargo (approved):
/lgtm

/hold
for potentially more defense.

pkg/jwks/jwks.go
mikehelmick: added... throughout the codebase

sethvargo (commented):
/lgtm
/unhold

PR #1298, opened 2021-01-20
Manually create metric descriptor at view registration time.

yegle:
Some of our metrics are recorded rarely. Prior to this change, we rely
on OpenCensus stackdriver exporter to create the metric descriptor for
us, which means some metric descriptor will not be created long after
the code change, blocking the creation of monitoring and alerting.

Note the CreateMetricDescriptor call is idempotent and safe to call
repeatedly.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -80,3 +80,8 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	honnef.co/go/tools v0.1.0
 )
+
+// FIXME: remove once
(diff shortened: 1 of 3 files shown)


mikehelmick:
see static analysis error - check for correct version of the import

go.mod
```diff
@@ -80,3 +80,8 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	honnef.co/go/tools v0.1.0
 )
+
+// FIXME: remove once
```
sethvargo, line 84: Nit: please use `TODO(ych)` or similar annotation.

PR #1301, opened 2021-01-20
Build all binaries at once (faster build)

sethvargo:
**Release Note**

```release-note
Switch to a faster build process
```

/assign @whaught


builders/promote.yaml
```diff
@@ -244,17 +244,17 @@ steps:
...
+- id: 'promote-jwks-updater'
   name: 'gcr.io/google.com/cloudsdktool/cloud-sdk:307.0.0-alpine'
   args:
   - 'bash'
   - '-eEuo'
   - 'pipefail'
   - '-c'
   - |-
-    gcloud run services update-traffic "jwks" \
+    gcloud run services update-traffic "jwks-updater" \
```
mikehelmick, line 257: don't we need to update the terraform as well?

sethvargo: Yea, I decided to just rename the binary to match the deploy instead of editing all these files.

mikehelmick (approved):

PR #1307, opened 2021-01-26
CSV marshaller for stats

whaught:
Issue https://github.com/google/exposure-notifications-verification-server/issues/1512

## Proposed Changes

* CSV marshaller for statistics

**Release Note**

```release-note
CSV marshaller for statistics
```

### DIFF
--- a/pkg/api/v1/stats_types.go
+++ b/pkg/api/v1/stats_types.go
@@ -14,7 +14,14 @@
 
 package v1
 
-import "time"
+import (
+	"bytes"
+	"encoding/csv"
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+)
 
 const (
 	// ErrorUnauthorized is returned if the provided bearer token is invalid.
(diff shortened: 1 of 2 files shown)


ryanwolcott321 (approved):

ryanwolcott321:
Approve

pkg/api/v1/stats_types.go
```diff
@@ -84,3 +94,56 @@ type PublishRequests struct {
...
+
+	w.Flush()
+	if err := w.Error(); err != nil {
+		return nil, fmt.Errorf("failed to create CSV: %w", err)
+	}
+
+	return b.Bytes(), nil
+}
+
+func join(arr []int64, sep string) string {
```
sethvargo, line 140: Can be a follow-up, but it'd be good to move this into icsv as a public function in case we need to use it elsewhere.

sethvargo: (And probably make "|" the separator as a constant and not a param)

sethvargo (approved):

pkg/api/v1/stats_types.go
sethvargo: @whaught ^

PR #1315, opened 2021-01-30
fix carriage returns in public keys

mikehelmick:
## Proposed Changes

* don't save carriage returns  `\r` in public key PEM format
* strip `\r` on comparison in JWKS service

**Release Note**

```release-note
JWKS Service - fix issue where manually added keys containing `\r` characters could prevent upgrading a health authority to use JWKS discovery.
```


sethvargo (approved):
/lgtm

Issue #1316, opened 2021-01-30
export-import scheduler


mikehelmick:
### TL;DR

found `ERROR: could not serialize access due to read/write dependencies among transactions (SQLSTATE 40001)`
in the logs

call site `exportimport/scheduler.go:51`

/assign

PR #1317, opened 2021-01-30
check for serialization error on lock and unlock

mikehelmick:
Towards #1316

## Proposed Changes

Provide primitives to fixed observed occurrences of postgres serialization contention.

* Add `LockRetry` that checks for pg serialization error and uses exponential backoff to retry lock acquire
* Unlock function uses 3 constant reties on a tight schedule, only reties on serialization errors

If this looks good - I will asses each db.Lock call site to determine if we should retry lock acquisition.

From the docs: https://www.postgresql.org/docs/9.5/transaction-iso.html

The Serializable isolation level provides the strictest transaction isolation. This level emulates serial transaction execution for all committed transactions; as if transactions had been executed one after another, serially, rather than concurrently. However, like the Repeatable Read level, **applications using this level must be prepared to retry transactions due to serialization failures**.

**Release Note**

```release-note
DB backed locks will retry unlock 3 times in quick succession if unable to serialize the transaction.
```

### DIFF
--- a/internal/database/lock.go
+++ b/internal/database/lock.go
@@ -22,7 +22,9 @@ import (
 	"time"
 
 	"github.com/google/exposure-notifications-server/pkg/logging"
+	"github.com/jackc/pgconn"
 	pgx "github.com/jackc/pgx/v4"
+	"github.com/sethvargo/go-retry"
 )
 
 var (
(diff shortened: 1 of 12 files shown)


internal/database/lock.go
```diff
@@ -89,6 +91,36 @@ func makeMultiUnlockFn(ctx context.Context, db *DB, lockIDs []string, expires ti
 	}
 }
 
+func serializationError(err error) bool {
```
sethvargo, line 94: Nit: this should be `isSerializationError`

internal/database/lock.go
```diff
@@ -89,6 +91,36 @@ func makeMultiUnlockFn(ctx context.Context, db *DB, lockIDs []string, expires ti
...
+func (db *DB) LockRetry(ctx context.Context, lockID string, ttl time.Duration, tryFor time.Duration) (UnlockFn, error) {
+	// Ignored error - it checks for < 0
+	b, _ := retry.NewExponential(time.Millisecond)
+	b = retry.WithMaxDuration(tryFor, b)
+
+	var unlock UnlockFn
+	err := retry.Do(ctx, b, func(ctx context.Context) error {
+		var err error
+		unlock, err = db.Lock(ctx, lockID, ttl)
+		if err != nil && serializationError(err) {
```
sethvargo, line 113: If you change isSerializationError to return false when err is nil, you can simplify this and future calls

internal/database/lock.go
```diff
@@ -89,6 +91,36 @@ func makeMultiUnlockFn(ctx context.Context, db *DB, lockIDs []string, expires ti
...
+	// See https://www.postgresql.org/docs/current/errcodes-appendix.html
+	if pgErr, ok := err.(*pgconn.PgError); !ok || pgErr.Code == "40001" {
+		return true
+	}
+	return false
+}
+
+// LockRetry will attempt to acquire a lock using exponential backoff, up to the tryFor time.
+// Delegates to db.Lock().
+func (db *DB) LockRetry(ctx context.Context, lockID string, ttl time.Duration, tryFor time.Duration) (UnlockFn, error) {
```
sethvargo, line 104: I'm slightly confused. You wrote this nice function `LockRetry`but it doesn't seem to be used anywhere?

mikehelmick: will do in follow up - wanted to make sure this was agreeable structure for retries.

mikehelmick:
/hold
will fix some call sites - we're seeing this more in dev

internal/database/lock.go
sethvargo: I agree

mikehelmick:
/close
in favor of https://github.com/google/exposure-notifications-server/pull/1320

PR #1322, opened 2021-01-30
improve locking handling for background jobs

mikehelmick:
## Proposed Changes

* add job lock to JWKS
* return success for other jobs if the reason for lock fail was the lock is already held
* goal - reduce alert noise 

**Release Note**

```release-note
Improve locking for background jobs: export-import, jwks, key rotation
```

### DIFF
--- a/internal/jwks/handle_update_all.go
+++ b/internal/jwks/handle_update_all.go
@@ -53,6 +53,7 @@ func (s *Server) handleUpdateAll() http.Handler {
 			return
 		}
 
+		mJWKSSuccess.M(1)
 		w.WriteHeader(http.StatusOK)
 		fmt.Fprint(w, http.StatusText(http.StatusOK))
 	})
--- a/internal/jwks/handle_update_all.go
+++ b/internal/jwks/handle_update_all.go
@@ -15,17 +15,37 @@
 package jwks
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
+	"time"
 
+	"github.com/google/exposure-notifications-server/internal/database"
 	"github.com/google/exposure-notifications-server/pkg/logging"
 )
+
+const jwksLockID = "jwks-import"
 
 func (s *Server) handleUpdateAll() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
-
 		logger := logging.FromContext(ctx).Named("handleUpdateAll")
+
+		unlock, err := s.manager.db.Lock(ctx, jwksLockID, time.Minute)
+		if err != nil {
+			logger.Warnw("unable to obtain lock", "lock", jwksLockID, "error", err)
+			if errors.Is(err, database.ErrAlreadyLocked) {
+				w.WriteHeader(http.StatusOK)
+				return
+			}
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		defer func() {
+			if err := unlock(); err != nil {
+				logger.Errorw("failed to unlock", "lock", jwksLockID, "error", err)
+			}
+		}()
 
 		if err := s.manager.UpdateAll(ctx); err != nil {
 			logger.Errorw("failed to update all", "error", err)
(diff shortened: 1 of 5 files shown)


sethvargo (commented):
This seems good. I'm worried about a case where we fail to acquire the lock for many, many tries though. Could we have some upper bound or quiescence to alert if the lock is never acquired after some time? 

Specifically, I'm worried about a bug or crash that fails to abort.

sethvargo (approved):
/hold

mikehelmick:
added open census metrics to indicate job success

internal/jwks/handle_update_all.go
```diff
@@ -15,24 +15,45 @@
...
+
 func (s *Server) handleUpdateAll() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
-
 		logger := logging.FromContext(ctx).Named("handleUpdateAll")
 
+		unlock, err := s.manager.db.Lock(ctx, jwksLockID, time.Minute)
+		if err != nil {
+			logger.Warnw("unable to obtain lock", "lock", jwksLockID, "error", err)
```
sethvargo, line 36: Same about moving this after AlreadyLocked

internal/jwks/handle_update_all.go
```diff
@@ -15,24 +15,45 @@
...
+			}
+		}()
+
 		if err := s.manager.UpdateAll(ctx); err != nil {
 			logger.Errorw("failed to update all", "error", err)
 			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
 			return
 		}
 
+		mJWKSSuccess.M(1)
```
sethvargo, line 56: I think you need to call `stats.Record`?

internal/jwks/metrics.go
```diff
@@ -0,0 +1,40 @@
...
+		"Number of successful completions of jwks handler", stats.UnitDimensionless)
+)
+
+func init() {
+	observability.CollectViews([]*view.View{
+		{
+			Name:        metrics.MetricRoot + "jwks_success",
+			Description: "Number of successful completions of jwks handler",
+			Measure:     mJWKSSuccess,
+			Aggregation: view.Sum(),
```
sethvargo, line 37: I think it's view.Count

internal/keyrotation/rotate.go
```diff
@@ -67,6 +65,7 @@ func (s *Server) handleRotateKeys() http.Handler {
 			return
 		}
 
+		mRotationSuccess.M(1)
```
sethvargo, line 68: stats.Record

internal/keyrotation/metrics.go
```diff
@@ -45,5 +46,11 @@ func init() {
 			Measure:     mRevisionKeysDeleted,
 			Aggregation: view.Sum(),
 		},
+		{
+			Name:        metrics.MetricRoot + "rotation_success",
+			Description: "Number of successful completions of rotate handler",
+			Measure:     mRotationSuccess,
+			Aggregation: view.Sum(),
```
sethvargo, line 53: I think you want count? 
 
> Count only counts the number of measurement points recorded. Distribution provides statistical summary of the aggregated data by counting how many recorded measurements fall into each bucket. Sum adds up the measurement values. LastValue just keeps track of the most recently recorded measurement value. All aggregations are cumulative.

internal/jwks/handle_update_all.go
```diff
@@ -15,24 +15,46 @@
...
+			}
+		}()
+
 		if err := s.manager.UpdateAll(ctx); err != nil {
 			logger.Errorw("failed to update all", "error", err)
 			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
 			return
 		}
 
+		stats.Record(ctx, mJWKSSuccess.M(1))
```
sethvargo, line 57: Handle this error?

internal/keyrotation/rotate.go
```diff
@@ -67,6 +65,7 @@ func (s *Server) handleRotateKeys() http.Handler {
 			return
 		}
 
+		stats.Record(ctx, mRotationSuccess.M(1))
```
sethvargo, line 68: Handle this error?

sethvargo (approved):
/lgtm

Two Qs about error handling that can be a follow-up

PR #1323, opened 2021-01-31
Update locking procedure to be strongly consistent

sethvargo:
This updates the existing AcquireLock and ReleaseLock procedures to be strongly consistent by introducing a slight behavior change: once a lock exists, it exists forever. Releasing a lock sets the timestamp to 01-01-1970 instead of the previous behavior which deleted the row. Additionally, instead of relying on a parent transaction for isolation, the statements can operate independently from each other. This should dramatically reduce the number of serialization deadlocks we see.

I'm like 99% certain we could drop the isolation level back to READ COMMITTED after this, but I didn't want to be too invasive here.

**Release Note**

```release-note
Update locking procedure to be strongly consistent
```

### DIFF
--- a/migrations/000056_UpdateLocks.down.sql
+++ b/migrations/000056_UpdateLocks.down.sql
@@ -1,0 +1,52 @@
+-- Copyright 2021 Google LLC
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--      http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+DROP INDEX IF EXISTS lock_lock_id_expires;
+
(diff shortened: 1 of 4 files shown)


migrations/000056_UpdateLocks.up.sql
```diff
@@ -0,0 +1,58 @@
+-- Copyright 2021 Google LLC
```
sethvargo, line 1: Diff against https://github.com/google/exposure-notifications-server/blob/fe59262d1950e933d9e3963054f1acf244a58b6f/migrations/000003_locking_procedures.up.sql

migrations/000056_UpdateLocks.down.sql
```diff
@@ -0,0 +1,20 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+DROP FUNCTION IF EXISTS AcquireLock;
```
mikehelmick, line 17: Shouldn’t down restore the old function?

mikehelmick:
Lock unit tests need updating now..

migrations/000056_UpdateLocks.down.sql
sethvargo: Done. Also dropped the index.

mikehelmick (approved):

PR #1324, opened 2021-02-01
Switch to ReadCommitted isolation level for locking

sethvargo:
I confirmed that this test I wrote demonstrates the behavior we're seeing on production. Switching to `ReadCommitted` fixes the error.

⚠️ There's also a behavior change on the unlock functions that I'd like close review on! ⚠️ I will call it out in the PR specifically.

**Release Note**

```release-note
Switch to ReadCommitted isolation level for locking
```

### DIFF
--- a/internal/database/lock.go
+++ b/internal/database/lock.go
@@ -40,8 +40,10 @@ type UnlockFn func() error
 // the transaction is rolled back.
 // The lockIDs are sorted by normal ascending string sort order before obtaining the locks.
 func (db *DB) MultiLock(ctx context.Context, lockIDs []string, ttl time.Duration) (UnlockFn, error) {
+	logger := logging.FromContext(ctx).Named("database.MultiLock")
+
 	if len(lockIDs) == 0 {
-		return nil, fmt.Errorf("no lockIDs presented")
+		return nil, fmt.Errorf("no lockIDs")
 	}
 
 	lockOrder := make([]string, len(lockIDs))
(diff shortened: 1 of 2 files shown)


internal/database/lock.go
```diff
@@ -50,39 +52,44 @@ func (db *DB) MultiLock(ctx context.Context, lockIDs []string, ttl time.Duration
 	sort.Strings(lockOrder)
 
 	var expires time.Time
-	err := db.InTx(ctx, pgx.Serializable, func(tx pgx.Tx) error {
+	if err := db.InTx(ctx, pgx.ReadCommitted, func(tx pgx.Tx) error {
 		for _, lockID := range lockOrder {
-			row := tx.QueryRow(ctx, `SELECT AcquireLock($1, $2)`, lockID, int(ttl.Seconds()))
+			row := tx.QueryRow(ctx, `SELECT AcquireLock($1, $2)`, lockID, int32(ttl.Seconds()))
```
sethvargo, line 57: The underlying postgres value is an int, so explicitly choosing 32 is safer

internal/database/lock.go
```diff
@@ -50,39 +52,44 @@ func (db *DB) MultiLock(ctx context.Context, lockIDs []string, ttl time.Duration
...
+				logger := logger.With("lock_id", lockID)
+
 				row := tx.QueryRow(ctx, `SELECT ReleaseLock($1, $2)`, lockID, expires)
 				var released bool
 				if err := row.Scan(&released); err != nil {
-					return err
+					return fmt.Errorf("failed to scan lock.released: %w", err)
 				}
-				if !released {
-					return fmt.Errorf("cannot delete lock %q that no longer belongs to you; it likely expired and was taken by another process", lockID)
```
sethvargo, line 83: ⚠️ Behavior change. It feels wrong (and makes testing _very_ difficult) to return an error if the lock was failed to release. With the new function, releasing the lock will _always_ succeed if the procedure finishes.

internal/database/lock.go
```diff
@@ -93,40 +100,46 @@ func makeMultiUnlockFn(ctx context.Context, db *DB, lockIDs []string, expires ti
...
-			`, lockID, expires)
+		return db.InTx(ctx, pgx.ReadCommitted, func(tx pgx.Tx) error {
+			row := tx.QueryRow(ctx, `SELECT ReleaseLock($1, $2)`, lockID, expires)
 			var released bool
 			if err := row.Scan(&released); err != nil {
-				return err
+				return fmt.Errorf("failed to scan lock.released: %w", err)
 			}
-			if !released {
-				return fmt.Errorf("cannot delete lock %q that no longer belongs to you; it likely expired and was taken by another process", lockID)
```
sethvargo, line 127: ⚠️ Behavior change. It feels wrong (and makes testing _very_ difficult) to return an error if the lock was failed to release. With the new function, releasing the lock will _always_ succeed if the procedure finishes.

sethvargo (commented):
/assign @mikehelmick 
/assign @whaught

whaught (approved):

PR #1325, opened 2021-02-02
Drop usage of serializable transactions

sethvargo:
I audited all uses, and none of them should be serializable; ReadCommitted is more than fine. I also removed the helper to discourage accidental future use.

**Release Note**

```release-note
Drop usage of serializable transactions
```

/assign @mikehelmick


mikehelmick:
needs a go mod tidy
removed dependencies

mikehelmick (approved):

PR #1337, opened 2021-02-03
Do not log lock acquisition failure when the lock is already held

sethvargo:
**Release Note**

```release-note
Do not log lock acquisition failure when the lock is already held
```

/assign @mikehelmick


mikehelmick (approved):

mikehelmick:
lint...

```
internal/export/worker.go:28:2: package "github.com/google/exposure-notifications-server/internal/database" is being imported more than once (ST1019)
	internal/export/worker.go:29:2: other import of "github.com/google/exposure-notifications-server/internal/database"
```

PR #1341, opened 2021-02-04
Pull logger and request_id from context in jwks service

sethvargo:
This updates the jwks service to populate and pull critical information from the request context instead of the initial context, similar to the verifications server.

Yes, a lot of this middleware is similar to the verification server, but it's _just_ slightly different enough that it doesn't make sense to try and unify (yet).

**Release Note**

```release-note
Pull logger and request_id from context in jwks service
```


whaught (approved):

sethvargo:
Wut... test failure

sethvargo:
Dunno why the go.mod got sad

PR #1343, opened 2021-02-04
Update ignore_changes to avoid recurring diff

sethvargo:
**Release Note**

```release-note
Update ignore_changes to avoid recurring diff in Terraform
```

/assign @whaught


mikehelmick (approved):
thanks!

whaught (approved):

whaught:
This is great - I think this'll make deployments far less scary to the ininitiated

PR #1349, opened 2021-02-05
using PathPrefix as a path catch-all so that v1alpha API works properly

### LINKED ISSUE #1348: v1alpha API fails with new mux handler
### TL;DR
Calls to exposure service /xxx are not handled by HandleV1Alpha1

**Expected behavior**
Calls like /publish should invoke the v1alpha handler

**Observed behavior**
Get a 404 response

### Reproduction
Hit key server exposure service with /xxx endpoint

**Additional information**
mux.NewRouter() matches "/" exactly, unlike http.NewServeMux() which is a catch-all

bschlaman:
Fixes #1348

## Proposed Changes

* Switch catch-all path to use gorilla/mux PathPrefix func to fix v1alpha handling

**Release Note**

```release-note
fixing exposure service handling of "/" path to fix v1alpha API
```

### DIFF
--- a/cmd/exposure/main.go
+++ b/cmd/exposure/main.go
@@ -84,7 +84,7 @@ func realMain(ctx context.Context) error {
 
 	// Serving of v1alpha1 is on by default, but can be disabled through env var.
 	if config.EnableV1Alpha1API {
-		r.Handle("/", handler.HandleV1Alpha1())
+		r.PathPrefix("/").Handler(handler.HandleV1Alpha1())
 	}
 
 	srv, err := server.New(config.Port)


cmd/exposure/main.go
```diff
@@ -84,7 +84,7 @@ func realMain(ctx context.Context) error {
 
 	// Serving of v1alpha1 is on by default, but can be disabled through env var.
 	if config.EnableV1Alpha1API {
-		r.Handle("/", handler.HandleV1Alpha1())
+		r.PathPrefix("/").Handler(handler.HandleV1Alpha1())
```
whaught, line 87: Do we want
```
r.Handle("", handler.HandleV1Alpha1())
```

I'm not sure if PathPrefix will match everything with `/`

mikehelmick: Suggest enumerating these instead of handling all paths.

"/"
"/publish"
"/publish/"
"/v1alpha1/publish"
"/v1alpha1/publish/"

bschlaman:
Closing for now, as I don't see evidence that urls besides "/" are actively being used in prod for v1alpha

Issue #1295, opened 2021-01-15
Update alert structure


mariliamelo:
### TL;DR
Make sure that we have SLO based alerts and all critical user journeys are covered.

sethvargo:
All background jobs have been refactored to be forward-progress alerts. What additional alerts are you thinking about @mariliamelo?

PR #1354, opened 2021-02-05
Separate paging and non-paging notification channels

mariliamelo:
Part of #1295

## Proposed Changes

* Add paging and non-paging channels for alerts

**Release Note**

```release-note
Paging and non-paging email channels should be configured in the Terraform files for each individual project.
```


sethvargo (approved):
This will require updates to all the configs in the infra repo

sethvargo (approved):
/lgtm

PR #1362, opened 2021-02-09
Check database connection in health check

sethvargo:
**Release Note**

```release-note
Check database connection in health check
```

### DIFF
--- a/internal/publish/publish.go
+++ b/internal/publish/publish.go
@@ -52,7 +52,7 @@ const (
 
 type Server struct {
 	config                *Config
-	serverenv             *serverenv.ServerEnv
+	env                   *serverenv.ServerEnv
 	transformer           *model.Transformer
 	database              *database.PublishDB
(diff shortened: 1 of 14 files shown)


internal/publish/publish.go
```diff
@@ -140,7 +140,7 @@ func (s *Server) Routes(ctx context.Context) *mux.Router {
 	r.Use(middleware.PopulateLogger(logger))
 	r.Use(middleware.ProcessMaintenance(s.config))
 
-	r.Handle("/health", server.HandleHealthz())
+	r.Handle("/health", server.HandleHealthz(s.env.Database()))
```
mikehelmick, line 143: I think we need at least some rate limiting here - at least on a per-instance basis
since this is not an authenticated endpoint

sethvargo: Done

mikehelmick (approved):

whaught (approved):

PR #1364, opened 2021-02-09
Add common observability code

yegle:


mikehelmick:
see build errors

sethvargo (commented):
@yegle I think this is already happening? We just refactored to use the o11y middleware and I _think_ it's doing the right thing.

yegle:
> @yegle I think this is already happening? We just refactored to use the o11y middleware and I _think_ it's doing the right thing.

The `WithBuildInfo` is hardcoded to use key server's build info. There's a slight change here to make it more general.

Let me split this to two PRs. It gets a bit confusing.

sethvargo (commented):
Are the go mod changes intentional?

pkg/observability/common.go
```diff
@@ -14,8 +14,95 @@
...
+)
+
+var (
+	// ResultOK add a tag indicating the API call is a success.
+	ResultOK = tag.Upsert(ResultTagKey, "OK")
+	// ResultNotOK add a tag indicating the API call is a failure.
+	ResultNotOK = ResultError("NOT_OK")
+)
+
+var (
```
sethvargo, line 76: I think a lot of this is already defined in https://github.com/google/exposure-notifications-server/blob/main/pkg/observability/observability.go. We may want to combine/reconcile the two files.

sethvargo: https://github.com/google/exposure-notifications-server/blob/main/pkg/observability/observability.go#L37

yegle: Yes, those are removed.

sethvargo (commented):
Sorry, I should have clarified. I think it's weird that we have constants across observability.go and common.go. I think we should merge the contents of both files into 1, pref observability.go since it follows the naming pattern of other packages.

sethvargo (approved):
/lgtm

PR #1366, opened 2021-02-09
Update WithBuildInfo() to be more generic.

yegle:
We can then share this method to be used by verification server.


sethvargo (approved):

pkg/observability/common.go
```diff
@@ -0,0 +1,7 @@
+package observability
```
mikehelmick, line 1: file missing copyright notice

whaught (approved):

PR #1371, opened 2021-02-10
Move e2e into integration, simplify helpers

sethvargo:
This also removes the e2e-terraform-ci (we have the linter in GitHub Actions now).

**Release Note**

```release-note
Move e2e into integration, simplify helpers
```


internal/publish/model/exposure_model.go
```diff
@@ -570,12 +570,12 @@ func (t *Transformer) TransformPublish(ctx context.Context, inData *verifyapi.Pu
 	if len(inData.Keys) == 0 {
 		msg := "no exposure keys in publish request"
 		logger.Debugf(msg)
-		return nil, fmt.Errorf(msg)
+		return &TransformPublishResult{}, fmt.Errorf(msg)
```
sethvargo, line 573: Twas bug

internal/publish/model/exposure_model.go
```diff
@@ -570,12 +570,12 @@ func (t *Transformer) TransformPublish(ctx context.Context, inData *verifyapi.Pu
...
 	}
 	if len(inData.Keys) > t.maxExposureKeys {
 		msg := fmt.Sprintf("too many exposure keys in publish: %v, max of %v is allowed", len(inData.Keys), t.maxExposureKeys)
 		logger.Debugf(msg)
-		return nil, fmt.Errorf(msg)
+		return &TransformPublishResult{}, fmt.Errorf(msg)
```
sethvargo, line 578: Twas bug

internal/project/random.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"encoding/hex"
+	"fmt"
+)
+
+// RandomHexString generates a random string of the provided length.
+func RandomHexString(len int) (string, error) {
```
whaught, line 25: Is this taken from Verification? I think we have a unit test that goes along with it (and I suppose we can have verification depend on this in the future)

sethvargo: It is. I can do that as a follow-up. I want to do a full diff and see if there are other things that should be moved over

PR #1374, opened 2021-02-12
Set all cron jobs to retry 3 times

mariliamelo:
Standardize all cloud scheduler jobs to retry 3 times.


sethvargo (approved):

mikehelmick:
What's the retry strategy that cloud scheduler uses?

sethvargo:
The default is immediately after failure. We can customize it if we want: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloud_scheduler_job#retry_count

PR #1375, opened 2021-02-12
add request test, use const for http status

whaught:
## Proposed Changes

* Add another test for rotation
* Replace http code number with consts for readability

### DIFF
--- a/internal/export/batcher.go
+++ b/internal/export/batcher.go
@@ -51,7 +51,7 @@ func (s *Server) handleCreateBatches() http.Handler {
 			if errors.Is(err, database.ErrAlreadyLocked) {
 				stats.Record(ctx, mBatcherLockContention.M(1))
 				logger.Infow("already locked")
-				w.WriteHeader(200)
+				w.WriteHeader(http.StatusOK)
 				return
 			}
 
@@ -94,11 +94,11 @@ func (s *Server) handleCreateBatches() http.Handler {
 			switch {
 			case errors.Is(err, context.DeadlineExceeded):
 				logger.Infow("batch creation timed out")
-				w.WriteHeader(200)
+				w.WriteHeader(http.StatusOK)
 				return
 			case errors.Is(err, context.Canceled):
 				logger.Infow("batch creation canceled")
-				w.WriteHeader(200)
+				w.WriteHeader(http.StatusOK)
 				return
 			default:
 				logger.Errorw("failed to create batches", "error", err)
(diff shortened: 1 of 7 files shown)


internal/export/batcher.go
```diff
@@ -94,11 +94,11 @@ func (s *Server) handleCreateBatches() http.Handler {
 			switch {
 			case errors.Is(err, context.DeadlineExceeded):
 				logger.Infow("batch creation timed out")
-				w.WriteHeader(200)
+				w.WriteHeader(http.StatusOK)
```
whaught, line 97: @mikehelmick these don't seem like the right statuses for these?

internal/export/batcher.go
```diff
@@ -51,7 +51,7 @@ func (s *Server) handleCreateBatches() http.Handler {
 			if errors.Is(err, database.ErrAlreadyLocked) {
 				stats.Record(ctx, mBatcherLockContention.M(1))
 				logger.Infow("already locked")
-				w.WriteHeader(200)
+				w.WriteHeader(http.StatusOK)
```
sethvargo, line 54: I actually have a note to go through and make helpers for these and always return JSON.

200 is the right code though I think. It means someone else is already running. We don't want Cloud Scheduler to retry in this case.

internal/keyrotation/rotate_test.go
```diff
@@ -32,6 +34,42 @@ import (
...
+	}
+
+	env := serverenv.New(ctx, serverenv.WithKeyManager(kms), serverenv.WithDatabase(testDB))
+	server, err := NewServer(config, env)
+	if err != nil {
+		t.Fatalf("got unexpected error: %v", err)
+	}
+	handler := server.handleRotateKeys()
+
+	r := httptest.NewRequest("GET", "/", nil)
```
sethvargo, line 59: Use NewRequestWithContext instead of New and then Clone

sethvargo (changes requested):

internal/export/batcher.go
mikehelmick: this one should probably be 500

internal/export/batcher.go
whaught: Agreed on that one (although maybe we should comment that it's for scheduler - in normal REST world I might expect 409 StatusConflict or something).

whaught: but what about the above in the export batcher enum?

mikehelmick: right, this is 200 so the CS doesn't report failure, I know it's a bit weird

mikehelmick (approved):

PR #1376, opened 2021-02-13
fix formatting on api docs

mikehelmick:
## Proposed Changes

* previous formatting didn't read great on pkg.go.dev

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/api/v1/exposure_types.go
+++ b/pkg/api/v1/exposure_types.go
@@ -68,9 +68,7 @@ const (
 
 // Publish represents the body of the PublishInfectedIds API call.
 // temporaryExposureKeys: Required and must have length >= 1 and <= 21 (`maxKeysPerPublish`)
-// healthAuthorityID: The identifier for the mobile application.
-//  - Android: The App Package AppPackageName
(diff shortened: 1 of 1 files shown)


pkg/api/v1/exposure_types.go
```diff
@@ -68,9 +68,7 @@ const (
 
 // Publish represents the body of the PublishInfectedIds API call.
 // temporaryExposureKeys: Required and must have length >= 1 and <= 21 (`maxKeysPerPublish`)
-// healthAuthorityID: The identifier for the mobile application.
-//  - Android: The App Package AppPackageName
```
whaught, line 72: Did you intend to remove `Android` and `IOS` or should we put those somewhere?

whaught: /lgtm /hold

whaught: /lgtm /hold

mikehelmick: yes, that was leftover from v1alpha1

PR #1382, opened 2021-02-16
removing version syntax to support GOPATH mode

bschlaman:
**Release Note**

```release-note
removing version from golangci-lint to support GOPATH mode
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ MD_FILES = $(shell find . -name \*.md)
 # locally. There is a chance that CI detects linter errors that are not found
 # locally, but it should be rare.
 lint:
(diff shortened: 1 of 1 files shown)


bschlaman:
Could also just turn go modules on

sethvargo:
I think we want to keep the version on so that we run the same version in CI as local dev.

bschlaman:
Sounds good, then I think we'll have to keep go modules on

sethvargo:
I think they need to stay off because otherwise they'll modify the go.mod and go.sum in the project

sethvargo:
@bschlaman does that make sense?

ryanwolcott321 (approved):

bschlaman:
Yes makes sense @sethvargo - let me see if I can find a good way to do this.  I've seen online that the only way might be to change into another directory where go.mod isn't there and return when done

PR #1384, opened 2021-02-16
Update security alerts for HumanAccessSecret and HumanDecryptedValue

sethvargo:
The Cloud KMS audit logs were not matching as expected. This fixes that.

**Release Note**

```release-note
Update accuracy for security alerts for HumanAccessSecret and HumanDecryptedValue
```

/assign @mariliamelo

### DIFF
--- a/terraform/alerting/alerts.tf
+++ b/terraform/alerting/alerts.tf
@@ -54,9 +54,9 @@ resource "google_logging_metric" "human_accessed_secret" {
   name    = "human_accessed_secret"
   project = var.project
   filter  = <<EOT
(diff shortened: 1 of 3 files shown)


terraform/alerting/alerts.tf
```diff
@@ -162,7 +163,7 @@ resource "google_monitoring_alert_policy" "HumanDecryptedValue" {
       duration = "60s"
```
mariliamelo, line 163: You probably need to set the duration to zero to be alerted as soon as it happens. The current config will only fire after the current condition has been sustained for 60 seconds.

PR #1387, opened 2021-02-16
jwks parallelization limits

### LINKED ISSUE #1380: jwks service times out
### TL;DR
The jwks service sporadically times out, likely when building the cache. 

We should consider moving computation to be more frequent and in-parallel, serving stale data when the result isn't rebuilt in time.

mikehelmick:
Fixes #1380

## Proposed Changes

* limit the number of parallel JWKS workers
* add overall timeout
* increase test coverage

**Release Note**

```release-note
Better handling of parallel requests in JWKS service.
```

### DIFF
--- a/internal/jwks/config.go
+++ b/internal/jwks/config.go
@@ -34,7 +34,7 @@ type Config struct {
 	Port string `env:"PORT, default=8080"`
 
 	// MaxRuntime is how long an individual handler should run.
-	MaxRuntime time.Duration `env:"MAX_RUNTIME, default=9m"`
+	MaxRuntime time.Duration `env:"MAX_RUNTIME, default=10m"`
 
 	// RequestTimeout is the client per-request timeout when accessing
 	// remote JWKS documents.
--- a/internal/jwks/config.go
+++ b/internal/jwks/config.go
@@ -33,11 +33,17 @@ type Config struct {
 
 	Port string `env:"PORT, default=8080"`
 
-	// RequestTimeout is the per-request amount of time to wait for a response
-	// before timing out.
-	RequestTimeout time.Duration `env:"REQUEST_TIMEOUT, default=5s"`
+	// MaxRuntime is how long an individual handler should run.
+	MaxRuntime time.Duration `env:"MAX_RUNTIME, default=9m"`
+
+	// RequestTimeout is the client per-request timeout when accessing
+	// remote JWKS documents.
+	RequestTimeout time.Duration `env:"REQUEST_TIMEOUT, default=30s"`
 
 	KeyCleanupTTL time.Duration `env:"HEALTH_AUTHORITY_KEY_CLEANUP_TTL, default=720h"` // 30 days
+
+	// MaxWorkers is the number of parallel JWKS updates that can occur.
+	MaxWorkers uint `env:"MAX_WORKERS, default=5"`
 }
 
 func (c *Config) DatabaseConfig() *database.Config {
--- a/internal/jwks/handle_update_all.go
+++ b/internal/jwks/handle_update_all.go
@@ -15,6 +15,7 @@
 package jwks
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/http"
@@ -29,7 +30,9 @@ const jwksLockID = "jwks-import"
 
 func (s *Server) handleUpdateAll() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		ctx := r.Context()
+		ctx, cancel := context.WithTimeout(r.Context(), s.config.MaxRuntime)
+		defer cancel()
+
 		logger := logging.FromContext(ctx).Named("handleUpdateAll")
 
 		unlock, err := s.manager.db.Lock(ctx, jwksLockID, time.Minute)
--- a/internal/jwks/jwks.go
+++ b/internal/jwks/jwks.go
@@ -23,7 +23,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"io/ioutil"
-	"math/rand" /* #nosec */
+	"math/rand" //nolint:gosec
 	"net/http"
 	"sort"
 	"strings"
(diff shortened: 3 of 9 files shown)


internal/jwks/config.go
```diff
@@ -33,11 +33,17 @@ type Config struct {
 
 	Port string `env:"PORT, default=8080"`
 
-	// RequestTimeout is the per-request amount of time to wait for a response
-	// before timing out.
-	RequestTimeout time.Duration `env:"REQUEST_TIMEOUT, default=5s"`
+	// MaxRuntime is how long an individual handler should run.
+	MaxRuntime time.Duration `env:"MAX_RUNTIME, default=9m"`
```
sethvargo, line 37: My OCD kindly requests you make this 10m.

internal/jwks/jwks.go
```diff
@@ -248,18 +251,41 @@ func (mgr *Manager) UpdateAll(ctx context.Context) error {
...
+	// Pre-filter out health authorities where discovery isn't enabled.
+	healthAuthorities := make([]*model.HealthAuthority, 0, len(allHealthAuthorities))
+	for _, ha := range allHealthAuthorities {
+		if ha.JWKS() {
+			healthAuthorities = append(healthAuthorities, ha)
+		}
+	}
+
+	// Crypto secure random isn't necessary here, just shuffling the scheduling order on each run.
+	rand.Shuffle(len(healthAuthorities), func(i, j int) {
```
sethvargo, line 268: This rand needs to be seeded, ideally with the crypto/rand helper thing I wrote

internal/jwks/jwks.go
```diff
@@ -248,18 +251,41 @@ func (mgr *Manager) UpdateAll(ctx context.Context) error {
...
+		healthAuthorities[i], healthAuthorities[j] = healthAuthorities[j], healthAuthorities[i]
+	})
+
 	var merr *multierror.Error
 	var merrLock sync.Mutex
-
+	sem := semaphore.NewWeighted(int64(mgr.maxWorkers))
```
sethvargo, line 274: This can overflow. I'd suggest making the config an int and asserting its non-negative somewhere

internal/jwks/jwks.go
```diff
@@ -248,18 +251,41 @@ func (mgr *Manager) UpdateAll(ctx context.Context) error {
...
 	var wg sync.WaitGroup
 	for _, ha := range healthAuthorities {
+		// Rate limit the number of concurrent workers with a semaphore.
+		if err := sem.Acquire(ctx, 1); err != nil {
+			logger.Errorw("failed to acquire semaphore", "error", err)
+			merrLock.Lock()
+			merr = multierror.Append(merr, fmt.Errorf("failed to processes %v: %w", ha.Name, err))
+			merrLock.Unlock()
+			break
```
sethvargo, line 283: continue?

internal/jwks/jwks_test.go
```diff
@@ -256,3 +257,63 @@ func TestStrip(t *testing.T) {
...
+		},
+		{
+			name:  "failure",
+			delay: 2 * time.Second,
+			err:   "failed to processes ha.1: reading connection",
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
```
sethvargo, line 284: t.Parallel

internal/verification/model/health_authority.go
```diff
@@ -37,6 +37,11 @@ type HealthAuthority struct {
 	EnableStatsAPI bool
 }
 
+// JWKS returns true if JWKS discovery is enabled for this health authority.
+func (ha *HealthAuthority) JWKS() bool {
```
sethvargo, line 41: JWKSEnabled or something more descriptive?

internal/jwks/handle_update_all.go
```diff
@@ -29,7 +30,9 @@ const jwksLockID = "jwks-import"
 
 func (s *Server) handleUpdateAll() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		ctx := r.Context()
+		ctx, cancel := context.WithTimeout(r.Context(), s.config.MaxRuntime)
```
sethvargo, line 33: I think you need to handle the context error somewhere else in this body then?

sethvargo (changes requested):

internal/jwks/handle_update_all.go
mikehelmick: this is to set up the semaphore so it will have a maximum wait time on the acquire

internal/jwks/jwks.go
mikehelmick: it's coming from a uint - the lint was because I was already converting it somewhere else

internal/jwks/jwks.go
mikehelmick: no - if we can't get the semaphore, it's almost certainly because the context timed out

internal/jwks/jwks.go
```diff
@@ -23,6 +23,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"io/ioutil"
+	"math/rand" /* #nosec */
```
sethvargo, line 26: I think it's `//nolint:gosec`

internal/jwks/jwks.go
```diff
@@ -248,18 +252,41 @@ func (mgr *Manager) UpdateAll(ctx context.Context) error {
...
 
+	// Pre-filter out health authorities where discovery isn't enabled.
+	healthAuthorities := make([]*model.HealthAuthority, 0, len(allHealthAuthorities))
+	for _, ha := range allHealthAuthorities {
+		if ha.JWKSEnabled() {
+			healthAuthorities = append(healthAuthorities, ha)
+		}
+	}
+
+	r := rand.New(cryptorand.NewSource())
```
sethvargo, line 268: //nolint:gosec

internal/jwks/jwks_test.go
```diff
@@ -256,3 +257,64 @@ func TestStrip(t *testing.T) {
...
+		},
+		{
+			name:  "failure",
+			delay: 2 * time.Second,
+			err:   "failed to processes ha.1: reading connection",
+		},
+	}
+
+	for _, tc := range cases {
+		//scopelint:ignore
```
sethvargo, line 284: Don't ignore that - that linter is actually telling you that something is wrong 😄 

You need `tc := tc` here. Otherwise you might run the same test N times where N is the last value in the range.

sethvargo (approved):
/lgtm

PR #1396, opened 2021-02-22
enforce that STATS_UPLOAD_MINIMUM is set to a value >= 10

mikehelmick:
## Proposed Changes

* enforce that STATS_UPLOAD_MINIMUM is set to a value >= 10

**Release Note**

```release-note
enforce that STATS_UPLOAD_MINIMUM is set to a value >= 10
```


sethvargo (approved):
/lgtm

PR #1397, opened 2021-02-23
Fixing key revision probability for generate and adding a report type confirmed override

bschlaman:
## Proposed Changes

* Check that the random value is < ChanceOfKeyRevision, not <= so that a chance of 0 really is 0 instead of 1/100
* Add env variable FORCE_CONFIRMED so that an exact number of exposures/keys can can be published

**Release Note**

```release-note
Fixing key revision probability for generate and adding a report type confirmed override
```

### DIFF
--- a/internal/generate/handle_generate.go
+++ b/internal/generate/handle_generate.go
@@ -137,7 +137,11 @@ func (s *Server) generateKeysInRegion(ctx context.Context, region string) error 
 
 		reportType := verifyapi.ReportTypeClinical
 		if !generateRevisedKeys {
-			reportType, err = util.RandomReportType()
+			if forceConfirmed := s.config.ForceConfirmed; forceConfirmed {
+				reportType = verifyapi.ReportTypeConfirmed
+			} else {
+				reportType, err = util.RandomReportType()
+			}
 			if err != nil {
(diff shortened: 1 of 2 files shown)


internal/generate/handle_generate.go
```diff
@@ -133,13 +133,17 @@ func (s *Server) generateKeysInRegion(ctx context.Context, region string) error
...
 		}
-		generateRevisedKeys := val <= s.config.ChanceOfKeyRevision
+		generateRevisedKeys := val < s.config.ChanceOfKeyRevision
 
 		reportType := verifyapi.ReportTypeClinical
 		if !generateRevisedKeys {
-			reportType, err = util.RandomReportType()
-			if err != nil {
-				return fmt.Errorf("failed to generate report type: %w", err)
+			if forceConfirmed := s.config.ForceConfirmed; forceConfirmed {
```
mikehelmick, line 140: this should just be
`if s.config.ForceConfirmed`

no reason to assign to another var

bschlaman: ah yep thanks, was leftover from some other testing

mikehelmick (approved):

PR #1432, opened 2021-03-05
Add stackdriver export failure alert

mariliamelo:
Part of #1295

### DIFF
--- a/docs/playbooks/alerts/StackdriverExportFailed.md
+++ b/docs/playbooks/alerts/StackdriverExportFailed.md
@@ -1,9 +1,7 @@
 # StackdriverExportFailed
 
-OpenCensus failed to export metrics to Stackdriver.
-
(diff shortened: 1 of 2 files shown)


docs/playbooks/alerts/StackdriverExportFailed.md
```diff
@@ -0,0 +1,29 @@
+# StackdriverExportFailed
+
+OpenCensus failed to export metrics to Stackdriver.
+
+This means all other alerts we have configured won't work as the alert depend
```
sethvargo, line 5: This sounds unnecessarily alarmist. There are many reasons it could fail, including the container is shutting down.

mariliamelo: Indeed. Let me update this playbook.

sethvargo (approved):

PR #1433, opened 2021-03-05
[WIP] API sketch for transporting vaccine status

mikehelmick:
## Proposed Changes

* Design sketch of how the APIs might change to capture and report on vaccine status for use in risk scoring

**Release Note**

```release-note
NONE
```


gurayAlsac:
Interesting. Is the idea that we would downweight an exposure from someone vaccinated? Or get a sense of vaccine dissemination/efficacy in light of potentially more resistant variants?

mikehelmick:
> Interesting. Is the idea that we would downweight an exposure from someone vaccinated? Or get a sense of vaccine dissemination/efficacy in light of potentially more resistant variants?

How this would be reflected in risk scoring is unclear - using this a a detector of mutations is a possible use.

sethvargo:
A few things:

1. I'd like a flag to be able to quickly turn this off. Introducing a new field _has_ broken some clients in the past, so we should have a runtime way to quickly disable this without doing a full new deployment.
2. Should we distinguish between "not given" and "unknown"?

PR #1441, opened 2021-03-09
Switch CODEOWNERS to team

sethvargo:
This way, we can manage the team and access controls via the team instead of managing individual people.

**Release Note**

```release-note
NONE
```


sethvargo:
This will also fix the fact that all codeowners are assigned. Using the team, it automatically picks 2 reviewers (round robin) and we can exclude people from reviews (Guray and Brad are already excluded)

whaught (approved):

PR #1443, opened 2021-03-09
add ability for export-importer to import an export file w/ travler=true

mikehelmick:
## Proposed Changes

* export-importer can set traveller status to true (for the whole import)

**Release Note**

```release-note
export-importer can set traveller status to true (for the whole import)
```

### DIFF
--- a/migrations/000079_ExportImportTraveler.up.sql
+++ b/migrations/000079_ExportImportTraveler.up.sql
@@ -16,6 +16,9 @@ BEGIN;
 
 ALTER TABLE exportimport
   ADD COLUMN traveler BOOL DEFAULT false;
+
+UPDATE exportimport SET traveler=false;
+
 ALTER TABLE exportimport
   ALTER COLUMN traveler SET NOT NULL;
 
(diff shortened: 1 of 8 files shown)


migrations/000079_ExportImportTraveler.up.sql
```diff
@@ -0,0 +1,22 @@
...
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+ALTER TABLE exportimport
+  ADD COLUMN traveler BOOL DEFAULT false;
+ALTER TABLE exportimport
```
sethvargo, line 19: I think this will either need to be a separate transaction or you need to do a data migration here. Setting a column as a default doesn't backfill existing rows.

```sql
ALTER TABLE add column;
UPDATE exportimport SET traveler = false;
ALTER COLUMN traveler SET NOT NULL;
```

sethvargo (changes requested):

sethvargo (approved):

Issue #1411, opened 2021-03-01
log based metrics for export file downloading


mikehelmick:
### TL;DR

Create terraform template so operators can automate the creation log logs based metrics around index.txt downloads through Cloud CDN (applies only when operating on GCP). 

The operator would need a list of these in their main.tf, i.e. a list of the URLs

metric type: counter
units: 1
The filter selection is 

```
resource.type="http_load_balancer"
httpRequest.requestUrl="https://Hostname/path/index.txt"
```

Label

iOS

![image](https://user-images.githubusercontent.com/92319/109551906-00a5bc00-7a86-11eb-9bf2-c55d429f779c.png)

Android

![image](https://user-images.githubusercontent.com/92319/109551940-0d2a1480-7a86-11eb-8fd5-a0795c7d8b2d.png)

PR #1447, opened 2021-03-10
Add logs-based metric for when exports are downloaded

sethvargo:
Fixes https://github.com/google/exposure-notifications-server/issues/1411

**Release Note**

```release-note
Add a new metric `export_file_downloaded`, which is emitted when a device downloads an export. The metric extracts the export path (for multi-tenant installations) and the platform via labels for further aggregation. This is on by default but can be disabled by setting `capture_export_file_downloads` to false. This only applies to Google Cloud.
```

### DIFF
--- a/terraform/alerting/alerts.tf
+++ b/terraform/alerting/alerts.tf
@@ -298,3 +298,37 @@ resource "google_monitoring_alert_policy" "HumanDecryptedValue" {
 
   notification_channels = [for x in values(google_monitoring_notification_channel.paging) : x.id]
 }
+
+resource "google_logging_metric" "export_file_downloaded" {
+  count = var.capture_export_file_downloads ? 1 : 0
+
+  name    = "export_file_downloaded"
+  project = var.project
+  filter  = <<EOT
+resource.type="http_load_balancer"
+httpRequest.requestUrl=~"/index.txt$"
+httpRequest.status=200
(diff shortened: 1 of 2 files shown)


terraform/alerting/alerts.tf
```diff
@@ -298,3 +298,37 @@ resource "google_monitoring_alert_policy" "HumanDecryptedValue" {
...
 }
+
+resource "google_logging_metric" "export_file_downloaded" {
+  count = var.capture_export_file_downloads ? 1 : 0
+
+  name    = "export_file_downloaded"
+  project = var.project
+  filter  = <<EOT
+resource.type="http_load_balancer"
+httpRequest.requestUrl=~"/index.txt$"
```
mikehelmick, line 309: This needs to be parameterized by the folder (for per-country active counts)

mikehelmick (changes requested):

terraform/alerting/alerts.tf
sethvargo: It is. It's named `path`, but I can change the name.

mikehelmick (dismissed):

sethvargo:
@mikehelmick can haz re-review? I had to add many escapings to match what the API expects.

mikehelmick (approved):

PR #1462, opened 2021-03-17
Not throwing away my shot

mikehelmick:
## Proposed Changes

* Vaccine status experiment.
* Will be contained to the experimental branch - if merged to main, much more work needs to be done
    * Add to federation
    * Add to export-importer
   
**Release Note**

```release-note
NONE
```


mikehelmick:
rebased on experimental after migrations pushed through main

internal/integration/integration_test.go
```diff
@@ -73,10 +74,17 @@ func TestEndToEnd(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	b := make([]byte, 1)
```
sethvargo, line 77: Can you use https://github.com/google/exposure-notifications-server/blob/main/pkg/util/generate.go#L47 here instead? I'm trying to centralize random number generate as much as possible.

sethvargo (approved):

PR #1466, opened 2021-03-17
Add tests for admin authorized_apps

sethvargo:
Also fix an issue with the serializeForm reflection for slices

**Release Note**

```release-note
Add tests for admin authorized_apps
```


internal/admin/handle_authorized_apps.go
```diff
@@ -54,11 +54,14 @@ func (s *Server) HandleAuthorizedAppsSave() func(c *gin.Context) {
 					ErrorPage(c, "Invalid request, app to edit not found.")
 					return
 				}
+				if authApp == nil {
```
sethvargo, line 57: Twas bug. This package returns nil if the app doesn't exist. I'm not sure if we should audit for instances or change the behavior of GetAuthorizedApp.

mikehelmick (approved):

PR #1483, opened 2021-03-22
Use an intermediate proxy for backups

sethvargo:
Unfortunately this is required to measure forward progress.

**Release Note**

```release-note
Introduce a new service: `backup`. Be sure to taint the null_resource builder during the Terraform apply to get the new service version.
```

### DIFF
--- a/internal/backup/handle_backup.go
+++ b/internal/backup/handle_backup.go
@@ -1,0 +1,136 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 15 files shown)


internal/backup/handle_backup.go
```diff
@@ -0,0 +1,136 @@
...
+		logger := logging.FromContext(ctx).Named("backup.HandleBackup")
+
+		req, err := s.buildBackupRequest(ctx)
+		if err != nil {
+			logger.Errorw("failed to build request", "error", err)
+			s.h.RenderJSON(w, http.StatusInternalServerError, err)
+			return
+		}
+
+		unlock, err := s.db.Lock(ctx, backupDatabaseLockID, s.config.MinTTL)
```
sethvargo, line 48: This isn't _entirely_ the same as the verification server. The lock prevents concurrent modifications, but it doesn't enforce a minTTL.

I can force it to enforce a minTTL by _not_ releasing the lock on `defer`. That felt hacky, but it gets us the exact behavior we want. Thoughts?

mikehelmick: right, it's not intended for the same thing, it was to prevent concurrent operations.

I think not releasing the lock is OK in the case of success to get the desired behaviour without making extensive modification. Comment it like crazy :)

sethvargo: Done, lmk about the comment

sethvargo:
@mikehelmick ready for another review

mikehelmick (approved):

PR #1484, opened 2021-03-23
Remove errorw and upgrade to latest ci lint

sethvargo:
**Release Note**

```release-note
Remove errorw and upgrade to latest ci lint
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -54,7 +54,6 @@ linters:
   - 'gosimple'
   - 'govet'
   - 'ineffassign'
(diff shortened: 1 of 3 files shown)


.golangci.yaml
```diff
@@ -54,7 +54,6 @@ linters:
   - 'gosimple'
   - 'govet'
   - 'ineffassign'
-  - 'interfacer'
```
sethvargo, line 57: This is deprecated apparently

mikehelmick (approved):

PR #1488, opened 2021-03-23
add metrics for missing public key and more logging

mikehelmick:
## Proposed Changes

* add metrics for when a HA exists, but has no public keys
* specifically log this type of issue

**Release Note**

```release-note
Add metrics for when a healthAuthorityID exists, but has no public keys
```

### DIFF
--- a/internal/publish/metrics.go
+++ b/internal/publish/metrics.go
@@ -29,6 +29,9 @@ var (
 
 	mLatencyMs = stats.Float64(publishMetricsPrefix+"requests", "publish requests", stats.UnitMilliseconds)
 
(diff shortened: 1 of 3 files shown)


internal/publish/metrics.go
```diff
@@ -29,6 +29,9 @@ var (
 
 	mLatencyMs = stats.Float64(publishMetricsPrefix+"requests", "publish requests", stats.UnitMilliseconds)
 
+	mNoPublicKeys = stats.Int64(publishMetricsPrefix+"no_pub_keys",
```
sethvargo, line 32: Nit, this should be:

```suggestion
	mNoPublicKey = stats.Int64(publishMetricsPrefix+"no_public_key",
```

sethvargo (approved):

PR #1489, opened 2021-03-23
don't set false values in export

mikehelmick:
## Proposed Changes

* don't set bool on false fields in export

**Release Note**

```release-note
NONE
```


internal/export/exportfile.go
```diff
@@ -156,7 +156,9 @@ func makeTEK(exp *publishmodel.Exposure) *export.TemporaryExposureKey {
 	if exp.IntervalCount != defaultIntervalCount {
 		pbek.RollingPeriod = proto.Int32(exp.IntervalCount)
 	}
-	pbek.Vaccinated = proto.Bool(exp.VaccineStatus)
+	if exp.VaccineStatus {
```
sethvargo, line 159: Can you comment why?

sethvargo (dismissed):

sethvargo (approved):

PR #1490, opened 2021-03-23
Switch export-importer to forward-progress alerting

sethvargo:
**Release Note**

```release-note
Switch export-importer to forward-progress alerting
```


internal/exportimport/handle_schedule.go
```diff
@@ -62,67 +63,77 @@ func (s *Server) handleSchedule() http.Handler {
...
-
-			resp, err := httpClient.Do(req)
-			if err != nil {
-				anyErrors = true
-				logger.Errorw("error downloading index file", "file", config.IndexFile, "error", err)
-				continue
-			}
+		var merr *multierror.Error
 
-			defer resp.Body.Close()
```
sethvargo, line 96: Twas bug. `defers` run when then function ends, not when the `for` loop ends. You could OOM for large numbers of schedules because you don't release the resources until everything has been processed. I moved the body into another function to ensure the body is released correctly.

internal/exportimport/handle_import.go
```diff
@@ -24,72 +24,77 @@ import (
...
 	"github.com/google/exposure-notifications-server/pkg/database"
 	"github.com/google/exposure-notifications-server/pkg/logging"
+	"github.com/hashicorp/go-multierror"
 
 	"go.opencensus.io/stats"
-	"go.opencensus.io/trace"
 )
 
-const lockPrefix = "import-lock-"
+const lockPrefix = "export-importer-worker-lock-"
```
mikehelmick, line 32: renaming these can cause things in flight to fail in weird ways

sethvargo: @mikehelmick fixed

mikehelmick (approved):

PR #1495, opened 2021-03-25
Also always log start and finish in debug mode

sethvargo:
Also clean up some unnecessary nil checks on multierror now that we've upgraded.

**Release Note**

```release-note
Also always log start and finish in debug mode for background jobs.
```


sethvargo:
```
directive `//nolint:staticcheck // may be set on v1 files.` is unused for linter staticcheck (nolintlint)
```

wut

whaught (approved):

PR #1519, opened 2021-04-29
add metrics for future dated JWT

mikehelmick:
## Proposed Changes

* introduce a new metric for future JWT (NBF and/or IAT in the future)

**Release Note**

```release-note
Adds a new metric 'en-server/jwt_not_yet_valid' that will be incremented when the server processes a JWT that is future dated according to either the IAT or NBF field of a properly signed verification certificate JWT.
```

### DIFF
--- a/internal/verification/phaverify.go
+++ b/internal/verification/phaverify.go
@@ -37,6 +37,7 @@ import (
 
 // ErrNoPublicKeys indicates no public keys were found when verifying the certificate.
 var ErrNoPublicKeys = errors.New("no active public keys for health authority")
+var ErrNotValidYet = errors.New("not valid yet (NBF or IAT) in the future")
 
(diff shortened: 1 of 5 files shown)


internal/verification/phaverify.go
```diff
@@ -129,15 +131,20 @@ func (v *Verifier) VerifyDiagnosisCertificate(ctx context.Context, authApp *aamo
...
+				return nil, ErrNotValidYet
+			}
+		}
 		return nil, err
 	}
 
 	if !token.Valid {
 		return nil, fmt.Errorf("invalid verificationPayload")
 	}
-	if err := claims.Valid(); err != nil {
```
mikehelmick, line 138: this call was redundant, .Valid is called inside the parse func

sethvargo (approved):

PR #1522, opened 2021-05-07
change default metric interval to 60s

mikehelmick:
## Proposed Changes

* change default metric publish interval to reduce volume

**Release Note**

```release-note
Change STACKDRIVER_REPORTING_INTERVAL to 60s by default.
```


sethvargo:
Can we override the envvar in dev and prod for a few days to make sure this doesn't mess with alerts on forward-progress alerting?

mikehelmick:
monitoring this in a live env before merge

PR #1528, opened 2021-06-14
Add admin console to load balancer

sethvargo:
🛑 ⚠️  I think this could cause downtime since it changes the hosts on the load balancer! ⚠️ 🛑  We don't have a mechanism like on the verification server to safely add new domains.

**Release Note**

```release-note
Add admin console friendly URL to load balancer list (if given).
```

### DIFF
--- a/terraform/lb.tf
+++ b/terraform/lb.tf
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 locals {
-  all_hosts = toset(concat(var.debugger_hosts, var.export_hosts, var.exposure_hosts, var.federationout_hosts))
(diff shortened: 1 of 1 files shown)


terraform/lb.tf
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 locals {
-  all_hosts = toset(concat(var.debugger_hosts, var.export_hosts, var.exposure_hosts, var.federationout_hosts))
+  all_hosts = toset(concat(var.admin_console_hosts, var.debugger_hosts, var.export_hosts, var.exposure_hosts, var.federationout_hosts))
```
mikehelmick, line 16: yes, this will cause an outage - so I think we should find a different way to make this happen

PR #1529, opened 2021-06-14
add validate signature tool

mikehelmick:
## Proposed Changes

* add easier tool for signature validation

**Release Note**

```release-note
Adds tool for signature validation.
```


sethvargo (approved):
This is good for now. I think it could be more helpful to search all the signatures and print out potential mismatches too (e.g. "found signature that verifies, but kid does not match")

PR #1530, opened 2021-06-14
Always display signature and public key on signinfo

sethvargo:
**Release Note**

```release-note
Always display signature and public key on signinfo view page.
```


mikehelmick (approved):

sethvargo:
CleanShot 2021-06-14 at 11 27 13@2x

PR #1542, opened 2021-08-04
call metric registration in a go routine and don't block sever startup

mikehelmick:
**Release Note**

```release-note
Create metric descriptors in the background during startup.
```

### DIFF
--- a/pkg/observability/stackdriver.go
+++ b/pkg/observability/stackdriver.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"runtime"
 	"strings"
+	"time"
 
 	monitoring "cloud.google.com/go/monitoring/apiv3/v2"
 	"github.com/google/exposure-notifications-server/pkg/logging"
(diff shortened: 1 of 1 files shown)


pkg/observability/stackdriver.go
```diff
@@ -121,12 +125,21 @@ func (e *stackdriverExporter) StartExporter(ctx context.Context) error {
...
-			return fmt.Errorf("failed to create MetricDescriptor: %w", err)
-		}
+		descriptorCreateRequests = append(descriptorCreateRequests, cmrdesc)
 	}
 
+	// register metrics in the background and don't block server startup on failures.
+	go func() {
+		logger.Infow("starting metric registration")
+		for _, cmrdesc := range descriptorCreateRequests {
+			_, err = mclient.CreateMetricDescriptor(ctx, cmrdesc)
```
sethvargo, line 135: I think this `ctx` has a timeout from the parent. I think we should create a new per-request context:

```go
ctx, done := context.WithTimeout(context.Background(), 10*time.Second)
defer done()
```

If we want to limit the overall time the function can run:

```go
go func() {
  ctx, done := context.WithTimeout(context.Background(), 30*time.Second)
  defer done()


  for _, cmrdesc := range descriptorCreateRequests {
    func() {
      subCtx, done := context.WithTimeout(ctx, 5*time.Second)
      defer done()

      if _, err := mclient.CreateMetricDescriptor(subCtx, cmrdesc); err != nil {
        logger.Errorw(...)
      }
    }()
  }
}()
```

mikehelmick: good catch - updated

sethvargo (approved):
@mikehelmick approved. You didn't enable auto-approve though, so you'll need to click the button.

PR #1544, opened 2021-08-09
Use timeout instead of passing context in o11y exporter

sethvargo:
Per the documentation, the provided context must not have a timeout, but we're consistently seeing context deadline exceeded errors. I think this is because we're passing in a context that has a timeout. I haven't been able to find a place where we actually call it with a timeout context, but either way, we shouldn't use the provided context on all requests since it comes from the outside. There's also already a timeout set on the export requests through the `Timeout` options property.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/observability/stackdriver.go
+++ b/pkg/observability/stackdriver.go
@@ -59,7 +59,6 @@ func NewStackdriver(ctx context.Context, config *StackdriverConfig) (Exporter, e
 	}
 
 	options := stackdriver.Options{
-		Context:                 ctx,
 		ProjectID:               projectID,
 		ReportingInterval:       config.ReportingInterval,
 		BundleDelayThreshold:    config.BundleDelayThreshold,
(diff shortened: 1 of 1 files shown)


pkg/observability/stackdriver.go
```diff
@@ -129,23 +128,30 @@ func (e *stackdriverExporter) StartExporter(ctx context.Context) error {
...
 
-	// register metrics in the background and don't block server startup on failures.
+	// Register metrics in the background and don't block server startup on
+	// failures.
+	//
+	// Future optimization: only register metrics once per buildinfo data.
 	go func() {
-		// Create a separate timeout for metric registration.
-		ctx, done := context.WithTimeout(context.Background(), time.Duration(5*len(descriptorCreateRequests))*time.Second)
+		// Overall metrics registration can take a maximum of 30s.
```
mikehelmick, line 136: I had this higher - because the 5s timeout on individual metrics is larger than 30s in practice it probably doesn't matter

mikehelmick (approved):

pkg/observability/stackdriver.go
sethvargo: It was 5s * number of metrics, which is identical to each having a max of 5 seconds, so I think it was strictly redundant?

PR #1545, opened 2021-08-11
Adding a field on TEKs for variant of concern.

flagxor:
This change introduces a new optional field on TEKs that can be used to identify a variant of concern.

**Release Note**

```release-note

TEKs can now include a variant of concern field, which can be used score variants differently.
This is download format change isn't yet supported in the reference server.

```


mikehelmick:
the final label descriptors are still under discussion - do not merge

PR #1579, opened 2021-12-21
Health Check Cache

jklaw90:
# Fixes #

If we have a failure threshold greater than 1, this would never be marked unhealthy.
It would return one unhealthy, and then healthy until the next `dbPingLimiter.Allow` and so on.

## Proposed Changes

* Cache the previous state, and use it until the next `dbPingLimiter.Allow`

**Release Note**

```release-note
cache health check status
```


sethvargo:
I think I'd rather return 429 instead of a cached result. There's no reason anyone should be hitting this endpoint more than 1/s.

jklaw90:
I was thinking in the case of liveness and readiness probes in k8s, there may be multiple probes calling this endpoint?
If it's not needed i can close it out.

PR #1581, opened 2021-12-26
Fix typo in function name

alam0rt:
## Proposed Changes

* Fix typo

### DIFF
--- a/pkg/verification/utils.go
+++ b/pkg/verification/utils.go
@@ -36,10 +36,10 @@ import (
 	"github.com/google/exposure-notifications-server/pkg/api/v1alpha1"
 )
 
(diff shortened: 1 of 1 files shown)


sethvargo (changes requested):
Thank you for the change. We actually should have removed this code a lonnnng time ago though (the comment says v0.3 release). We should probably just remove this code now.

@mikehelmick thoughts?

mikehelmick:
> Thank you for the change. We actually should have removed this code a lonnnng time ago though (the comment says v0.3 release). We should probably just remove this code now.
> 
> @mikehelmick thoughts?

I had the same thought - removal PR
http://github.com/google/exposure-notifications-server/pull/1582

PR #1598, opened 2022-04-24
Use generics for cache

sethvargo:
I'm not actually convinced this is a _good_ idea, but it's nice to avoid the type assertion.

**Release Note**

```release-note
NONE
```


sethvargo:
This does _technically_ change the contract of the cacher. It used to be able to cache anything, even a mix of types. This forces an instance of the cacher to cache the same types.

sethvargo:
Lol some of the linters don't understand the generics syntax:

```text
Error: pkg/cache/cache.go:156:1: receiver-naming: receiver name c should be consistent with previous receiver name i for invalid-type (revive)
```

PR #1602, opened 2022-07-06
add background mark and seep cache reclamation

mikehelmick:
## Proposed Changes

* Remove expired cache entries proactively, rather than wait for them to be accessed again.

**Release Note**

```release-note
Proactive cache reclamation, should have minimal improvement in RAM usage.
```

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -35,6 +35,8 @@ type Cache[T any] struct {
 	data        map[string]item[T]
 	expireAfter time.Duration
 	mu          sync.RWMutex
+	stopChan    chan bool
+	ticker      *time.Ticker
 }
 
 type item[T any] struct {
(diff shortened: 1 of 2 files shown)


pkg/cache/cache.go
```diff
@@ -52,10 +54,50 @@ func New[T any](expireAfter time.Duration) (*Cache[T], error) {
...
+		if t > v.expiresAt {
+			go c.purgeExpired(k, v.expiresAt)
+		}
+	}
+}
+
+// Stop will shutdown the background cleanup for the cache.
+func (c *Cache[T]) Stop() {
+	c.ticker.Stop()
+	c.stopChan <- true
```
sethvargo, line 100: I think you want to `close` this instead.

pkg/cache/cache.go
```diff
@@ -52,10 +54,50 @@ func New[T any](expireAfter time.Duration) (*Cache[T], error) {
...
+}
+
+// mark takes a read lock to search for expired entries and
+// enqueues deletions in separate background functions.
+func (c *Cache[T]) mark(t int64) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	for k, v := range c.data {
+		if t > v.expiresAt {
+			go c.purgeExpired(k, v.expiresAt)
```
sethvargo, line 92: This is a bug. `k` is shadowed, so this will only ever expire the first key. Need to `k := k`.

pkg/cache/cache.go
mikehelmick: added the close to the receiver

pkg/cache/cache.go
mikehelmick: good catch, added a test to cover this and fixed.

sethvargo (approved):

PR #1611, opened 2022-09-12
Gracefully shutdown prometheus exporter

sethvargo:
**Release Note**

```release-note
Gracefully shutdown the Prometheus metrics exporter when the server shuts down. This only applies if the Prometheus metrics exporter is enabled (it's off by default). This fixes a low probability issue that some metrics may be lost during shutdown.
```

### DIFF
--- a/pkg/server/metrics.go
+++ b/pkg/server/metrics.go
@@ -17,41 +17,71 @@ package server
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
 	"strings"
+	"time"
 
 	"contrib.go.opencensus.io/exporter/prometheus"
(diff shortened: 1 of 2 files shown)


sethvargo:
FWIW, this doesn't _completely_ eliminate the possibility of metrics loss, but it reduces it. We're still triggering the shutdown in a goroutine, which doesn't block shutdown.

sethvargo:
> FWIW, this doesn't _completely_ eliminate the possibility of metrics loss, but it reduces it. We're still triggering the shutdown in a goroutine, which doesn't block shutdown.

I lied - I moved it inline so this will catch all cases.

whaught (approved):

PR #1613, opened 2022-10-20
Update to bootstrap 5

sethvargo:
**Release Note**

```release-note
Upgrade to the latest UI kit version.
```


internal/admin/templates/bottom.html
```diff
@@ -1,10 +1,8 @@
 {{define "bottom"}}
 </div>
 
-  <script src="https://code.jquery.com/jquery-3.5.1.min.js"
```
sethvargo, line 4: Bootstrap doesn't need jquery anymore and I verified that we're not using it anywhere.

internal/admin/templates/top.html
```diff
@@ -1,13 +1,20 @@
 {{define "top"}}
-<html>
+<html lang="en">
 <head>
   <title>{{if .title}}{{.title}}{{else}}Exposure Notifications{{end}}</title>
-  <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.5.2/css/bootstrap.min.css"
-    integrity="sha384-JcKb8q3iqJ61gNV9KGb8thSsNjpSL0n8PARn9HuZOnIxN0hoP+VmmDGMN5t9UJ0Z" crossorigin="anonymous">
-  <link rel="stylesheet"
-    href="https://cdnjs.cloudflare.com/ajax/libs/open-iconic/1.1.1/font/css/open-iconic-bootstrap.min.css"
```
sethvargo, line 8: We used these icons exactly two places. It wasn't worth it. I removed it.

whaught (approved):
I assume you've loaded up these pages and they all look right

internal/admin/templates/bottom.html
whaught: this is awesome. Should load faster too!

mikehelmick (approved):

sethvargo:
@whaught yea - I clicked around and did all the things

